PR #46, opened 2014-02-27
Add 'Toggle Tree Side' menu item

heyvito:
This adds a new menu item called 'Toggle Tree Side', that toggles the tree view side between right and left

### DIFF
--- a/lib/tree-view.coffee
+++ b/lib/tree-view.coffee
@@ -106,7 +106,7 @@ class TreeView extends ScrollView
 
   attach: ->
     return unless atom.project.getPath()
-    @removeClass('panel-right, panel-left')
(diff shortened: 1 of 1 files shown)


TeckniX:
I'm currently on  `Version 0.174.0` yet I do not see this preference within the settings? I saw it within the 'Open my Config' file, but this thread made it seemed like it was added to the actual <kbd>cmd</kbd>+<kbd>,</kbd> settings along with the 'Show Tabs', 'Show Line Numbers'.

I'm all for hacking config files, but for something as simple as a true/false, when the above 'Show...' have a similar true/false flag alread in the settings, then that's where I expected it.

laicosly:
Thank you so much for this!   #CodeOnTheLeft4Life

PR #48, opened 2014-02-27
Copy cut paste

renz45:
Add copy/cut and paste file/directory functionality. This also uses localStorage to store the copy/cut path so that pasting across different instances of atom becomes possible.

Let me know if there is anything I missed, this is my first attempt at Atom stuff :)

### DIFF
--- a/keymaps/tree-view.cson
+++ b/keymaps/tree-view.cson
@@ -10,6 +10,16 @@
   'ctrl-|': 'tree-view:reveal-active-file'
   'ctrl-0': 'tree-view:toggle-focus'
 
+'.platform-darwin .tree-view':
+  'cmd-c': 'tree-view:copy'
+  'cmd-x': 'tree-view:cut'
+  'cmd-v': 'tree-view:paste'
+
+'.platform-win32 .tree-view':
+  'ctrl-c': 'tree-view:copy'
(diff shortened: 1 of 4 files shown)


hectorsabina:
HI, I can't copy-past several selected files in in the tree view. If I try selecting several files and copying then, it will always result in only one of them being pasted in the new location?

This happens to me when I try copying using any of these ways:
* Select several files using cmd+click then right click on any of the selected files, and copying from the contextual menu.
* Select several files using cmd+click then click on the Edit menu on top, and click on Copy.
* Select several files using cmd+click then using the shortcut (cmd+c).

Pasting in the new location always results in only one item being pasted. Also, I can't see the commands shown in the gif above selecting multiple files (copy selected/cut selected).

Using atom for Mac, 1.18

psamim:
I do not have copy or paste in my context menu, Atom 1.18.0 on Archlinux, using Nuclide.

image

peto85:
This is just for the sake of confirming it's not showing up for me:
screen shot 2017-07-17 at 10 05 13

I'll try to add Nuclide, and see if it makes any difference.

Igor10k:
@hectorsabina same here. ⌘X, ⌘C, ⌘V shortcuts work with multiple files just fine though.

PR #52, opened 2014-02-28
Preview behavior (like Sublime preview)

akitoeki:
Doesn't open a new tab every time the user click on a file item (like in Sublime), instead, it creates a atom.workspace.activePane.previewingItem, which will be removed if the user click on other file. It stays if the user doubleClick on the filename

tab

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,4 +1,10 @@
 # Tree View package
- Tree View for Atom Editor with preview mode
 
-![](https://dl.dropboxusercontent.com/u/16625639/tab.gif)
+Explore and open files in the current project.
+
+Press `cmd-\` to open/close the Tree view and `ctrl-0` to focus it.
+
+When the Tree view has focus you can press `a`, `m`, or `delete` to add, move
+or delete files and folders.
+
+![](https://f.cloud.github.com/assets/671378/2241932/6d9cface-9ceb-11e3-9026-31d5011d889d.png)


techmag:
Added sublime-tabs for now but yes add my vote to the behaviour wanted camp. 

The link for Preview Tabs (another package listed here) goes nowhere?

ahuth:
@techmag, preview tabs is [here](https://github.com/ahuth/preview-tabs). I pulled the package itself from atom.io because it stopped working after an Atom update, and I didn't have the motivation to fix it.

MateuszG:
+1

arielnoname:
+1. It's so annoying to end with thousands of files opened.

Blainegunn:
I cannot continue to use this editor if this isn't changed. After browsing through a new project I have 100+ files open.

tayeke:
Good news guys, they've added this to the core project, I guess this particular thread hasn't been updated yet though. You just have to go to the "tabs" package in your package manager and then click on settings. There is a checkbox "Use Preview Tabs".

arielnoname:
Yeah, It's been around for a while now.

gerrywastaken:
@tayeke Cheers, I didn't know about that. I think it should be on by default though. It's a great feature.

traviswimer:
Regardless of if this should or shouldn't be on by default, I never would have thought to look at the "Tabs" package settings. (I also found a useful "VCS coloring" setting there.)

It seems to me that the main Atom settings page should somehow include package settings as well and maybe a way of searching those settings. Currently its difficult to know which packages have settings or which one is likely to have a setting you are looking for.

mwean:
Is there a way to show the preview without hitting enter (or double-clicking) in the sidebar? In Sublime, you can just move through the sidebar and the preview updates in realtime.

alexbeletsky:
+1 any chances that could be merged? With large projects is quickly become a major pain in the neck..

mnquintana:
@alexbeletsky This feature was added a long time ago in a different PR. You need to enable Use Preview Tabs in the settings for the tabs package:

settings_-__users_machistequintana_code_black-screen_-_atom

alexbeletsky:
@mnquintana oh, thank you very much. I didn't catch the fact it's already merged. 

Works as expected and makes me happy :)

PS. I would make this option is "On" by default, but that's another issue I guess.

CrudamCode:
@mnquintana  thanks :+1:

mnquintana:
Since this is an old PR that was superseded by another one, I'm going to lock this thread. If you find any issues with the current implementation of Preview Tabs, do a cursory search in the issues to see if it's already been reported, and if not, feel free to open a new issue. Thanks! :smile:

PR #109, opened 2014-04-02
Add close-tree-after-open-file behaviour

StephanHoyer:
Add action to close the tree after opening a file (NERDTree-style)


kevinsawicki:
Sorry for the delay in reviewing this. It no longer merges cleanly and it is quite out of date now as well.

I'm going to close it out but please comment and update it if this still needs to be addressed.

PR #183, opened 2014-07-05
Add option to filter dot files.

matugm:
This lets you hide those annoying hidden files and directories like ***.ssh*** or ***.bashrc*** and you can still open them with ***Ctrl + P*** if you really need to.

image


izuzak:
You should now be able to hide all files and folders beginning with a dot by using a `.*` pattern for Ignored Names and turning on "Hide ignored names" in the settings for tree view. See my GIF here: https://github.com/atom/atom/issues/1966#issuecomment-139991595. Adding a separate option just for this doesn't seem like it would give much benefit over the available approach, so I'm closing this. If there's a reason to keep this open or I'm missing anything -- please let me know. :v:

PR #230, opened 2014-09-26
Added functionality to focus parent folder on deletion

### LINKED ISSUE #60: Losing highlight focus on deletion of files/folders
When you delete a file in a directory using the hotkeys, it'd be nice if the highlight focus moved to the next file or at least to the top of the directory that you were in so that you don't have to click back to where you were or scroll all the way down using the arrows. Right now the highlight focus disappears so if you press the down arrow, the highlight starts back at the top of the directory

deiga:
Fixes #60

### DIFF
--- a/lib/tree-view.coffee
+++ b/lib/tree-view.coffee
@@ -414,8 +414,8 @@ class TreeView extends ScrollView
     dialog.attach()
 
   removeSelectedEntries: ->
-    parents = _.uniq(_.map(@selectedPaths(),  (path) ->
-      return path.substring(0, path.lastIndexOf('/'))
+    parents = _.uniq(_.map(@selectedPaths(),  (selected_path) ->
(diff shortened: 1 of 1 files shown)


johngeorgewright:
:+1:

I also think it maybe nice to focus on the previous sibling... and if there isn't one, the parent folder.

benogle:
> I also think it maybe nice to focus on the previous sibling... and if there isn't one, the parent folder.

I agree with this. @deiga if you're still down for messing with this, I can guide it to merge. 

We'll also need a spec.

cc @atom/feedback

deiga:
Yeah, sure! I'll look at getting a spec done :)

thedaniel:
@deiga did you need help getting a spec going here?

deiga:
@thedaniel possibly, yeah. I seem to be a bit short on time :P

PR #244, opened 2014-10-17
WIP use the new panel API

benogle:
This works, but the specs are broken currently.

The tree view relies on `detach` to hide and `attach` to show. These concepts are pretty baked into the tests. 

The new panel API always attaches the node, then just hide/shows it. 

@kevinsawicki feel free to start over if you want to. This is just here for reference.

New Panel API: https://github.com/atom/atom/pull/3837


kevinsawicki:
The Tree View was migrated to use the panel API in https://github.com/atom/tree-view/pull/266

PR #247, opened 2014-10-23
Add feature to automatically reveal the current active file in the tree ...

vraravam:
...view when the editor is opened.


vraravam:
Sorry - I dont have enough knowledge about coffeescript and/or the atom API to get this working - any help would be appreciated such that this can finally end up in the OOB functionality for tree-view.

50Wliu:
Superseded by #754.

Issue #235, opened 2014-10-02
Add an option for case-sensitive sort order


zhanzhenzhen:
The sort order is now case-insensitive. But that brings the inconsistency with the GitHub web pages. Could you make it case-sensitive, or add an option for that?

hnordt:
Looks like Atom doesn't like case-sentitive sorting. Atom `sort-lines` package doesn't support case-sensitive sorting too.

TrevorBurnham:
I've created a PR for this: https://github.com/atom/tree-view/issues/916

lee-dohm:
Thank you everyone for your feedback and interest.

We've decided that we aren't going to add this feature and that we aren't going to accept PRs for more customized sort orders in the Tree View.

cristian-sima:
Why not? It is usefull for the case of a Redux/React project, organized with a lot of modules like here https://www.smashingmagazine.com/2016/09/how-to-scale-react-applications/

![image](https://user-images.githubusercontent.com/12199042/36597879-29991f4a-18b3-11e8-9cc9-404d505ea40f.png)

lee-dohm:
1. Because sort orders are one of those fiddly things where everyone wants their own super-custom preferences met
1. Offering more than one means that we:
    1. Have to design a set of options for them
    1. Have to come up with descriptive, easy-to-understand language for what are often very subjective preferences
    1. Have to spend a lot of time diagnosing whether people's issues are real bugs or just another sort order that they want

Or ... we can punt on the whole thing and rely on the fact that the tree-view is open source and if people really want a different sort order, they can fork the tree-view to add and support it themselves.

We've opted for the latter.

Issue #64, opened 2014-03-03
Files/directories sorted in the tree view


jlecour:
Hi,

I don't understand why some people can even like having directories grouped/sorted together, then files grouped/sorted together, but I respect that.

However, coming from a Mac only background, I'm very used to having all items in the tree view sorted alphabetically, regardless of their type, to have for example the `config` directory just near my `config.ru` file.

Is there a setting to do that ?
If not, it should be :)

Thanks a lot for making Atom and letting us play with it while in beta.

damiendr:
Would also like the option to sort by date. My use case is note-taking, where it is convenient to display the most recently updated notes at the top (like Notational Velocity).

50Wliu:
@damiendr, could you please open another issue for this?  The directory/file sorting functionality that seems to be the main point of this issue has been fixed.

damiendr:
@50Wliu done, just opened #343 for that.

PR #260, opened 2014-11-24
Alphabetical tree-view sorting (files and directories mixed together, Mac OS X finder style)

mscottford:
An experimental implementation for addressing #64, and perhaps a path towards addressing #235.

I thought I would open the conversation about this approach before I take the time to:

- [x] add a configuration parameter alphabetical sorting
- [x] create specs
- [x] search for odd edge cases

tree-sort-demo


thedaniel:
Since this is the way Finder does it when you sort by name, would you expect the tree-view on OS X to match it out of the box? If so, we would need to decide how to set the defaults on the other two platforms as well.

thedaniel:
I think my preference would be to leave it as it is by default but provide this option, and/or possibly add it to the context menu on the top-level tree item.

mscottford:
I think it's okay to have alphabetical sorting be the default, since that's the current behavior. And it's the current behavior of similar text editors. Textmate is the only one I can think of off the top of my head that sorted it's tree view using "Finder style".

thedaniel:
I opened a PR against your branch that makes the existing specs pass again: https://github.com/mscottford/tree-view/pull/1

mscottford:
Thanks @thedaniel!

mscottford:
I'm getting an odd error when I try to use this branch via `apm link`. 

screenshot 2015-01-13 21 18 05


@thedaniel  Any thoughts for how to work around it? I've tried googling (unless my google-fu needs some work), and I've tried running `apm update`.

kevinsawicki:
Try `apm rebuild`

mscottford:
That did the trick. Thanks!

mscottford:
Specs have been added. Please let me know if there's anything else that needs to be wrapped up to make this merge-able.

thedaniel:
I'm generally pro this - any other comments from @atom/core?

kevinsawicki:
I'm pro on this overall, I think the setting name could be improved though, the word `inline` throws me off a bit when trying to understand at first glance.

* `showFoldersBeforeFiles`
* `sortFoldersBeforeFiles`
* `showFilesAfterFolders`
* `sortFilesAfterFolders`

Any other ideas?

thedaniel:
`groupFolders` | `groupDirectories` ?

nathansobo:
I like @kevinsawicki's suggestion of `sortFoldersBeforeFiles`.

mscottford:
`sortFoldersBeforeFiles` works for me. I think that would be much more clear. Naming things is hard. :)

mscottford:
Name updated. Please take another peek, and let me know if there's anything else that needs to be addressed.

thedaniel:
:+1: :ship:

thedaniel:
This will be in Atom 0.177.0.

jmeas:
@thedaniel, did this make it into 0.177.0? I'm not seeing the option.

Atom v0.177.0, tree-view v0.154.0

kevinsawicki:
It should be in 0.177:

screen shot 2015-02-03 at 1 19 30 pm

mscottford:
@jmeas It's there, but you have to pull up the settings for the tree-view package to find it.

1. From settings, click on Packages
2. Search for tree-view
3. Click the Settings button in the tree view search result
4. You should see something similar to the screenshot that @kevinsawicki posted.

jmeas:
Yup, I'm familiar with adjusting package settings. It's just not there!

 

:open_mouth:

mscottford:
Looks like the package is disabled. I wonder if that could be why.

jmeas:
> Looks like the package is disabled. I wonder if that could be why.

Ha, wow – so enabling it did nothing, but after enabling it and closing out/reopening Atom, the setting then decided to show up. Interesting!

Thanks for the help @mscottford :+1:

PR #265, opened 2014-11-28
add w-resize cursor rule

valtido:
This should fix the cursor problem (on windows) as a fallback, if `col-resize` is not recognised then `w-resize` should be the default on the handle

### DIFF
--- a/stylesheets/tree-view.less
+++ b/stylesheets/tree-view.less
@@ -15,6 +15,7 @@
     top: 0;
     bottom: 0;
     width: 10px;
+    cursor: w-resize;
     cursor: col-resize;
     z-index: 3;
   }


kevinsawicki:
@mnquintana does the tree-view still need this on Windows like you added to panes in https://github.com/atom/atom/pull/7310 ?

mnquintana:
@kevinsawicki It looks like it's already in the stylesheet at https://github.com/atom/tree-view/blob/master/styles/tree-view.less#L58

kevinsawicki:
@mnquintana awesome, I missed that, closing this out then since this should be fixed already, @valtido please comment if you still experience this issue on Atom 1.0.3+

PR #306, opened 2015-01-05
Add enableVcsColoring config variable

saschagehlich:
The `tree-view.enableVcsColoring` configuration variable specifies
whether or not the tree view should change the color of modified /
newly added files and directories. Defaults to true.

(Bad branch naming, sorry.)

### DIFF
--- a/lib/directory-view.coffee
+++ b/lib/directory-view.coffee
@@ -44,8 +44,9 @@ class DirectoryView extends HTMLElement
     @expand() if @directory.isExpanded
 
   updateStatus: ->
-    @classList.remove('status-ignored', 'status-modified', 'status-added')
-    @classList.add("status-#{@directory.status}") if @directory.status?
+    if atom.config.get('tree-view.enableVcsColoring')
+      @classList.remove('status-ignored', 'status-modified', 'status-added')
+      @classList.add("status-#{@directory.status}") if @directory.status?
 
   subscribeToDirectory: ->
     @subscriptions.add @directory.onDidAddEntries (addedEntries) =>
--- a/lib/directory-view.coffee
+++ b/lib/directory-view.coffee
@@ -44,9 +44,9 @@ class DirectoryView extends HTMLElement
     @expand() if @directory.isExpanded
 
   updateStatus: ->
-    if atom.config.get('tree-view.enableVcsColoring')
-      @classList.remove('status-ignored', 'status-modified', 'status-added')
-      @classList.add("status-#{@directory.status}") if @directory.status?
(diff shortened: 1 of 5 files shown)


lib/directory-view.coffee
```diff
@@ -44,8 +44,9 @@ class DirectoryView extends HTMLElement
     @expand() if @directory.isExpanded
 
   updateStatus: ->
-    @classList.remove('status-ignored', 'status-modified', 'status-added')
-    @classList.add("status-#{@directory.status}") if @directory.status?
+    if atom.config.get('tree-view.enableVcsColoring')
+      @classList.remove('status-ignored', 'status-modified', 'status-added')
+      @classList.add("status-#{@directory.status}") if @directory.status?
```
lee-dohm: Should this be:

```coffee
@classList.remove('status-ignored', 'status-modified', 'status-added')
@classList.add("status-#{@directory.status}") if atom.config.get('tree-view.enableVcsColoring') and @directory.status?
```

This way the status classes are always removed, just selectively added. I can imagine scenarios where Atom is started with the setting enabled and then it is disabled without restarting Atom or reloading the window ... the classes will still be there for some parts of the tree, but not others. And I don't see any tests covering this particular scenario.

lib/file-view.coffee
```diff
@@ -30,8 +30,9 @@ class FileView extends HTMLElement
     @updateStatus()
 
   updateStatus: ->
-    @classList.remove('status-ignored', 'status-modified',  'status-added')
-    @classList.add("status-#{@file.status}") if @file.status?
+    if atom.config.get('tree-view.enableVcsColoring')
+      @classList.remove('status-ignored', 'status-modified',  'status-added')
+      @classList.add("status-#{@file.status}") if @file.status?
```
lee-dohm: Same as `directory-view.coffee`

saschagehlich:
@lee-dohm Thanks - fixed! :)

benogle:
@saschagehlich can you rebase this on master? 

Anyone from @atom/feedback wanna give this a gut check?

maxbrunsfeld:
@saschagehlich Could you explain why you'd want to turn this feature off? Is it because it's highlighting incorrectly for you in some cases?

saschagehlich:
Basically because I used to think that the coloring feature is annoying - but now I actually like it ;)

maxbrunsfeld:
Ok, so what do you think; can we close this for now? I just ask because I think it's good to avoid adding extra settings when possible.

benogle:
> but now I actually like it

:+1:

PR #312, opened 2015-01-07
Add a generic way of hiding files by regex.

apeiron:
This introduces a new config value, `atom.tree-view.hidePattern`, which stores a regex to use to determine whether to hide a file or directory from the tree view.

Please be harsh! This is my first time writing coffeescript and my first time contributing to atom. I am not attached to the implementation, but defintely want the feature.

### DIFF
--- a/lib/directory.coffee
+++ b/lib/directory.coffee
@@ -94,6 +94,16 @@ class Directory
       for ignoredPattern in @ignoredPatterns
         return true if ignoredPattern.match(filePath)
 
+    if atom.config.get('tree-view.hidePattern')
+      basePath = path.basename(filePath)
+      hidePattern = atom.config.get('tree-view.hidePattern')
+      try
+        match = basePath.match(hidePattern)
+      catch
+        console.warn("Bad regex " + hidePattern)
+        return false
+      return true if match
+
     false
 
   # Does given full path start with the given prefix?
--- a/lib/directory.coffee
+++ b/lib/directory.coffee
@@ -100,7 +100,7 @@ class Directory
       try
         match = basePath.match(hidePattern)
       catch
-        console.warn("Bad regex " + hidePattern)
+        console.warn("Bad hidePattern regex: " + hidePattern)
         return false
       return true if match
 
--- a/lib/main.coffee
+++ b/lib/main.coffee
@@ -13,6 +13,9 @@ module.exports =
     showOnRightSide:
       type: 'boolean'
       default: false
+    hidePattern:
+      type: 'string'
+      default: ''
 
   treeView: null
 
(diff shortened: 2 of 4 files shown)


lee-dohm:
Is `core.ignoredNames` insufficient for this?

ignored-names

If regex support is an absolute requirement, I'd rather see `core.ignoredNames` enhanced to support regexs. There is already a decent amount of confusion over the two different methods (`tree-view.hideIgnoredNames` and `tree-view.hideVcsIgnoredFiles`) for hiding paths in the Tree View, adding a third might be pushing it.

I'll also take a look through your code to give tips where I can see them whether or not this is ultimately accepted.

lib/directory.coffee
```diff
@@ -94,6 +94,16 @@ class Directory
...
+    if atom.config.get('tree-view.hidePattern')
+      basePath = path.basename(filePath)
+      hidePattern = atom.config.get('tree-view.hidePattern')
+      try
+        match = basePath.match(hidePattern)
+      catch
+        console.warn("Bad hidePattern regex: " + hidePattern)
+        return false
+      return true if match
+
```
lee-dohm: CoffeeScript supports string interpolation, so this might be more clear:

```coffee
console.warn("Bad hidePattern regex: #{hidePattern}")
```

You don't need to have the match variable, it could be rewritten this way:

```coffee
try
  return basePath.match(hidePattern)
catch
  console.warn("Bad hidePattern regex: #{hidePattern}")
  return false
```

Also, rather than using `console.warn`, you should probably use the new [notifications package](https://github.com/atom/notifications).

I'd also probably rewrite the section to not call `atom.config.get` twice:

```coffee
hidePattern = atom.config.get('tree-view.hidePattern')
return unless hidePattern
```

`atom.config.get` isn't a performance hog, but still.

apeiron:
Ah, I didn't realize I could pass a glob to |core.ignoredNames|. I
was going to look at minimatch to see what it did, but somehow
never got to doing so. Doh. Anyway! Given that it supports globs,
I think I can make it work for my use case of not wanting to see
dotfiles.

That said, I'll take the suggestions you made on my code (thank
you!) and look at extending |core.ignoredNames| to support regex,
if that's something people are interested in.

lee-dohm:
The [minimatch file globbing](https://www.npmjs.com/package/minimatch) is pretty robust. And since you can separate file globs with commas but not regular expressions, no special support needs to be added to the settings-view package to allow people to edit an array of values.

So, this is just my opinion (and I'm not the last word here) ... but I'm not sure the few edge cases where regular expressions outshine globs for selecting file paths justify the effort it will take to support them for this use case. I could be completely missing something though.

thedaniel:
This PR is nearing its first birthday without a merge, overlaps with ignored names, so I'm going to close it.  Thank you, though!

Issue #304, opened 2015-01-04
Open in new split pane


johngeorgewright:
Any thoughts on opening files from tree-view into new or existing split panes? If there's already a way to do it, can you point me in the direction?

kevinsawicki:
> If there's already a way to do it, can you point me in the direction?

You could implement it in a custom package but there isn't built-in support for it currently.

johngeorgewright:
OK. I'll have a go.

Would you be opposed to accepting a pull request so it's part of "tree-view", or is it preferable to have as a separate package?

kevinsawicki:
> Would you be opposed to accepting a pull request so it's part of "tree-view"?

Yes, the fuzzy finder has support for opening into splits so this would make it consistent with that flow.

johngeorgewright:
Ideal. I was just about to get started, however, when I stumbled on a problem already. How does one develop on this project? I cloned it, "linked" it, but on opening atom it says they contains "native modules that aren't compatible with this version of Atom". Do I need a different version of Atom, other than the current stable version?

johngeorgewright:
Found the problem. The version of pathfinder need to be locked to patch updates only.

johngeorgewright:
Hey there.

Functionally I've got it working. At the moment you can open the selected file in a new split by using: `ctrl-[left|right|up|down]`. And you can open a selected file in to an existing split by using `ctrl-[1-9]` where the number is the pane index.

I'd like to further improve it by displaying the panes' numbers when holding `ctrl`. But that could be a later improvement.

I'm also having some serious problems writing tests for "opening files in existing panes". I've been stuck for a couple of days and was wondering if anyone had a little time to maybe give me an idea how I can do it?

johngeorgewright:
Actually... I might as well just start the pull request process at this point.

francoiscabrol:
Thank you I would like to have this feature! I think that to add the vim keys is a good idea, since a lot of users (as me) are familiar with. But ctrl-right/left is already use under MacOS to switch between virtual desktops. More, using the atom core "pane:split-*" default keybindings would be easier, no?

johngeorgewright:
Ah. Yes, I was weighing up weather to use the `ctrl-k [direction]` combo. Now I understand why it's there. Easy change.

johngeorgewright:
What about opening in existing panes? Should that also be `ctrl-k [pane-number]`?

francoiscabrol:
[...]

PR #319, opened 2015-01-09
Opening files in new and existing split panes

johngeorgewright:
See #304 

tree-view-split

Open the selected file in a new split by using: `ctrl-(left|right|up|down)`. And you can open a selected file in to an existing split by using `ctrl-[1-9]` where the number is the pane index.


johngeorgewright:
S'alright... managed to gets the tests working :smile:

package.json
```diff
@@ -14,7 +14,7 @@
     "event-kit": "^1.0.0",
     "fs-plus": "^2.3.0",
     "minimatch": "~0.3.0",
-    "pathwatcher": "^2.0.12",
+    "pathwatcher": "~2.0.12",
```
kevinsawicki: This should probably stay as `^`, was there a reason you needed to change it?

spec/tree-view-spec.coffee
```diff
@@ -15,6 +15,9 @@ waitsForFileToOpen = (causeFileToOpen) ->
 describe "TreeView", ->
   [treeView, root, sampleJs, sampleTxt, workspaceElement] = []
 
+  selectEntry = (path) ->
```
kevinsawicki: Usually when a file has `path = require 'path'` at the top, we try not to use `path` as a variable name since it conflicts, for example, `path.join` would be tricky to use inside this function if needed.

What do you thinking about calling it `pathToSelect` instead?

package.json
johngeorgewright: The version that was downloaded was causing an incompatible error to be thrown by atom.

See #304

spec/tree-view-spec.coffee
johngeorgewright: Sure thing.

johngeorgewright:
I've changed the key mapping to use the core `pane:split-*` keys.

I.e: Now you'd use `[ctrl|cmd]-k left` to open a new split left. Opening splits in existing panes remains the same, although you'd use the `cmd-[pane]` in mac.

johngeorgewright:
How are you feeling about this request? Something you'd merge?

lib/tree-view.coffee
```diff
@@ -117,6 +121,10 @@ class TreeView extends View
      'tree-view:toggle-vcs-ignored-files': -> toggleConfig 'tree-view.hideVcsIgnoredFiles'
      'tree-view:toggle-ignored-names': -> toggleConfig 'tree-view.hideIgnoredNames'
 
+    [0..8].forEach (index) =>
+      atom.commands.add @element, "tree-view:open-selected-entry-pane-#{index + 1}", =>
```
kevinsawicki: What do you think about making this `tree-view:open-selected-entry-in-pane-#{index + 1}` just so it reads a bit more naturally when in the command palette?

kevinsawicki:
> How are you feeling about this request? Something you'd merge?

Yeah, this is really cool, I just left one minor comment, let me know what you think.

lib/tree-view.coffee
johngeorgewright: That's cool with me.

johngeorgewright:
Done :smile:

kevinsawicki:
Awesome, thanks for this :ship:

PR #320, opened 2015-01-09
Added tree-view:reveal-active-file-if-open command

AustP:
I wanted to add a feature to [atom/fuzzy-finder](https://github.com/atom/fuzzy-finder) that would reveal the active file after opening it. Unfortunately `tree-view:reveal-active-file` will open the tree view if it's currently not open. So I added a command that checks that the view is open before revealing the active file.

### DIFF
--- a/lib/main.coffee
+++ b/lib/main.coffee
@@ -27,6 +27,7 @@ module.exports =
       'tree-view:toggle': => @createView().toggle()
       'tree-view:toggle-focus': => @createView().toggleFocus()
       'tree-view:reveal-active-file': => @createView().revealActiveFile()
+      'tree-view:reveal-active-file-if-open': => @createView().revealActiveFile() if @treeView?
       'tree-view:toggle-side': => @createView().toggleSide()
       'tree-view:add-file': => @createView().add(true)
       'tree-view:add-folder': => @createView().add(false)


AustP:
Any thoughts on this?

kevinsawicki:
So why do you only want to reveal when the tree view if it is already open?

AustP:
I'm not sure if checking that the tree-view is open is the best way of doing it, but I did want to have a discussion on the idea. In my development environment, I coded those changes. (Adding a config to fuzzy-finder that reveals the active file upon opening.)

The first test I did started with the tree-view open. I opened the fuzzy-finder, typed in a name and hit enter. It loaded up and the file was selected in the tree-view. Perfect.

The next test I did started with the tree-view closed. I opened the fuzzy-finder, typed in a name and hit enter. It loaded up and the tree-view popped open and the file was selected.

The popping open of the tree view seems off. So there are 3 ways we can go with this:

1. Make the `reveal-active-file` command depend on state.
2. Make a new command available (`reveal-active-file-if-open`) that depends on state.
3. Do nothing and say that popping open the tree view is the intended effect.

Personally I think number 2 is best, but I could be persuaded otherwise.

kevinsawicki:
Just throwing another idea out there, what do you think about adding a config setting to the tree view instead that always reveals the active file?

So if you switch tabs or open files (via the fuzzy finder), the file is always revealed if the tree view is already open.

That way the fuzzy finder wouldn't need a specific enhancement, it would just be a general preference that if people want the active file selected and revealed in the tree view, they can have it.

PR #322, opened 2015-01-12
Open multiple folders in tree-view

gstack:
We've been needed this. Any acknowledgements?


maxbrunsfeld:
I think this functionality is already implemented now.

PR #324, opened 2015-01-13
Rename tags to use custom elements

benogle:
Caveat: the `tree-view` element is at the root of the package's markup. `.tree-view` pointed at the list and was a child of the resizer and scroller. The resizer will likely be pulled out into the panel eventually.

So to replace the `.tree-view` exactly, one would need to use `tree-view .list-tree`. Maybe this is not ideal.

Refs https://github.com/atom/atom/issues/5020

### TODO

* [x] Use `tree-view` tag in place of `.tree-view`
* [ ] Use `tree-view-file` tag in place of `.file`
* [ ] Use `tree-view-directory` tag in place of `.directory`

### DIFF
--- a/keymaps/tree-view.cson
+++ b/keymaps/tree-view.cson
@@ -10,7 +10,7 @@
   'ctrl-|': 'tree-view:reveal-active-file'
   'ctrl-0': 'tree-view:toggle-focus'
 
-'.platform-darwin .tree-view':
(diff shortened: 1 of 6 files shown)


benogle:
After talking to @nathansobo, we decided to punt on this until we are ready to do a larger rethink of the tree-view. Changing all these elements has implications and generates a lot more work than originally perceived.

PR #337, opened 2015-01-22
Added "Automatically Reveal File" config

AustP:
This adds the config option to Automatically Reveal the Active File. The 3 values are `Never`, `Upon Opening`, and `Upon Focusing`. I also added specs for this.

@kevinsawicki This is in reference to #320.

### DIFF
--- a/lib/main.coffee
+++ b/lib/main.coffee
@@ -13,6 +13,14 @@ module.exports =
     showOnRightSide:
       type: 'boolean'
       default: false
+    automaticallyRevealFile:
+      type: 'string'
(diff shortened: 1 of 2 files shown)


AustP:
Before merging this, I just thought of a few things.

1. We should probably make sure the file isn't in a hidden folder before revealing it. (Maybe this belongs in the revealActiveFile function)
2. VCS files shouldn't be revealed. (Doing a git commit within Atom doesn't really need to be revealed)

Thoughts?

AustP:
I changed my mind. I don't like this.

PR #342, opened 2015-01-27
Removed /root switch when opening a directory. Just used select instead.

zephraph:
When opening a directory, running explorer.exe /root,[directory] prohibits the user from navigating outside of that directory without an error. Select can be used for both directories and files and gives the desired results. This fixes an [issue](https://github.com/atom/atom/issues/5218) reported on atom.


kevinsawicki:
Great, thanks for fixing this.

zephraph:
Welcome!

PR #344, opened 2015-01-27
Change tree directory loading to use async IO calls

mostafaeweda:
* Deprecates the synchronous version of getEntries
* Introduce a new getEntries that's 80% copied from: https://github.com/atom/node-pathwatcher/blob/master/src/directory.coffee#L224-L246
Maybe in the future, the tree directory and file classes should inherit from the Atom API's File/Directory classes which now exist in `node-pathwatcher`

Related, but not a dependency: https://github.com/atom/node-pathwatcher/pull/56


lib/file.coffee
```diff
@@ -43,6 +40,12 @@ class File
   onDidStatusChange: (callback) ->
     @emitter.on('did-status-change', callback)
 
+  loadRealPath: ->
```
kevinsawicki: Why was this changed to use a promise? It doesn't seem like the promise is used anywhere, apologies if I'm missing something.

kevinsawicki:
Just curious, what will happen if you open and close a directory quickly multiple times and multiple promises get created to list/create the entries? How is that handled?

lib/file.coffee
mostafaeweda: You are correct, int this package, promises aren't used anywhere else and this PR introduces with requiring `Q`
personally, I don't like promises, but I wanted to be consistent with Atom's core usages of promises.

Since the package was doing most of the code synchronously, there wasn't a clear cut for me whether to use promises or callbacks.

I don't mind refactoring it back to callback if you think including q and using promises is confusing for this package

kevinsawicki: No, I'm totally cool using promises in this package.

I just didn't see why it was needed in this specific case.

mostafaeweda:
I see that the rendering of that is handled by the `directory-view` to remove items on collabse and  https://github.com/mostafaeweda/tree-view/blob/master/lib/directory-view.coffee#L118 and the `onDid$` events and subscriptions are making sure the listing under the directory is always in sync with the most recent resolved promise results.

lib/file.coffee
mostafaeweda: :+1: just for standardizing a away to do async stuff

lib/directory.coffee
```diff
@@ -187,33 +189,76 @@ class Directory
 
     directories.concat(files)
 
+
+  # Public: Reads file entries in this directory from disk asynchronously.
+  #
+  #   * `entries` An {Array} of {File} and {Directory} objects.
+  getEntries: () ->
```
kevinsawicki: These `()` aren't needed.

lib/directory.coffee
```diff
@@ -225,8 +270,7 @@ class Directory
   # changes.
   expand: ->
     @isExpanded = true
-    @reload()
-    @watch()
+    @reload().then () => @watch()
```
kevinsawicki: The `()` before the `=>` aren't needed.

lib/directory.coffee
```diff
@@ -149,7 +151,7 @@ class Directory
           when 'change' then @reload()
           when 'delete' then @destroy()
 
-  getEntries: ->
+  getEntriesSync: ->
```
kevinsawicki: Will this still be called anywhere?

mostafaeweda: No, I think I should remove it!
Coming in the next commit

PR #361, opened 2015-02-09
Alt-click to open in Split

SebastianSzturo:
**Problem**: Split panes are a powerful tool, but usually hard to access for beginners. 
**Solution**: alt-click to open the entry in a split pane. 

recoridng

Any thoughts on this?

Open Questions:
- Should alt-click always open in the same split pane or create new split views?
- Controlling the position of the split pane with other modifiers like shift?
- Should we add the "Split Up/Down/Right" menu items from the tab menu to the tree-view?

### DIFF
--- a/lib/tree-view.coffee
+++ b/lib/tree-view.coffee
@@ -192,6 +192,7 @@ class TreeView extends View
     switch e.originalEvent?.detail ? 1
       when 1
         @selectEntry(entry)
+        @openSelectedEntrySplit 'horizontal', 'after' if entry instanceof FileView and e.altKey
         @openSelectedEntry(false) if entry instanceof FileView
         entry.toggleExpansion(isRecursive) if entry instanceof DirectoryView
       when 2


kevinsawicki:
>  but usually hard to access for beginners. 

How about in the context menu then? With split up, left, down, and right items.

This is available on the editor's context menu so having it in the tree view as well would feel natural to me.

SebastianSzturo:
@kevinsawicki Yes thats a very good idea. I think that was the first thing I was searching for after finding out the tabs have it in their context menu. Created a pull-request at #369 

Do you think there is place for alt-click (or any other click modifier) to open split views in the tree view or should I extract that into a package?

kevinsawicki:
> Do you think there is place for alt-click (or any other click modifier) to open split views in the tree view or should I extract that into a package?

I'm not really seeing a place since since if there are keybindings and also a context menu, it feels well covered for both mouse people and keyboard people.

PR #368, opened 2015-02-13
Add support for multiple root folders

maxbrunsfeld:
Refs atom/atom#770.

* [x] Add command & context menu item for adding a new root folder
* [x] Render multiple root folders
* [x] Make sure all commands work correctly with multiple root folders
  * [x] `tree-view:reveal-active-file`
  * [x] `tree-view:paste`
  * [x] `tree-view:add-file`
  * [x] `tree-view:add-folder`
* [x] Make sure all git-integrations work properly w/ multiple repos
  * [x] copy dialog
  * [x] add dialog
  * [x] move dialog
  * [x] git status colors
* [x] Don't make the code uglier.


menus/tree-view.cson
```diff
@@ -39,6 +39,7 @@
     {'label': 'Split Right', 'command': 'tree-view:open-selected-entry-right'}
     {'type': 'separator'}
 
+    {'label': 'Add Root Folder', 'command': 'tree-view:add-root-folder'}
     {'label': 'Copy Full Path', 'command': 'tree-view:copy-full-path'}
```
maxbrunsfeld: @atom/core Does anyone have an opinion about the naming of this command and the *renaming* of the two above? We thought `Add Folder` seemed to similar to `Add Root Folder`, so we renamed it to `New Folder`, and did the same for `File`.

I thought about renaming the actual commands too, but I held off because I didn't want to break people's key-bindings.

nathansobo: Yeah, I like your idea of having "add folder" mean "add a new root folder" and "new folder" mean "create a new folder in this location".

nathansobo:
You are cranking sir.

craighooghiem:
Would be great if we could also add new root folders even if they are a child folder of the existing root folder. Handy for adding asset folders, library directories, etc.

mrameezraja:
Is it possible to have a solution of projects like in Visual Studio ?

maxbrunsfeld:
@mrameezraja We do eventually want to add some kind of project file feature, but we're not sure yet how exactly we want it to work. If you have ideas, please add them on this issue: https://github.com/atom/atom/issues/5728.

PR #369, opened 2015-02-14
Add split commands to context menu

SebastianSzturo:
recording

Now we have the same split commands available as in the editors context menu and the tab context menu.
As suggested in #361


kevinsawicki:
Awesome, thanks for adding these.

kevinsawicki:
This will be included in the next Atom release, 0.18

PR #396, opened 2015-03-04
Changing default order to natural order

victorperin:
This correction make files organizations more "human like".
Before:
order

After:
natural_order

it fixes issues:  #391 and #395

### DIFF
--- a/lib/directory.coffee
+++ b/lib/directory.coffee
@@ -157,6 +157,7 @@ class Directory
       names = fs.readdirSync(@path)
     catch error
       names = []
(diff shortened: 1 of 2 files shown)


maxbrunsfeld:
Nice work. Would you mind updating the tests that are failing? It would probably just mean changing the expected order of the entries.

victorperin:
Thanks for the tip, I'll work on that.

victorperin:
Fixed! =)

maxbrunsfeld:
Thanks!

PR #406, opened 2015-03-19
Remove "Reveal in Tree View" from items without file paths

mnquintana:
Part of atom/settings-view#354

This removes Reveal in Tree View from appearing in settings-view's context menu, where it never really made sense.

Also please let me know if there would be a performance hit from the `'atom-pane .item-views > *:not(.settings-view)'` selector – generally there is with `*`, but I'm not sure if that still applies to how Atom parses selectors for menu entries.

Also let me know if you think I should add a spec for this (I'm thinking maybe).

- [x] Depends on atom/atom#6383

### DIFF
--- a/menus/tree-view.cson
+++ b/menus/tree-view.cson
@@ -94,10 +94,10 @@
     {'label': 'Paste', 'command': 'tree-view:paste'}
   ]
 
-  'atom-pane .item-views > *:not(.settings-view)': [
+  'atom-pane atom-text-editor': [
     {'label': 'Reveal in Tree View', 'command': 'tree-view:reveal-active-file'}
   ]
 
-  'atom-pane .tab.active:not([data-type="SettingsView"])': [
+  'atom-pane .tab.active[data-type="TextEditor"]': [
     {'label': 'Reveal in Tree View', 'command': 'tree-view:reveal-active-file'}
   ]
--- a/menus/tree-view.cson
+++ b/menus/tree-view.cson
@@ -94,10 +94,10 @@
     {'label': 'Paste', 'command': 'tree-view:paste'}
   ]
 
-  'atom-pane .item-views': [
+  'atom-pane .item-views > *:not(.settings-view)': [
     {'label': 'Reveal in Tree View', 'command': 'tree-view:reveal-active-file'}
   ]
 
-  'atom-pane .tab.active': [
+  'atom-pane .tab.active:not([data-type="SettingsView"])': [
     {'label': 'Reveal in Tree View', 'command': 'tree-view:reveal-active-file'}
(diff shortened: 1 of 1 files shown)


maxbrunsfeld:
I think it would also make sense to avoid showing `Reveal in Tree View` on things like the Deprecation Cop view, Markdown Preview, etc. Maybe we should *only* include this option for `atom-text-editor` elements. That would make the selector a little simpler. What do you think @mnquintana?

mnquintana:
Ah yeah, that would be better! (I wasn't sure if we'd ever want it to appear in something other than `atom-text-editor`)

kevinsawicki:
> I wasn't sure if we'd ever want it to appear in something other than atom-text-editor

Yeah, keeping it available on images and archive files would be nice

mnquintana:
> Yeah, keeping it available on images and archive files would be nice

The other downside with only showing it for `atom-text-editor` is that it still appears in the context menu for mini editors in settings-view. Hm.

maxbrunsfeld:
This is getting a little more complicated, but maybe there should be a `has-file-path` class added to the `Pane`'s `.item-views` div when the active item implements `::getPath`. That way, the selector could be `atom-pane .item-views.has-file-path`.

> it still appears in the context menu for mini editors in settings-view.

I think we could handle that by adding `:not([mini])` to the selector.

mnquintana:
> This is getting a little more complicated, but maybe there should be a has-file-path class added to the Pane's .item-views div when the active item implements ::getPath. That way, the selector could be atom-pane .item-views.has-file-path.

I like that approach – it seems a lot cleaner to me. I'll take a look at it.

mnquintana:
@maxbrunsfeld @kevinsawicki I've been working on adding this to core, but I have a few questions as I write the spec since I'm still new to hacking on Atom:
- Would it be the `PaneModel` I'd expect to implement `::getPath`?
- Should I test that by creating mocks like in the spec at https://github.com/atom/atom/blob/master/spec/pane-view-spec.coffee#L87?

kevinsawicki:
> Would it be the PaneModel I'd expect to implement ::getPath?

Yup, I'd think you'd sniff for a `getPath` function in https://github.com/atom/atom/blob/7ea71deb96f264448676385f17279659846a324d/src/pane-element.coffee#L68 and toggle the class there.

> Should I test that by creating mocks like in the spec at https://github.com/atom/atom/blob/master/spec/pane-view-spec.coffee#L87?

Yup, sounds good

mnquintana:
@maxbrunsfeld @kevinsawicki This is ready for review now

maxbrunsfeld:
:ship:

PR #413, opened 2015-03-29
preview selected file on key 'right'

monsterkodi:
* Added a new command `preview-selected-entry` and attached it to the `right` cursor key.
* The `preview-selected-entry` command is the same as `open-selected-entry` but with a false `activatePane` argument for the `openSelectedEntry` method.
* Added specs to test the new behaviour. 
Fixed specs for `tree-view:open-selected-entry`, they didn't test the command properly.

### DIFF
--- a/keymaps/tree-view.cson
+++ b/keymaps/tree-view.cson
@@ -57,7 +57,7 @@
   'ctrl-9': 'tree-view:open-selected-entry-in-pane-9'
 
 '.tree-view':
-  'right': 'tree-view:expand-directory'
+  'right': 'tree-view:preview-selected-entry'
   'ctrl-]': 'tree-view:expand-directory'
   'l': 'tree-view:expand-directory'
   'left': 'tree-view:collapse-directory'
--- a/lib/tree-view.coffee
+++ b/lib/tree-view.coffee
@@ -106,6 +106,7 @@ class TreeView extends View
      'tree-view:recursive-expand-directory': => @expandDirectory(true)
      'tree-view:collapse-directory': => @collapseDirectory()
      'tree-view:recursive-collapse-directory': => @collapseDirectory(true)
+     'tree-view:preview-selected-entry': => @openSelectedEntry(false)
      'tree-view:open-selected-entry': => @openSelectedEntry(true)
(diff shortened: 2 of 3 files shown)


thedaniel:
I think this is going to need to be rethought in the context of atom's preview tabs (cf. https://github.com/atom/tabs/pull/145), though I'm generally pro adding a preview keybinding to tree-view. Thoughts?

You can simulate this behavior by turning on preview tabs in the tabs package, then hitting enter will open a preview tab that will go away unless it's modified. So this could be reimplemented as a keybinding to tree-view:open-selected-entry that then refocuses the tree-view.

It might be confusing if the tabs package has preview tabs turned off, though. Then right-arrow would be something like 'open without focus' in that context, which is perhaps useful but also potentially confusing.

@atom/feedback any thoughts here? I'm not sure if @monsterkodi is still interested in this after six months but previewing/opening does seem to be like a very comfortable binding for right-arrow in the tree view.

monsterkodi:
yes, I would still be interested in a fix for this, even after six month ;-)

I don't think there would be any problem with the preview tabs settings on or off.

The point is this: 
I can open/preview a file without setting focus on the editor by clicking with the mouse in the treeview. 
But I can't do the same via the keyboard.

monsterkodi:
I made myself a command that implements your suggestion, so I don't have to wait another half a year.
thanks for the hint :-)

```coffee-script
atom.commands.add '.tree-view', 'tree-view:preview', ->
    for panel in atom.workspace.getLeftPanels()
        if panel.item.constructor.name == "TreeView"
            treeView = panel.item
            treeView.openSelectedEntry()
            setTimeout treeView.toggleFocus.bind(treeView), 1
            return
```

I still think that somthing similar should be implemented in the tree-view itself.

thedaniel:
Glad you were able to solve it for yourself - I'm going to tag this with `help-wanted` in case anyone else wants to pick it up again for tree-view.

monsterkodi:
you helped me help myself :-) thank you.

kuychaco:
Feature added in #712. Thanks @monsterkodi!

monsterkodi:
Thank you, kuyachaco.

PR #414, opened 2015-03-29
file name filter as a service

monsterkodi:
Hello, I recently created a package to filter the tree view: [tree-view-filter](https://github.com/monsterkodi/tree-view-filter)
In my first attempt I just set the display style of the file entries to `none`.  
The problem with that approach is that the navigation with the keyboard doesn't work as expected: the hidden items still get keyboard focus.
I figured that i can't resolve this without some changes in the tree-view package itself.

With the changes in this pull request I could get the tree-view-filter package to work as expected without changing the current functionality of the tree-view. 

Please let me know If I you want me to change anything.

### DIFF
--- a/lib/service.coffee
+++ b/lib/service.coffee
@@ -17,6 +17,11 @@ module.exports =
                 @fileNameFilterFunctions.push filterFunc
             @
   
+        removeFileNameFilterFunction: (filterFunc) =>
+            if filterFunc? and filterFunc in @fileNameFilterFunctions
+                @fileNameFilterFunctions.splice @fileNameFilterFunctions.indexOf(filterFunc), 1
+            @
+  
         isFileNameFiltered: (filePath) =>
             for filterFunc in @fileNameFilterFunctions
                 if filterFunc(filePath) == true


monsterkodi:
P.S.: the branch [use_tree_view_service_to_filter_files](https://github.com/monsterkodi/tree-view-filter/tree/use_tree_view_service_to_filter_files) contains a version of the tree-view-filter that uses the new tree-view service

monsterkodi:
Could someone please let me know if this has a chance of getting merged?
Any feedback is appreciated.
Thanks in advance.

monsterkodi:
still needs review :)

aphecetche:
Hi,

Any news about this one ? 

As a former user of XCode wanting to switch to Atom, I find myself in a position where I really miss(ed) the tree view live filter (plus the ability to list only the files with a git status, but that's another story...)

The current tree-view-filter package is indeed working expect for, as noted, that it's not 100% live (i.e. the not-expanded directories don't get the filter applied when you then expand them). 

Regards,

tlnagy:
Hey @50Wliu, any chance that you or someone else could view these changes? This seems like it would be useful.

monsterkodi:
Hello, 
I found a way to make the desired behaviour work without changes needed in the tree-view package. I still think that the solution in this PR is better (more general and less prone to break with future changes to the tree-view), but after two years I think its time for some hacks :-)

PR #421, opened 2015-04-08
Rename "Remove Project Folder" to "Detach Project Folder"

bhuga:
I ran into this today. I had accidentally added my home directory as a project folder, and the option to fix that was to "remove" it. It wasn't clear to me whether or not this folder was going to be removed from the disk. I chose not to try it and went looking for help.

tree-view-spec coffee - _users_ben_github_tree-view - atom

Not sure if "detach" is the right word and I have not tested this locally. Just opening up a PR to start a discussion.

### DIFF
--- a/lib/tree-view.coffee
+++ b/lib/tree-view.coffee
@@ -122,7 +122,7 @@ class TreeView extends View
      'tool-panel:unfocus': => @unfocus()
      'tree-view:toggle-vcs-ignored-files': -> toggleConfig 'tree-view.hideVcsIgnoredFiles'
      'tree-view:toggle-ignored-names': -> toggleConfig 'tree-view.hideIgnoredNames'
-     'tree-view:remove-project-folder': (e) => @removeProjectFolder(e)
+     'tree-view:detach-project-folder': (e) => @removeProjectFolder(e)
 
     [0..8].forEach (index) =>
       atom.commands.add @element, "tree-view:open-selected-entry-in-pane-#{index + 1}", =>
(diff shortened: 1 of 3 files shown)


maxbrunsfeld:
@bhuga Yeah, it was especially scary sounding when it was `Remove Root Folder`. I thought `Remove Project Folder` felt a little better but I can see that it's still somewhat ambiguous. 

I think if we went with `Detach`, we'd have to reference that terminology elsewhere to make it fit in. Would `Remove Folder from Project` or even `Remove Folder from Current Project` be any better?

50Wliu:
Would `Close Folder` be ok?

mnquintana:
@50Wliu That makes me think it would close (collapse) the folder if it were expanded

maxbrunsfeld:
@kevinsawicki do you have any thoughts on this? What does Eclipse call this action?

kevinsawicki:
>  What does Eclipse call this action?

Eclipse just calls it delete and then prompts you to remove it or remove + delete it.

I'm not sure detach is that intuitive here, maybe adding suffix like `Remove Project Folder from View` or `Remove Project Folder from Window` would be better, just a though.

bhuga:
`Remove from View` gets the point across to me, but I would be the first to admit that it's clunky. I'm fine to close this if nobody's happy with an alternative, too. It's not a big problem, just something to learn.

mnquintana:
FWIW Sublime calls this `Remove Folder from Project`

bhuga:
Closing as stale.

PR #422, opened 2015-04-08
WIP: Add Move command to context menu

mnquintana:
This has thrown me a few times – even though I know "Rename" can be used for moving files and directories, not having a "Move" entry in the context menu makes it seem like you can't move files in tree-view. This adds a "Move" command which does almost the same thing as "Rename".

The "Move" command now places the cursor at the start of the file name, instead of starting with the file name selected (which makes more sense for "Rename").

screencap1


mnquintana:
I'd still love to add a Move command one of these days, but this PR is waaaaay too old now, so I'm gonna close this out

PR #427, opened 2015-04-13
Only show pane split commands in context menu if file is selected

mnquintana:
The pane split commands only work if used on a file. This PR removes the pane split commands from every element in tree-view that isn't a file.

Only downside is that now the pane split commands appear at the top of the menu – changing that depends on atom/atom#6270 getting implemented (/cc @jessegrosjean)


mnquintana:
@atom/non-github-maintainers Anyone want to take a look?

thomasjo:
Looks good to me™ :sparkles:

Issue #455, opened 2015-05-28
Consider changing tree-view:toggle-vcs-ignored-files


lee-dohm:
People run into issues because they believe the editor view is focused and press <kbd>I</kbd> which toggles the VCS ignored files to display or not. Since the change is non-obvious, in many cases it can go undetected and then is deemed unexpected or buggy behavior.

Options to be considered:

* Change the default keybinding from <kbd>I</kbd> to <kbd>Cmd/Ctrl+I</kbd> (first checking if this key bind is available)
* Animate or otherwise make the change to the tree view more obvious
* Make the fact that the tree view is focused more obvious on the default themes
* Others?

/cc @simurai

mnquintana:
Refs https://github.com/atom/one-dark-ui/issues/52

bronson:
Also, maybe put a tiny `i` button in the upper right of the treeview for discoverability?  Hovering would show a tooltip about ignored files and showing the key combo.

I had no idea this feature even existed, super glad it does.  Happy to open a new issue if this is hijacking.

bronson:
Part of the confusion might be that the treeview's tab is focused differently.

Normally, when a tab receives the blue highlight, the old tab gives up its highlight.  When treeview is focused, however, the old tab doesn't change its appearance at all.

![treeview focus](https://cloud.githubusercontent.com/assets/1776/7866477/aa76c996-0524-11e5-848b-244a99c6af15.gif)

simurai:
> Also, maybe put a tiny i button in the upper right of the treeview for discoverability? Hovering would show a tooltip about ignored files and showing the key combo.

Is this something that people wanna change often? Adding a toggle button that is permanently visible might make it a bit too prominent.

> Normally, when a tab receives the blue highlight, the old tab gives up its highlight. When treeview is focused, however, the previous tab doesn't change its appearance at all.

That's a good point, I'll see if the tab marker can be disabled when focusing on the tree-view.

PR #456, opened 2015-05-28
Change VCS toggle keybind

lee-dohm:
Applies to #455 

Changes the keybinding from <kbd>I</kbd> to <kbd>Cmd/Ctrl+I</kbd>.

### DIFF
--- a/keymaps/tree-view.cson
+++ b/keymaps/tree-view.cson
@@ -16,6 +16,7 @@
   'cmd-v': 'tree-view:paste'
   'ctrl-f': 'tree-view:expand-directory'
   'ctrl-b': 'tree-view:collapse-directory'
+  'cmd-i': 'tree-view:toggle-vcs-ignored-files'
   'cmd-k right': 'tree-view:open-selected-entry-right'
   'cmd-k l': 'tree-view:open-selected-entry-right'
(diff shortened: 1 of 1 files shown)


izuzak:
How exactly does this help resolve #455? The exact keybinding doesn't seem like the problem here.

lee-dohm:
It doesn't actually resolve #455. But this change is related to that Issue in that it was suggested as a means of making the accidental activation of the command more rare. @kevinsawicki brought it up on Slack and I wrote up the Issue and this PR to get the discussion going.

izuzak:
Thanks for explaining, @lee-dohm. I'm concerned that this change will 1) annoy users who are used to the existing keybinding (I'm pretty used to it myself), and 2) not actually solve the problem. I'd be :+1: on waiting with this change until we've explored other options, since the problem itself doesn't affect that many users.

Also, can I talk you into extending the description in #455 with :link: to issues or discussion where this problem was reported?

lee-dohm:
> I'd be :+1: on waiting with this change until we've explored other options

Agreed.

> Also, can I talk you into extending the description in #455 with :link: to issues or discussion where this problem was reported?

Sure thing!

lelepride:
Github

PR #457, opened 2015-05-29
Ctrl-\ is hardly accessible on foreign keyboards; consider switching 

bentolor:
background-tips announces Ctrl-\ as shortcut to toggle the tree view. Honestly - I have *no* clue how to do this on my german keyboard, because I already need to press AltGr (= Alt + Ctrl) to enter "\" alone. 

Ctrl-K Ctrl-B works fine. Please consider switching the order of the key bindings to make them more accessible for a wider audience

toggle-treeview


bentolor:
I attached a merge request to this issue: But I'm not that happy with the result
- It changes the displayed key binding for background-tips but not in the menu. Any clue why?
- I observed further bindings like `tree-view:reveal-active-file` whose displayed key combination are not matching their actual combination

On the other hand stumbling over `tree-view:reveal-active-file` helped me to figure out how to use the default key binding: 
- `Ctrl-AltGr-<` is equal to `Ctrl-\`  (unexpected) and   
- `Shift-Ctrl-AltGr-<` equal to `cmd-|` in the keymap file (expected), but not the shortcut displayed in Atoms menu: `Ctrl-Shift-\`

kankaristo:
+1 for this. I had no idea `Ctrl-k Ctrl-b` was even there.

The tree view can also be toggled with `Ctrl-<` and `Ctrl-§` on my keyboard (Finnish/Swedish layout). Those keys are between left shift and z, and above the tab key respectively.

Also, the `Ctrl-|` key binding for `tree-view:reveal-active-file` is inaccessible for me. It would be `Ctrl-AltGr-<`, but this is recognized as `Ctrl-<` and it toggles the tree view instead. Something like `Ctrl-K Ctrl-R` would be a nice alternative.

kankaristo:
I also noticed that `Ctrl-Shift-§` also works for `tree-view-reveal-active-file`.

They probably shouldn't work, but `Ctrl-§` and `Ctrl-Shift-§` actually feel quite natural, with `§` being above `Tab`. :)

PR #463, opened 2015-06-04
Double click file in preview to unpreview tab

forivall:
Currently, this is a hacky proof of concept, but I'd like some feedback on this change, which brings back a feature that was in sublime-tabs, where a file in the tree view could be double-clicked to make the tab permanent / not a preview.

What is the proper approach to do this, how should it be tested, etc.? I'd like to finish this feature properly so that it's suitable to be merged into core!

Fixes https://github.com/atom/tabs/issues/156

### DIFF
--- a/lib/tree-view.coffee
+++ b/lib/tree-view.coffee
@@ -95,6 +95,9 @@ class TreeView extends View
 
     @on 'mousedown', '.tree-view-resize-handle', (e) => @resizeStarted(e)
 
+    @on 'dblclick', 'li[is=tree-view-file]', (e) =>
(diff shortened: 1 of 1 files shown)


kevinsawicki:
I think it might be better to start this as a pull request in the tabs package.

You could listen for the double click event there on the tree view and react accordingly to clear the preview.

I think it might make it easier to do there to ensure that only preview tab for the active pane is cleared.

kevinsawicki:
Closing this out, I added it via the tabs package in https://github.com/atom/tabs/commit/73658a9391ce91aa60c43dd710b3a6fc4c816df9

PR #468, opened 2015-06-05
Fix scroll on reveal active file

### LINKED ISSUE #271: the reveal command does not scroll the tree view when needed
that is, when the revealed file is out of view. It should scroll the tree view to make the file visible

viddo:
fix-scroll

Fixes #271 by using `scrollIntoViewIfneeded` instead, removes some of the unnecessary calculations too, and added a test case to verify the changes


kevinsawicki:
Awesome, thanks for this, great to have specs and great to delete that calculation code :1234:

PR #469, opened 2015-06-06
support nested repos

jarig:
Fixes https://github.com/atom/atom/issues/2203

This patch works quite well even with >200 nested .git repos, though with some limitations:
1. If there are a lot of nested. git repos then start-up time might noticeably increase(with ~200 repos it was 10 sec). The reason is that GitRepository object seems to be not fully async while refreshing status.
2. Status refresh of GitRepositry objects on window_focus event is disabled, so if repos changed out of the Atom it might reflect wrong status (but there is extra Atom command for refreshing those).

repo_highlights

### DIFF
--- a/lib/directory-view.js
+++ b/lib/directory-view.js
@@ -73,6 +73,7 @@ class DirectoryView {
     this.element.header = this.header
     this.element.entries = this.entries
     this.element.directoryName = this.directoryName
+    this.element.refreshRepoStatus = this.refreshRepoStatus.bind(this)
   }
 
   updateIcon () {
@@ -131,6 +132,17 @@ class DirectoryView {
   reload () {
     if (this.isExpanded) {
       this.directory.reload()
+    }
+  }
+
+  refreshRepoStatus (isRecursive = true, includeCollapsed = false) {
+    this.directory.refreshRepoStatus()
+    if (isRecursive) {
+      for (let entry of this.entries.children) {
+        if (entry.classList.contains('directory') && (entry.isExpanded || includeCollapsed)) {
+          entry.refreshRepoStatus(true, includeCollapsed)
+        }
+      }
     }
   }
 
--- a/lib/directory.js
+++ b/lib/directory.js
@@ -4,7 +4,7 @@ const {CompositeDisposable, Emitter} = require('atom')
 const fs = require('fs-plus')
 const PathWatcher = require('pathwatcher')
 const File = require('./file')
-const {repoForPath} = require('./helpers')
+const {repoForPath, getRepoCacheSize} = require('./helpers')
 
 module.exports =
 class Directory {
@@ -29,6 +29,10 @@ class Directory {
     if (fs.isCaseInsensitive()) {
       this.lowerCasePath = this.path.toLowerCase()
       this.lowerCaseRealPath = this.lowerCasePath
+    }
+    this.repo = repoForPath(this.path)
+    if (this.repo && atom.config.get('tree-view.refreshVcsStatusOnProjectOpen') >= getRepoCacheSize()) {
+      this.refreshRepoStatus()
     }
 
     if (this.isRoot == null) {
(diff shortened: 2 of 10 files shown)


lib/directory-view.coffee
```diff
@@ -27,10 +27,11 @@ class DirectoryView extends HTMLElement
       iconClass = 'icon-file-symlink-directory'
     else
       iconClass = 'icon-file-directory'
-      if @directory.isRoot
-        iconClass = 'icon-repo' if repoForPath(@directory.path)?.isProjectAtRoot()
+      relPath = repoForPath(@directory.path)?.relativize(@directory.path + '/')?.length
+      if relPath == 0
```
50Wliu: Looks like this can be `unless relPath`

50Wliu:
Right now Travis is reporting quite a few lint errors - you should fix those.

In addition, this will need specs before it can be considered for merging.

lib/directory-view.coffee
jarig: unless triggers in case of 'undefined' as well, but in this case it should trigger only if string is empty(but not undefined).

jarig:
I've fixed all probs with existing specs and added 4 new ones (12 tests) covering all new functionality. Also created GIF showing what it is about

wjwwood:
This works for me, but the diffs in the gutter of changed files doesn't seem to work anymore.

jarig:
Rebased on latest changes.

>This works for me, but the diffs in the gutter of changed files doesn't seem to work anymore.
Well yeah, that should be fixed on git-plus (or whatever plugin you use) side.
Instead of taking project root it should take the nearest upwards .git folder.

However git diffs with flat project structure should still work.

ruffsl:
This is awesome

50Wliu:
Sorry, but it looks like there's more conflicts :grimacing:.

jarig:
Re-based again, with previous re-base accidentally took some extra files to commit(like packages.json), fixed that.

mikekelly1:
@jarig Do modified files in sub-repositories show up when you press CTRL-T with your fix?

jarig:
>Do modified files in sub-repositories show up when you press CTRL-T with your fix?

Unfortunately not. Atom relies on the fact that there is only 1 GitProvider per project, but it should be changed to support multiple, then those providers can be updated from TreeView and in this way other things will work as well.

phord:
fyi - I have a version (phord/tree-view@fad00868726a35231d73a10f683aac8bb3271b81) that does manage to track modified file subscriptions successfully.  But startup is slow.  I'm working on improving it, though.  Suggestions welcome.

jarig:
@phord , @mikekelly1 
I've made changes to support ctrl+t as well as highlighting of modified lines in files. Performance is quite good, tested with ~250 nested repos.
Tests are failing though, will check later why, in "real life" all works :)

jarig:
Tests passing as well now,

wjwwood:
+1

Zireael07:
Why is this not merged?

ivadim:
+1

cthill:
+1

storrgie:
Need this please

Zireael07:
What is the reason this is not merged?

joshaber:
I'm not sure this is the right place for this fix. Ideally we'd make whatever changes are necessary in `GitRepository` or `GitRepositoryAsync` in Atom itself, so that all users of `GitRepository*` can take advantage of these fixes.

nabeelomer:
Why aren't you merging this?

joshaber:
See my comment above: https://github.com/atom/tree-view/pull/469#issuecomment-172080867

iDev0urer:
Definitely a feature that's needed. Can we get the ball rolling on this one?

jarig:
I've rebased all changes again and fixed all tests and linting. So now it's fully green.

Zireael07:
Hopefully this gets merged soon.

lee-dohm:
Thanks for all the work that has been invested in this pull request.

In our weekly triage meeting, we discussed this change and have decided to not accept it. Besides the caveats mentioned at the top, we don't feel that this change has been validated against a wide enough variety of use cases to not have hidden pitfalls. And because [there have been significant usability impacts around this functionality before](https://github.com/atom/github/issues/1835#issuecomment-505191660), we can't take the risk of unleashing this untested on the entire user base.

Thanks again for the feedback and the work that you've put into this change.

jarig:
> And because there have been significant usability impacts around this functionality before

Is that real pitfall or theoretical? It never happened in real life, as current algo respects project boundaries.

PR #478, opened 2015-06-14
Add drag-and-drop feature

abm0:
Added drag-and-drop feature that uses native HTML5 drag and drop API.

animation

### DIFF
--- a/lib/directory-view.coffee
+++ b/lib/directory-view.coffee
@@ -35,6 +35,7 @@ class DirectoryView extends HTMLElement
     @directoryName.textContent = @directory.name
     @directoryName.dataset.name = @directory.name
     @directoryName.dataset.path = @directory.path
+    @directoryName.draggable = true
 
     if @directory.isRoot
       @classList.add('project-root')
--- a/lib/file-view.coffee
+++ b/lib/file-view.coffee
@@ -14,6 +14,7 @@ class FileView extends HTMLElement
     @fileName.textContent = @file.name
     @fileName.dataset.name = @file.name
     @fileName.dataset.path = @file.path
+    @fileName.draggable = true
 
     if @file.symlink
       @fileName.classList.add('icon-file-symlink-file')
--- a/lib/tree-view.coffee
+++ b/lib/tree-view.coffee
@@ -3,6 +3,7 @@ shell = require 'shell'
 
 _ = require 'underscore-plus'
 {BufferedProcess, CompositeDisposable} = require 'atom'
+{repoForPath, relativizePath} = require "./helpers"
 {$, View} = require 'atom-space-pen-views'
 fs = require 'fs-plus'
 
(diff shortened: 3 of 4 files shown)


styles/tree-view.less
```diff
@@ -59,3 +59,7 @@
     }
   }
 }
+
+.name.dragenter {
+  color: #949292;
```
mnquintana: Since this is a hard-coded color, it won't change to reflect the user's active UI theme. Would you mind changing this to [use one of the UI variables](https://atom.io/docs/latest/hacking-atom-creating-a-theme#variables)?

styles/tree-view.less
```diff
@@ -59,3 +59,7 @@
     }
   }
 }
+
+.name.dragenter {
```
mnquintana: This selector is in the global scope, so this will affect anything with the classes `.name` and `.dragenter` – can you change this to at least be scoped inside `.tree-view`? Ideally the selector should probably be `.tree-view .entry .name`, but nesting it inside the existing `.tree-view` block would be best.

lib/tree-view.coffee
```diff
@@ -755,3 +784,49 @@ class TreeView extends View
...
+    e.target.classList.add('dragenter')
+
+  onDragLeave: (e) ->
+    e.stopPropagation()
+
+    e.target.classList.remove('dragenter')
+
+  # Public: Handle entry name object dragstart event
+  #
+  # returns noop
```
mnquintana: You don't need to specify the return value if it's noop

abm0: If you mean the documentation, I've just made it as in neighbour functions. I can remove those "return" lines if it's critical

mnquintana:
Would you mind adding specs for this? Thanks!

lib/directory-view.coffee
```diff
@@ -35,6 +35,7 @@ class DirectoryView extends HTMLElement
     @directoryName.textContent = @directory.name
     @directoryName.dataset.name = @directory.name
     @directoryName.dataset.path = @directory.path
+    @directoryName.draggable = true
```
mnquintana: It would be great if it were the `.list-item` parent element that were draggable, rather than the name - currently, you have to have the cursor over the text, otherwise the item can't be dragged, and if you're dragging it to a folder, it has to be dropped on the folder's text.

lib/file-view.coffee
```diff
@@ -14,6 +14,7 @@ class FileView extends HTMLElement
     @fileName.textContent = @file.name
     @fileName.dataset.name = @file.name
     @fileName.dataset.path = @file.path
+    @fileName.draggable = true
```
mnquintana: It would be great if it were the .list-item parent element that were draggable, rather than the name - currently, you have to have the cursor over the text, otherwise the item can't be dragged, and if you're dragging it to a folder, it has to be dropped on the folder's text.

mnquintana:
Currently, list items can only be dragged and dropped within the Tree View - what do you think about making it so they can be dropped into a pane and opened in that pane? And dropped into a new tab?

lib/directory-view.coffee
abm0: Nesting of `.list-item` block differs in directory and file elements, so if I put draggable on it, I'll have to add some messy if-s to get initial and target paths. I can simply add `display: block` to `.name` block, so the draggable area would be rendered up to tree-view right border (attached screenshot). What do you think?

preview2

abm0:
> what do you think about making it so they can be dropped into a pane and opened in that pane? And dropped into a new tab?

Sure, I think, I can implement that

Issue #251, opened 2014-11-09
ctrl-0 clashes with hotkey for reset font size


CheeseSucker:
The tree view seems to take precedence, so this is only a minor inconvenience when you want to reset font size. There are of course workarounds:

1) Use mousewheel or ctrl + and ctrl - to find appropriate size
2) Use command palette
3) Rebind on user level

jlogar:
Binding he key <kbd>ctrl</kbd>+<kbd>0</kbd> to 'window:reset-font-size' doesn't seem to cut it in version 0.179.0. Using windows. It just focuses the tree view.

bentolor:
Also stumbled over this issue: Atom Menu announces Ctrl-0 under "View > Reset Font Size". 

I think the shortcut is more widely used to reset font-size than focusing parts of the UI (IMHO Alt-1..9 are common shortcuts for that)

ardrigh:
Is this being worked on? It's still an issue in 0.290.0 on Windows.

I use CTRL+0 to reset my fonts in a number of apps, but it is broken in Atom due to the tree view focus overriding it.

mnquintana:
I just fixed this in https://github.com/atom/tree-view/pull/479, but I'm not sure about the keybinding – I changed it to <kbd>Alt-0</kbd> – what do people think about that for switching focus?

awayken:
:+1: That works for me, @mnquintana.

PR #479, opened 2015-06-14
Fix keybinding conflict with window:reset-font-size

mnquintana:
Fixes https://github.com/atom/tree-view/issues/251

### DIFF
--- a/keymaps/tree-view.cson
+++ b/keymaps/tree-view.cson
@@ -8,7 +8,7 @@
   'ctrl-\\': 'tree-view:toggle'
   'ctrl-k ctrl-b': 'tree-view:toggle'
   'ctrl-|': 'tree-view:reveal-active-file'
-  'alt-0': 'tree-view:toggle-focus'
+  'alt-\\': 'tree-view:toggle-focus'
 
 '.platform-darwin .tree-view':
   'cmd-c': 'tree-view:copy'
--- a/keymaps/tree-view.cson
+++ b/keymaps/tree-view.cson
@@ -8,7 +8,7 @@
   'ctrl-\\': 'tree-view:toggle'
   'ctrl-k ctrl-b': 'tree-view:toggle'
   'ctrl-|': 'tree-view:reveal-active-file'
-  'ctrl-0': 'tree-view:toggle-focus'
+  'alt-0': 'tree-view:toggle-focus'
 
 '.platform-darwin .tree-view':
   'cmd-c': 'tree-view:copy'


mnquintana:
@atom/feedback What do you think about the keybinding?

lee-dohm:
I don't have a problem with it, though I am concerned about changing a default that has existed in Atom for over a year. This should be well documented and communicated.

50Wliu:
> I am concerned about changing a default that has existed in Atom for over a year

To be honest I don't think anyone on Windows/Linux really expected this behavior.  It's pretty universal to have <kbd>Ctrl+0</kbd> bound to resetting the zoom size.

mnquintana:
Aside from the fact that the keybinding's changed, what do you think about the keybinding I chose to replace it? Does <kbd>Alt-0</kbd> seem reasonable / idiomatic for Win / Linux?

lee-dohm:
It's been a few years since I've used Windows. I would typically think of <kbd>Tab</kbd> or <kbd>Ctrl+Tab</kbd> for moving around sub-views in Windows programs. I'd check to see if Visual Studio has some key binding for moving between the Tree View, Edit View and others.

lee-dohm:
According to [this list of Visual Studio 2015 key shortcuts](http://visualstudioshortcuts.com/2015/) the "Next View" accelerator is <kbd>Ctrl+Page Down</kbd>. But if you look at the Team and Team Foundation sections, they use <kbd>Ctrl+0</kbd> as the beginning of a multi-key accelerator for lots of "switch to other views" things. <kbd>Alt+0</kbd> is used for a single, obscure function.

ardrigh:
Tree View currently uses the following:

<kbd>Ctrl+\\</kbd> for `Toggle` 
<kbd>Ctrl+Shift+\\</kbd> for `Reveal Active File`

Would <kbd>Alt+\\</kbd> be a better key binding for `Focus` than <kbd>Alt+0</kbd>?

I can't see <kbd>Alt+\\</kbd> used for anything else, unless another package is using it

mnquintana:
So after reviewing editors and IDEs on Windows, I've come to the conclusion that there is no consistent keybinding for this kind of behavior. So in this case, I'm going to go with whatever feels most intuitive for *Atom* users - I think the two candidates are <kbd>Alt+\</kbd> and <kbd>Alt+0</kbd>. The former preserves one key from the current keybinding, but the latter seems more consistent with other Tree View keybindings. What do people think about these two options?

mnquintana:
Alright, the new keybinding is going to be <kbd>Alt-\</kbd> since it seems the most consistent with other Tree View keybindings. Thanks everyone for the feedback!

Issue #484, opened 2015-06-18
Modified status (VCS) not updated for directories when repo is symlink


acontreras89:
*Modified* status (VCS) is not updated for directories when the project is a symbolic link. Steps to reproduce:
* Launch Atom from inside a symbolic link to a repository
* Modify a file inside a directory with no VCS status
* Use command `Core: Save`

... Directory's VCS status should be set, but it is not...
* Use command `Windows: Reload`
* Modify the file again as to clear its VCS status (manually undo the changes)

... Directory's VCS status should be cleared, but again, it is not.

System info:
* Ubuntu 14.04 LTS
* Atom 0.210.0
* tree-view 0.172.0

acontreras89:
@mnquintana, @benogle, can you guys take a look at this?
I also created https://github.com/atom/tree-view/pull/487 which solved the problem for me.

izuzak:
This is semi-related to https://github.com/atom/tree-view/issues/69, see https://github.com/atom/tree-view/issues/69#issuecomment-59159891

kankaristo:
I'm able to reproduce this, and as @acontreras89 said, this could be related to https://github.com/atom/tabs/issues/182.

PR #487, opened 2015-06-19
Update VCS status properly for directories with symlink parents

acontreras89:
Fix https://github.com/atom/tree-view/issues/484

### DIFF
--- a/lib/directory.coffee
+++ b/lib/directory.coffee
@@ -64,7 +64,7 @@ class Directory
     return unless repo?
 
     @subscriptions.add repo.onDidChangeStatus (event) =>
-      @updateStatus(repo) if event.path.indexOf("#{@path}#{path.sep}") is 0
+      @updateStatus(repo) if event.path.indexOf("#{@realPath}#{path.sep}") is 0
     @subscriptions.add repo.onDidChangeStatuses =>
       @updateStatus(repo)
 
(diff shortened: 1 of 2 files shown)


lib/directory.coffee
```diff
@@ -64,7 +64,7 @@ class Directory
     return unless repo?
 
     @subscriptions.add repo.onDidChangeStatus (event) =>
-      @updateStatus(repo) if event.path.indexOf("#{@path}#{path.sep}") is 0
+      @updateStatus(repo) if event.path.indexOf("#{@realPath}#{path.sep}") is 0
```
acontreras89: Should we keep the old condition as well? For files (which are working fine) both `@path` and `@realPath` are checked ([here](https://github.com/atom/tree-view/blob/master/lib/file.coffee#L53) and [here](https://github.com/atom/tree-view/blob/master/lib/file.coffee#L76).)
I haven't seen `repo`'s `onDidChangeStatus` event containing a path with a symlink directory, but I guess that would be a little safer to do.

mnquintana:
This looks good, but would you mind adding a spec to help prevent regressions in the future? Thanks!

acontreras89:
:+1: There you go @mnquintana

acontreras89:
Hey @mnquintana , could you take a look at this again? :smile:

acontreras89:
It'd be nice if someone could review this :)

mnquintana:
Sorry about that! Been super busy - I'll take a look tonight.

mnquintana:
Awesome, thanks so much for fixing this! :rocket:

acontreras89:
My pleasure!

PR #490, opened 2015-06-20
Add a file icon service

lee-dohm:
This is intended to allow for a more flexible customization of the icons used for files (though it could be extended to folders too). A package could provide a service that is called to determine the icon class to associate with a file in the tree view and then provide styles to display whatever icon they wish.

See the [file-icons package](https://atom.io/packages/file-icons), the [file-type-icons package](https://atom.io/packages/file-type-icons) as well as a number of UI themes that perform this kind of customization. They currently depend on parsing the path, where this is problematic is that there is sometimes a disconnect between the grammar that Atom uses for a file and the type of file that the icon says it is.

This could also allow for more complex logic like recognizing that a project is a Rails project rather than standard Ruby and changing some of the icons.

### DIFF
--- a/lib/default-file-icons.coffee
+++ b/lib/default-file-icons.coffee
@@ -1,0 +1,23 @@
+fs = require 'fs-plus'
+path = require 'path'
+
+class DefaultFileIcons
+  getIconForPath: (filePath) ->
+    extension = path.extname(filePath)
+
+    if fs.isSymbolicLinkSync(filePath)
+      'icon-file-symlink-file'
+    else if fs.isReadmePath(filePath)
+      'icon-book'
+    else if fs.isCompressedExtension(extension)
+      'icon-file-zip'
+    else if fs.isImageExtension(extension)
+      'icon-file-media'
+    else if fs.isPdfExtension(extension)
+      'icon-file-pdf'
+    else if fs.isBinaryExtension(extension)
+      'icon-file-binary'
+    else
+      'icon-file-text'
+
+module.exports = DefaultFileIcons
--- a/lib/default-file-icons.coffee
+++ b/lib/default-file-icons.coffee
@@ -2,7 +2,7 @@ fs = require 'fs-plus'
 path = require 'path'
 
 class DefaultFileIcons
-  getIconForPath: (filePath) ->
+  iconClassForPath: (filePath) ->
     extension = path.extname(filePath)
 
     if fs.isSymbolicLinkSync(filePath)
--- a/package.json
+++ b/package.json
@@ -23,7 +23,7 @@
     "coffeelint": "^1.9.7"
   },
   "consumedServices": {
-    "tree-view.file-icons": {
+    "atom.file-icons": {
       "versions": {
         "1.0.0": "consumeFileIcons"
       }
--- a/package.json
+++ b/package.json
@@ -23,7 +23,7 @@
     "coffeelint": "^1.9.7"
   },
   "consumedServices": {
-    "file-icons": {
+    "tree-view.file-icons": {
       "versions": {
         "1.0.0": "consumeFileIcons"
       }
--- a/package.json
+++ b/package.json
@@ -21,5 +21,12 @@
   },
   "devDependencies": {
     "coffeelint": "^1.9.7"
+  },
+  "consumedServices": {
+    "file-icons": {
+      "versions": {
+        "1.0.0": "consumeFileIcons"
+      }
+    }
   }
 }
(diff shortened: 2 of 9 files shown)


forivall:
:+1:

This would be great since extensions could call native apis, which can even return a default icon image file for a given file - gtk/gnome: https://lazka.github.io/pgi-docs/#Gio-2.0/classes/FileIcon.html#Gio.FileIcon.get_file (same api in gjs/seed)

I would urge you to consider the option to allow services to return file names in addition to css classes.

thomasjo:
This looks great. Let's :ship:? At least get a discussion going to avoid this silently dying. /cc @atom/feedback

benogle:
Cool

lib/default-file-icons.coffee
```diff
@@ -0,0 +1,23 @@
+fs = require 'fs-plus'
+path = require 'path'
+
+class DefaultFileIcons
+  getIconForPath: (filePath) ->
```
benogle: Should this be named something to do with a css class?

package.json
```diff
@@ -21,5 +21,12 @@
   },
   "devDependencies": {
     "coffeelint": "^1.9.7"
+  },
+  "consumedServices": {
+    "file-icons": {
```
maxbrunsfeld: I think we should maybe call the service `tree-view.file-icons`, to promote the convention of using package names as service namespaces.

lib/default-file-icons.coffee
maxbrunsfeld: Yeah, I like that idea. Maybe `iconClassForPath`? It seems like we usually don't put `get` in the method name when it has the pattern `{thing}For{OtherThing}`.

lee-dohm: Done.

package.json
lee-dohm: Done.

simurai:
Should there also be a file icon service for the tabs? I think most icon packages add icons to both, tree-view and tabs.

lee-dohm:
Ah, that's a good point, @simurai. It's been a while since I worked on this. The original idea was for both the tabs package and the tree-view package to consume the same service. In that case, would it be better to keep the name of the service as `file-icons`? Or does it still work as `tree-view.file-icons` it just happens that tabs also consumes it?

lee-dohm:
@atom/feedback Ping? Should I leave the name of the service as `tree-view.file-icons`? Or should it go back to `file-icons`?

maxbrunsfeld:
> most icon packages add icons to both, tree-view and tabs.

Yeah, that is a good point; I didn't realize that. Yeah, `file-icons` is probably better then, or maybe `atom.file-icons`.

thomasjo:
> Yeah, that is a good point; I didn't realize that. Yeah, `file-icons` is probably better then, or maybe `atom.file-icons`.

`core.file-icons`?

maxbrunsfeld:
We've got a [couple](https://github.com/atom/atom/blob/master/src/workspace.coffee#L56) [of](https://github.com/atom/atom/blob/master/src/project.coffee#L49) `atom`-prefixed services already, so I'd say we should probably go with that over `core`.

thomasjo:
> We've got a couple of atom-prefixed services already, so I'd say we should probably go with that over core.

Agreed :+1:

benogle:
We good on this?

maxbrunsfeld:
:+1: @lee-dohm Are you interested in updating `tabs` so that it consumes this service?

lee-dohm:
Yes, I can take care of that later tonight.

maxbrunsfeld:
:ship:?

maxbrunsfeld:
:metal: This is nice @lee-dohm.

lee-dohm:
Thanks! I'm looking forward to updating my [file-type-icons package](https://atom.io/packages/file-type-icons) to support it :grinning:

benogle:
:+1:

hgGeorg:
There is an issue with this, at least on the windows version: when you open a folder using the right-click context menu, then the service won't be called and everything just uses the default icons until the treeview is re-expanded.

hgGeorg:
Also there seems to be no way of adding multiple classes. Returning `'first second'` throws an InvalidCharacterError because of the space character. Returning an array `['first', 'second']` will result in the class `first,second` to be added instead.

lee-dohm:
@hgGeorg The PR has already been merged. If there are bugs or suggestions for enhancements, please open new Issues.

hgGeorg:
@lee-dohm My bad. Opened two separate issues: #652 and #653

PR #492, opened 2015-06-23
Consistent behavior of HOME and END keys

monoblaine:
When we press the HOME key, the focus should be set on the topmost (or the bottommost, when END is pressed) file in the tree view. All the IDEs I've ever used behave this way.


kevinsawicki:
Thanks for adding this :+1:

PR #499, opened 2015-07-01
Fix cannot read property 'getPath' of null issue

jwuwork:
This occurs when running unit tests with `apm test`.


kevinsawicki:
I believe this was fixed by https://github.com/atom/tree-view/commit/81d575cbe5ada3182b89b6b60ef22b649337b7d7 and should no longer be an issue on latest master, please comment if it is.

PR #500, opened 2015-07-01
Fix unit test for scrolling file into view to use fuzzy logic

jwuwork:
Should probably create a better unit test.  But in the mean time, this should help fix the scroll file into view test so that tests pass both on Travis and when running the tests locally.


thomasjo:
This seems reasonable to me.

/cc @kevinsawicki This resolves the issue I pinged you on with regards to v0.174.0.

kevinsawicki:
Thanks!

Issue #196, opened 2014-07-24
.gitignore cascaded negations not working correctly


rafitorres:
Let's say I add the following rules to my `.gitignore` to ignore everything in the root except the `wp-content` directory:

```
/*
!.gitignore
!wp-content/
```

So far this works correctly (only `.gitignore` and `wp-content/` are shown in the tree view). Now I add the following rules to ignore everything in the `wp-content` directory, except the `plugins` and `themes` directories

```
wp-content/*
!wp-content/plugins/
!wp-content/themes/
```

Before Atom 0.116.0, the plugins and themes directories were shown correctly in the tree view, but now the `wp-content` directory disappears completely. My machine's git (v2.0.0) parses the .gitignore correctly and sees changes in `plugins` and `themes`.

I see that 0.116.0 moved to libgit2 0.21.0. If this is what's causing this behavior I'll open an issue with them.

kylehotchkiss:
+1, still getting this in latest version ( atom 0.169.0 / tree view 0.147.0 )

50Wliu:
I think I may be having the opposite problem (or it could just be that my gitignore skills are bad).
I have the following in my `.gitignore` file:
```
addons/sourcemod/scripting/**
addons/sourcemod/scripting/include/**
!addons/sourcemod/scripting/include/
!addons/sourcemod/scripting/include/freak_fortress_2*.inc
```
`freak_fortress_2.inc` etc. show up properly and respond to changes correctly, but any other file in `/include/` shows up as newly added (aka green).

![Oh noes](https://cloud.githubusercontent.com/assets/2766036/5846755/554e9c26-a194-11e4-90df-b33e4e615d45.png)

The full `.gitignore` file is available [here](https://raw.githubusercontent.com/50DKP/FF2-Official/stable/.gitignore).

irobert91:
Ubuntu 14.04 LTS, tree-view 0.162.0, atom 0.184.0
![atom_tree](https://cloud.githubusercontent.com/assets/5888095/6554540/0e877fc2-c65d-11e4-9ba1-cc1fc2d1cc02.png)

stramel:
I believe this is related to this issue. I have a git ignored folder with a negated git ignore folder inside it. When clicking `Hide VCS Ignored Files` it hides the highest ignored file and all of its contents. I don't think it should be hiding that folder if it contains project files still.

Showing VCS Ignored Files:
![Shown VCS Files](https://imgur.com/0eUvZP9)

[...]

PR #524, opened 2015-07-17
Tries to solve issue #196

albertein:
This is not meant to be accepted as it is as the fix for issue #196 but it's meant to open a discussion
on how it should be properly fixed.

The problem is that `lib/directory.coffe` gets loaded lazily for every folder currently expanded on the ui,
doing it so means it'st not possible to know if there is any subfolder or subfile at any level which is not ignored
and hence should be shown.

This commits does a deep first search on all subdirectories when enumerating a directory on `directory/getEntries`.
If there is at least one non ignored file on any child on any level the directory is not ignored.

Please note that specs are not included yet since this might not be the best way to solve the issue.

### DIFF
--- a/lib/directory.coffee
+++ b/lib/directory.coffee
@@ -155,11 +155,37 @@ class Directory
           when 'change' then @reload()
           when 'delete' then @destroy()
 
-  getEntries: ->
+  getChilds: ->
     try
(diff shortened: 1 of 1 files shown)


pablasso:
Anyone else has an opinion on this? how can we properly fix this?

PR #525, opened 2015-07-18
Allow reordering project folders via drag and drop

forivall:
Fixes https://github.com/atom/tree-view/issues/691

This uses the same logic that moving tabs uses to reposition project folders within the tree view.

move-project-folders

* [x] implement
* [x] update to interoperate with file movement dragging
* [x] fix placeholder width
* [x] add tests

Future additions:
* reduce amount of duplicated code - needs coordination with tabs-view and drag-and-drop file moving
* allow dragging in folders from the filesystem - needs electron support first, unfortunately. Or it's just broken in kwin.

### Bugfixes

* [x] When dragging project folders, don't allow movement of the folder itself
* [x] Allow movement below last project folder (https://github.com/atom/tree-view/pull/525#issuecomment-252341762) - should also cover when dragging to an empty tree view
* [x] Make sure placeholder (blue bar) is visible when dragging above first project (https://github.com/atom/tree-view/pull/525#issuecomment-252702782)

### DIFF
--- a/lib/project-folder-dnd-handler.coffee
+++ b/lib/project-folder-dnd-handler.coffee
@@ -7,11 +7,7 @@ _ = require 'underscore-plus'
 module.exports =
 class ProjectFolderDragAndDropHandler
   constructor: (@treeView) ->
-    @treeView.on 'dragstart', '.project-root-header', @onDragStart
     @treeView.on 'dragend', '.project-root-header', @onDragEnd
-    @treeView.on 'dragleave', @onDragLeave
-    @treeView.on 'dragover', @onDragOver
-    @treeView.on 'drop', @onDrop
 
     RendererIpc.on('tree-view:project-folder-dropped', @onDropOnOtherWindow)
 
@@ -20,11 +16,11 @@ class ProjectFolderDragAndDropHandler
     RendererIpc.removeListener('tree-view:project-folder-dropped', @onDropOnOtherWindow)
 
   onDragStart: (event) =>
-    event.originalEvent.dataTransfer.setData 'atom-event', 'true'
+    unless $(event.target).closest('.project-root-header').size()
+      return
 
-    element = $(event.target).closest('.project-root-header')
+    event.originalEvent.dataTransfer.setData 'atom-event', 'true'
     projectRoot = $(event.target).closest('.project-root')
-    event.originalEvent.dataTransfer.setDragImage(projectRoot[0], 0, 0)
     directory = projectRoot[0].directory
 
     event.originalEvent.dataTransfer.setData 'project-root-index', projectRoot.index()
@@ -56,12 +52,13 @@ class ProjectFolderDragAndDropHandler
 
   onDragOver: (event) =>
     unless event.originalEvent.dataTransfer.getData('atom-event') is 'true'
-      event.preventDefault()
-      event.stopPropagation()
       return
 
-    event.preventDefault()
-    event.stopPropagation()
+    entry = event.currentTarget
+    if entry.classList.contains('selected')
+      @clearDropTarget()
+      return
+
     newDropTargetIndex = @getDropTargetIndex(event)
     return unless newDropTargetIndex?
 
@@ -148,6 +145,17 @@ class ProjectFolderDragAndDropHandler
     else
       projectRoots.index(element) + 1
 
+  isMovingProjectFolders: (event) ->
+    target = $(event.target)
+    return null if @isPlaceholder(target)
+
+    element = target.closest('.project-root-header')
+    return false unless element.length
+
+    elementCenter = element.offset().top + element.height() / 2
+    return true if event.originalEvent.pageY < elementCenter
+
+    false
 
   getPlaceholder: ->
     @placeholderEl ?= $('<li/>', class: 'placeholder')
--- a/lib/project-folder-dnd-handler.coffee
+++ b/lib/project-folder-dnd-handler.coffee
@@ -1,0 +1,174 @@
+BrowserWindow = null # Defer require until actually used
+RendererIpc = require 'ipc'
+
+{$, View} = require 'atom-space-pen-views'
+_ = require 'underscore-plus'
+
+module.exports =
+class ProjectFolderDragAndDropHandler
+  constructor: (@treeView) ->
+    @treeView.on 'dragstart', '.project-root-header', @onDragStart
+    @treeView.on 'dragend', '.project-root-header', @onDragEnd
+    @treeView.on 'dragleave', @onDragLeave
+    @treeView.on 'dragover', @onDragOver
+    @treeView.on 'drop', @onDrop
+
+    RendererIpc.on('tree-view:project-folder-dropped', @onDropOnOtherWindow)
+
+  # unused
+  unsubscribe: ->
+    RendererIpc.removeListener('tree-view:project-folder-dropped', @onDropOnOtherWindow)
+
+  onDragStart: (event) =>
+    event.originalEvent.dataTransfer.setData 'atom-event', 'true'
+
+    element = $(event.target).closest('.project-root-header')
+    projectRoot = $(event.target).closest('.project-root')
+    event.originalEvent.dataTransfer.setDragImage(projectRoot[0], 0, 0)
+    directory = projectRoot[0].directory
+
+    event.originalEvent.dataTransfer.setData 'project-root-index', projectRoot.index()
+
+    rootIndex = -1
+    _.find(@treeView.roots, (root, index) -> root.directory is directory and ((rootIndex = index) or true))
+    event.originalEvent.dataTransfer.setData 'from-root-index', rootIndex
+    event.originalEvent.dataTransfer.setData 'from-root-path', directory.path
+    event.originalEvent.dataTransfer.setData 'from-process-id', @getProcessId()
+    event.originalEvent.dataTransfer.setData 'from-routing-id', @getRoutingId()
+
+    event.originalEvent.dataTransfer.setData 'text/plain', directory.path
+
+    if process.platform is 'darwin'
+      pathUri = "file://#{directory.path}" unless @uriHasProtocol(directory.path)
+      event.originalEvent.dataTransfer.setData 'text/uri-list', pathUri
+
+  uriHasProtocol: (uri) ->
+    try
+      require('url').parse(uri).protocol?
+    catch error
+      false
+
+  onDragLeave: (event) =>
+    @removePlaceholder()
+
+  onDragEnd: (event) =>
+    @clearDropTarget()
+
+  onDragOver: (event) =>
+    unless event.originalEvent.dataTransfer.getData('atom-event') is 'true'
+      event.preventDefault()
+      event.stopPropagation()
+      return
+
+    event.preventDefault()
+    event.stopPropagation()
+    newDropTargetIndex = @getDropTargetIndex(event)
+    return unless newDropTargetIndex?
+
+    @removeDropTargetClasses()
+
+    projectRoots = $(@treeView.roots)
+
+    if newDropTargetIndex < projectRoots.length
+      element = projectRoots.eq(newDropTargetIndex).addClass 'is-drop-target'
+      @getPlaceholder().insertBefore(element)
+    else
+      element = projectRoots.eq(newDropTargetIndex - 1).addClass 'drop-target-is-after'
+      @getPlaceholder().insertAfter(element)
+
+  onDropOnOtherWindow: (fromItemIndex) =>
+    paths = atom.project.getPaths()
+    paths.splice(fromItemIndex, 1)
+    atom.project.setPaths(paths)
+
+    @clearDropTarget()
+
+  clearDropTarget: ->
+    element = @treeView.find(".is-dragging")
+    element.removeClass 'is-dragging'
+    element[0]?.updateTooltip()
+    @removeDropTargetClasses()
+    @removePlaceholder()
+
+  onDrop: (event) =>
+    event.preventDefault()
+    {dataTransfer} = event.originalEvent
+
+    return unless dataTransfer.getData('atom-event') is 'true'
+
+    fromProcessId = parseInt(dataTransfer.getData('from-process-id'))
+    fromRoutingId = parseInt(dataTransfer.getData('from-routing-id'))
+    fromRootPath  = dataTransfer.getData('from-root-path')
+    fromIndex     = parseInt(dataTransfer.getData('project-root-index'))
+    fromRootIndex = parseInt(dataTransfer.getData('from-root-index'))
+
+    toIndex = @getDropTargetIndex(event)
+
+    @clearDropTarget()
+
+    if fromProcessId is @getProcessId()
+      unless fromIndex is toIndex
+        projectPaths = atom.project.getPaths()
+        projectPaths.splice(fromIndex, 1)
+        if toIndex > fromIndex then toIndex -= 1
+        projectPaths.splice(toIndex, 0, fromRootPath)
+        atom.project.setPaths(projectPaths)
+    else
+      projectPaths = atom.project.getPaths()
+      projectPaths.splice(toIndex, 0, fromRootPath)
+      atom.project.setPaths(projectPaths)
+
+      if not isNaN(fromProcessId) and not isNaN(fromRoutingId)
+        # Let the window where the drag started know that the tab was dropped
+        browserWindow = @browserWindowForProcessIdAndRoutingId(fromProcessId, fromRoutingId)
+        browserWindow?.webContents.send('tree-view:project-folder-dropped', fromIndex)
+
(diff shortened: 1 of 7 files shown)


alex88:
+1 this would be  great, specially with a lot of folders not alphabetically ordered

forivall:
I've updated this to work with the drag and drop changes in master. The semantics are that project folder dragging will occur if the drop occurs over the top half of a project folder entry, or over a file entry. The placeholder will appear if and only if project folder movement will occur, otherwise, selection will appear.

Alternate semantics that I've thought about are that file movement should only occur if the user is holding down the shift key, similar to many file managers.

brunogaspar:
+1

lib/helpers.coffee
```diff
@@ -15,3 +15,11 @@ module.exports =
       camelizedAttr = property.replace /\-([a-z])/g, (a, b) -> b.toUpperCase()
       styleObject[camelizedAttr] = value
     styleObject
+
+  ensureOpaqueBackground: (styleObject) ->
```
50Wliu: What is this for?

lib/project-folder-dnd-handler.coffee
```diff
@@ -0,0 +1,178 @@
...
+      return
+
+    event.originalEvent.dataTransfer.setData 'atom-event', 'true'
+    projectRoot = $(event.target).closest('.project-root')
+    directory = projectRoot[0].directory
+
+    event.originalEvent.dataTransfer.setData 'project-root-index', projectRoot.index()
+
+    rootIndex = -1
+    _.find(@treeView.roots, (root, index) -> root.directory is directory and ((rootIndex = index) or true))
```
50Wliu: `(rootIndex = index) or true)` I might be missing something, but won't that always evaluate to true?

lib/helpers.coffee
forivall: It's from this PR: https://github.com/atom/tree-view/pull/565 ; I accidentally merged it in at some point. I can rebase it out if needed.

lib/project-folder-dnd-handler.coffee
forivall: It's an overly clever way to do `_.first`, while also returning the index. I should replace it with a for loop, since that would make it clearer.

50Wliu:
Can you please rebase on top of master so that only the changes specific to this PR get displayed in the diff?  It gets confusing reviewing when there's changes from other, unrelated PRs.

forivall:
@50Wliu Rebased. The other PR isn't 100% unrelated, but the fix is disjoint. :wink:

marjakm:
+1

forivall:
Rebased and updated. Tests all work.

@50Wliu ping!

mxstbr:
Is there a particular reason this wasn't merged yet? I'd love to have this feature.

krzysztof-sikorski:
Please make this behavior optional (toggled by a setting). Too many times I had accidentally dragged an entry instead of clicking on it.

anothernode:
@mxstbr I guess someone with the proper authorization needs to review and approve the code (there's a "needs-review" label attached).

thedaniel:
Sorry about the delay. I linked this locally to test and found this issue on my first action: if you open the editor with a single project, and drag that project and drop it onto one of its own subfolders, an error is displayed. We need to not only do nothing on drop in this case, we shouldn't be highlighting / selecting tree view entries as the project is dragged over them.

mxstbr:
@forivall and update to the above bug report?

forivall:
@mxstbr haven't had time yet, was focused on #582

heruan:
Is this still under review? I'd love to have it on the next release!

flcoder:
I really want this... I haven't looked at the code.  Should I bother attempting to work on this or is resolution coming soon?

forivall:
@flcoder My bad, I'll update this within a week. Hopefully tonight.

forivall:
@thedaniel that's actually an issue inherited from the drag-drop used to move files/folders. I'll create a separate PR to fix that, since I understand the drag-drop code, but it's unrelated to the issue of rearranging project folders.

There is the question of: should project folders be movable into other project folders or not? Let me know if y'all need a more in depth explanation.

garretwilson:
> should project folders be movable into other project folders or not?

Maybe they should. Maybe they shouldn't. But let's not complicate things. We all agree we should be able to rearrange the "projects" in Atom. I say put the rest in another issue. If you're not careful, there will be a big back-and-forth about the "extra" features, and another year will pass before we get even the part that everybody is in agreement on.

thedaniel:
> should project folders be movable into other project folders or not?

I don't think so personally,  nested projects don't exist now and whether to add them or not is a decision to be made outside of the tree-view repo, and shouldn't hold up this work.

forivall:
Just to clarify, by movable, I mean moving that actual folder on disk.

Ben3eeE:
If we allow project folders to be movable (on disk) into other project folders wouldn't that just create a lot of accidental project moving when this gets merged? I think it should be disallowed.

forivall:
@Ben3eeE note that that's the current behaviour without this feature. The file moving via drag and drop was added after I started this pr, but I didn't receive any feedback  about overriding it (see comment above)

Ben3eeE:
@atom/feedback any thoughts about last few comments?

iolsen:
> @atom/feedback any thoughts about last few comments?

I agree with the emerging sentiment that we should move forward on this PR without complicating it with the project-folder-moving-on-disk issue. Let's debate that separately, if the debate still wants to happen.

iolsen:
Attempting to summarize a Slack conversation I just had with @Ben3eeE:

- We should merge this PR with strictly the reordering of projects functionality. 
- Dropping projects inside other projects, probably moving them on disk, may or may not happen but we won't consider it for this PR.
- To that end, we can't merge this until dropping projects inside other projects is handled better. It should simply be disallowed. #952 might address that. If it doesn't it should be addressed here.

Ben3eeE:
Found bug, if the last project folder is expanded in the `tree-view` the bar indicating that move is possible never disappears. But move never happens when dropping.

drag on last

forivall:
Turns out that my memory while writing https://github.com/atom/tree-view/pull/525#issuecomment-251858592 was incorrect. Project folders can't be moved into folders by default. I'll completely disable the movement of project folders as folders into other folders.

Sorry. :heart: 🇨🇦

Ben3eeE:
Two more:

1. If the target `tree-view` doesn't have any project folders at all you can't drop there.
2. The blue bar for the top `tree-view` folder seems to be above the entry causing it to be almost invisible.

forivall:
Fixed https://github.com/atom/tree-view/pull/525#issuecomment-252349495
IIRC, i thought that you could always move the root folder, because once I made root folders draggable, they were movable by the on-disk movement. But this wasn't enabled by the on-disk movement that was in core.

Header edited to track other bugfixes.

Ben3eeE:
With the latest commit we are not blocked by #952 to merge this. #952 is still something we should fix but not something we need to consider for this PR since it is a separate bug.

@forivall Let me know when you have fixed the bugs I reported and I will test once again.

jorgeorpinel:
Is this usable? I'd like to try merging this into my tree-view package. I've never done it before though, do I just need to find my local git repo and merge away?

forivall:
@Ben3eeE It's good to go!

@jorgeorpinel I've been doing that for the past year, since I first opened this PR! But it looks like this will finally get merged in soon. :tada:

Ben3eeE:
@forivall Looking good but I am having issues when dragging between windows tested on Atom `1.13.0-dev-4d26419` and `1.12.0-beta2`. 

drag folder changes folder

forivall:
I was about to say I couldn't replicate it, and then i realized that you specified beta. Can replicate, looking at it now.

forivall:
@Ben3eeE Should I make the change backwards compatible with the old `require('ipc')` or only use `{ipcRenderer} = require('electron')`? My initial fix will be backwards compatible.

forivall:
@Ben3eeE Fixed! Works on the current stable atom, and the beta, with no backwards-compat junk needed.

Ben3eeE:
@forivall Can't find any more bugs :+1:

lib/project-folder-dnd-handler.coffee
```diff
@@ -0,0 +1,168 @@
...
+    @placeholderEl = null
+
+  isPlaceholder: (element) ->
+    element.is('.placeholder')
+
+  getWindowId: ->
+    @processId ?= atom.getCurrentWindow().id
+
+  browserWindowForId: (id) ->
+    BrowserWindow ?= require('remote').require('browser-window')
```
50Wliu: I believe this should be `require('electron').remote. BrowserWindow`.

You can check by running the specs while using Atom Beta and making sure no Electron deprecations are logged.

forivall: Unfortunately, i was stubbing that method, so I didn't catch this in the unit tests.
Good catch, thanks!

forivall:
@50Wliu I've removed use of that deprecated api too.

BinaryMuse:
Thanks for the PR, @forivall! One concern I have right now is performance; dragging a project folder is causing a lot of style recalculation:

tree-view-safe-dnd-reg-plus-scroll

This is with other packages disabled; with my usual package set, certain events (like scroll events) lock up the main thread for up to 15 seconds.

I experimented with removing the placeholder:

```diff
diff --git a/lib/project-folder-dnd-handler.coffee b/lib/project-folder-dnd-handler.coffee
index 0d39b5a..e895553 100644
--- a/lib/project-folder-dnd-handler.coffee
+++ b/lib/project-folder-dnd-handler.coffee
@@ -42,7 +42,7 @@ class ProjectFolderDragAndDropHandler
       false
 
   onDragLeave: (event) =>
-    @removePlaceholder()
+    # @removePlaceholder()
 
   onDragEnd: (event) =>
     @clearDropTarget()
@@ -65,10 +65,10 @@ class ProjectFolderDragAndDropHandler
 
     if newDropTargetIndex < projectRoots.length
       element = projectRoots.eq(newDropTargetIndex).addClass 'is-drop-target'
-      @getPlaceholder().insertBefore(element)
+      # @getPlaceholder().insertBefore(element)
     else
       element = projectRoots.eq(newDropTargetIndex - 1).addClass 'drop-target-is-after'
-      @getPlaceholder().insertAfter(element)
+      # @getPlaceholder().insertAfter(element)
 
   onDropOnOtherWindow: (event, fromItemIndex) =>
     paths = atom.project.getPaths()
@@ -82,7 +82,7 @@ class ProjectFolderDragAndDropHandler
     element.removeClass 'is-dragging'
     element[0]?.updateTooltip()
     @removeDropTargetClasses()
-    @removePlaceholder()
+    # @removePlaceholder()
 
   onDrop: (event) =>
     event.preventDefault()
```

and it seems to help quite a bit:

tree-view-safe-dnd-without-placeholder

and scroll events don't lock up the thread any more.

Perhaps it's worth looking into another method for rendering the placeholder, and I can give it a closer look.

forivall:
@BinaryMuse Sure! I originally based the placeholder code from the placeholder code for rearranging tabs. I'll start by looking at if there are any new improvements there that I can port over, and otherwise have some fun with debugging layout perf.

BinaryMuse:
@forivall Great, thanks!

One other quick note: it looks like we should clean up the `ProjectFolderDragAndDropHandler` subscriptions in `TreeView#deactivate` so they don't leak when the package is deactivated (you can see the leak by running the specs; the EventEmitter module warns about it).

forivall:
I updated the css to be equivalent to the tabs css, and changed the events so that the placeholder isn't removed/re-added when it shouldn't be removed. This appeared to bring down the style recalculation to similar levels as the tabs extension.

Dragging tree-view roots:
tree-view-root-dragging

Dragging tabs:
tabs-dragging

I also call the cleanup function in deactivate.

Let me know if it still has issues for you. Any further improvement would be experimentation with the web animation api. :sunglasses:

peterschmidler:
any progress on this? missing it badly ;)

Ben3eeE:
All project folders collapsed and it seems fine. Expand one and 💥 
drag root 1-2

BinaryMuse:
@forivall @Ben3eeE This seems to be because we're calling `@getPlaceholder().insertBefore/After(element)` over and over again from `RootDragAndDropHandler#onDragOver` even if the drop target index hasn't changed. I recommend the following modification to cache the previous result and bail early if it's the same:

```diff
diff --git a/lib/root-drag-and-drop.coffee b/lib/root-drag-and-drop.coffee
index af68e01..829fd15 100644
--- a/lib/root-drag-and-drop.coffee
+++ b/lib/root-drag-and-drop.coffee
@@ -24,6 +24,7 @@ class RootDragAndDropHandler
     @treeView.on 'drop', '.tree-view', @onDrop
 
   onDragStart: (e) =>
+    @lastDropTargetIndex = null
     e.originalEvent.dataTransfer.setData 'atom-event', 'true'
     projectRoot = $(e.target).closest('.project-root')
     directory = projectRoot[0].directory
@@ -75,6 +76,8 @@ class RootDragAndDropHandler
 
     newDropTargetIndex = @getDropTargetIndex(e)
     return unless newDropTargetIndex?
+    return if @lastDropTargetIndex is newDropTargetIndex
+    @lastDropTargetIndex = newDropTargetIndex
 
     @removeDropTargetClasses()
 ```

This resolves the jank on my machine.

Finally, I wonder if it makes more sense to put the placeholder above the hovered project if you're on the upper half of the *entire* entry, not just the header.

dnd-root-half

In that case, we'd additionally need

```diff
diff --git a/lib/root-drag-and-drop.coffee b/lib/root-drag-and-drop.coffee
index 829fd15..62edee5 100644
--- a/lib/root-drag-and-drop.coffee
+++ b/lib/root-drag-and-drop.coffee
@@ -157,9 +158,7 @@ class RootDragAndDropHandler
 
     return 0 unless projectRoot.length
 
-    element = projectRoot.find('.project-root-header')
-
-    elementCenter = element.offset().top + element.height() / 2
+    elementCenter = projectRoot.offset().top + projectRoot.height() / 2
 
     if e.originalEvent.pageY < elementCenter
       projectRoots.index(projectRoot)

```

forivall:
@BinaryMuse Hmm, I didn't think that reinserting the same element would cause the issue, and I wasn't seeing jank on my machines (OSX or Linux, stable atom). But cool, that's a simpler solution than what I briefly tried over my lunch break (debouncing).

Also, I originally _did_ have the placeholder position based on the center of the entire entry, but switched it to the header during the phase where on-disk dragging was allowed. :+1: to reverting back to that placeholder positioning.

Ben3eeE:
@BinaryMuse Nice :+1: 

I will test this again later tonight to verify functionality and responsiveness on my machine.

Ben3eeE:
Accidentally did this I don't know what I expected to happen 😆 
525 bug

Similarly in reverse:
525 bug 2

forivall:
@Ben3eeE Fixed by https://github.com/atom/tree-view/pull/525/commits/7dbae5036477138c700bb79ce3e6086164b9b4d3 😆

forivall:
@BinaryMuse I also added the 

> Finally, I wonder if it makes more sense to put the placeholder above the hovered project if you're on the upper half of the entire entry, not just the header.

in edf9e5b. Try it out!

BinaryMuse:
@forivall This looks awesome, great work! I'm going to look into the spec failures (which have nothing to do with your PR) and once that's resolved I'll merge this. 🎉

JetUni:
@forivall THANK YOU SO MUCH!!!!! 🎉🎉

peterschmidler:
How can I use this feature?

thomasjo:
If you want it right now, you'll have to build Atom from source. If you can wait, this will be available in the next beta release.

forivall:
@peterschmidler you can also clone this repository and `npm install && apm rebuild && apm link .`

50Wliu:
Or just `apm install && apm link`.

BinaryMuse:
Keep in mind that installing the package from a local source via `apm link` will shadow the built-in package and keep you from using the latest version when Atom updates.

KSXGitHub:
What about `apm install tree-view`?

BinaryMuse:
@KSXGitHub I guess to be more specific, *any* mechanism by which you place a package with the same name as a core, bundled package in your `.atom/packages` directory will shadow the bundled package (that's why doing it makes the feature work. :) But if you forget to remove it, Atom will continue to use the one you installed instead of the bundled one, even if the bundled one is newer.

thomasjo:
I highly advise people to **not** install core packages directly — it only leads to tears.

Core packages should only be symlinked in dev mode when you are actively working on the package.

Since this issue has gone completely off-topic, and needlessly notifying 21+ people is unnecessary, I'm locking this thread. If someone wants to discuss this further, do it on the Slack channel or on Discuss 🙇

Issue #515, opened 2015-07-09
Quick jump


johngeorgewright:
Something that I'd love to see is a quick way of jumping from one part of the tree to another. A bit like:

![tree-view-jump](https://cloud.githubusercontent.com/assets/1423566/8591471/6b273556-2620-11e5-96eb-324f3ae7d052.jpg)

Maybe toggling it on and off with the classic ViM `/` search key? If people would be at all keen for this too, I don't mind opening a pull request.

PR #532, opened 2015-07-24
Jumping around the tree-view with shortcuts

johngeorgewright:
**For the people afraid of mice... "clickady click"**

An attempt at #515.

Bored of pressing `down down down down down` or `up up up up up`? Problem solved. Hit `/` and every entry in the tree is marked with a number. Type the number to select the entry.

shortcuts

You may notice that not **all** the tests are there. However, I thought I'd begin the pull request process to get feedback while I finish them off.

### DIFF
--- a/lib/tree-view.coffee
+++ b/lib/tree-view.coffee
@@ -1,5 +1,6 @@
 path = require 'path'
 shell = require 'shell'
+{Emitter} = require 'event-kit'
 
 _ = require 'underscore-plus'
 {BufferedProcess, CompositeDisposable} = require 'atom'
@@ -37,6 +38,7 @@ class TreeView extends View
     @scrollTopAfterAttach = -1
     @selectedPath = null
     @ignoredPatterns = []
+    @emitter = new Emitter()
 
     @handleEvents()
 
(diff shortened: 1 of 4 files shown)


lib/tree-view.coffee
```diff
@@ -671,6 +677,54 @@ class TreeView extends View
...
+  enableShortcuts: ->
+    @emitter.on 'did-toggle-directory', @updateAllTreeStartIndexes
+    @updateAllTreeStartIndexes()
+    @element.classList.add 'shortcuts'
+
+  updateAllTreeStartIndexes: =>
+    @removeShortcuts()
+    @setStartIndexesWithin 0, @element.querySelector('.tree-view-scroller')
+
+  setStartIndexesWithin: (index, el) ->
```
johngeorgewright: Actually, this is a bonkers name. Perhaps something more along the lines of `addShortcutsToListTrees`?

damassi:
It would also be nice if in addition to shortcuts, pressing `/` would also allow you to do a quick search and jump directly to the item similar to how NERDTree does it in Vim.  

(And a nice nice nice one would be the addition of paging -- meaning, `ctrl+f` to scroll one page down and `ctrl+b` to page up)

johngeorgewright:
Yeh, that could work too. Perhaps the search and paging could be added a later date though. You can already page up and down, and with this, the search is half way there.

johngeorgewright:
So, I take it this isn't going to be accepted. Would it be a better idea to make a search functionality that the other comments were suggesting? Or would it be better to make a separate package to fulfil one of the ideas?

PR #534, opened 2015-07-25
Respect other panels when resizing. Fixes #533

### LINKED ISSUE #533: Resizing doesn't respect other panes
Resizing tree-view doesn't take other panels in account when adjusting the width. It was [noticed](https://github.com/suda/tool-bar/issues/73) when the [tool-bar package](https://github.com/suda/tool-bar) is installed, but happens with any panel added to the left: 
```
var panel = document.createElement('div');
panel.appendChild(document.createTextNode('test'));
atom.workspace.addLeftPanel({ item: panel, priority: 10 });
```
resize tree-view

Since Atom 0.210.0 up till 1.0.2

jerone:
Fixes #533
Ref https://github.com/suda/tool-bar/issues/73

### DIFF
--- a/lib/tree-view.coffee
+++ b/lib/tree-view.coffee
@@ -217,8 +217,7 @@ class TreeView extends View
     return @resizeStopped() unless which is 1
 
     if atom.config.get('tree-view.showOnRightSide')
-      docWidth = $(document.body).width()
-      width = docWidth - pageX - (docWidth - @outerWidth() - @offset().left)
+      width = @outerWidth() + @offset().left - pageX
     else
       width = pageX - @offset().left
     @width(width)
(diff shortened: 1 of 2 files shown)


lib/tree-view.coffee
```diff
@@ -217,9 +217,10 @@ class TreeView extends View
     return @resizeStopped() unless which is 1
 
     if atom.config.get('tree-view.showOnRightSide')
-      width = $(document.body).width() - pageX
+      docWidth = $(document.body).width()
+      width = docWidth - pageX - (docWidth - @outerWidth() - @offset().left)
```
kevinsawicki: Since `docWidth` is here twice separated by a `-`, I think this could just be:

```coffee
width = @outerWidth() + @offset().left - pageX
```

jerone: @kevinsawicki You are right; updated, tested and committed.

kevinsawicki:
Awesome, thanks for this :+1:

PR #138, opened 2014-05-10
Add drag & drop

saschagehlich:
This PR adds drag and drop functionality to the `tree-view` package.

tree-view-dnd

PR #543, opened 2015-07-30
Drag and drop

saschagehlich:
Based on @abm0's PR last month: https://github.com/atom/tree-view/pull/478

tree-view

Changes I made:

 * Make whole FileView / DirectoryView draggable
 * Added specs
 * Use `selected` class for hovering

### DIFF
--- a/lib/directory-view.coffee
+++ b/lib/directory-view.coffee
@@ -9,6 +9,8 @@ class DirectoryView extends HTMLElement
     @subscriptions.add @directory.onDidDestroy => @subscriptions.dispose()
     @subscribeToDirectory()
 
+    @draggable = true
+
     @classList.add('directory', 'entry',  'list-nested-item',  'collapsed')
 
     @header = document.createElement('div')
@@ -35,7 +37,6 @@ class DirectoryView extends HTMLElement
     @directoryName.textContent = @directory.name
     @directoryName.dataset.name = @directory.name
     @directoryName.dataset.path = @directory.path
-    @directoryName.draggable = true
 
     if @directory.isRoot
       @classList.add('project-root')
--- a/lib/directory-view.coffee
+++ b/lib/directory-view.coffee
@@ -35,6 +35,7 @@ class DirectoryView extends HTMLElement
     @directoryName.textContent = @directory.name
     @directoryName.dataset.name = @directory.name
     @directoryName.dataset.path = @directory.path
+    @directoryName.draggable = true
 
     if @directory.isRoot
       @classList.add('project-root')
--- a/lib/tree-view.coffee
+++ b/lib/tree-view.coffee
@@ -809,9 +809,7 @@ class TreeView extends View
   #
   # returns noop
   onDragStart: (e) ->
-    $target = $(e.target)
-
-    initialPath = $target.data("path")
+    initialPath = $(e.target).data("path")
 
     e.originalEvent.dataTransfer.effectAllowed = "move"
     e.originalEvent.dataTransfer.setData("initialPath", initialPath)
--- a/lib/tree-view.coffee
+++ b/lib/tree-view.coffee
@@ -96,11 +96,11 @@ class TreeView extends View
 
     @on 'mousedown', '.tree-view-resize-handle', (e) => @resizeStarted(e)
 
-    @on 'dragstart', '.entry .name', (e) => @onDragStart(e)
+    @on 'dragstart', '.entry', (e) => @onDragStart(e)
 
-    @on 'dragenter', '.entry.directory > .header > .name', (e) => @onDragEnter(e)
+    @on 'dragenter', '.entry.directory > .header', (e) => @onDragEnter(e)
 
-    @on 'dragleave', '.entry.directory > .header > .name', (e) => @onDragLeave(e)
+    @on 'dragleave', '.entry.directory > .header', (e) => @onDragLeave(e)
 
     @on 'dragover', '.entry', (e) => @onDragOver(e)
 
@@ -798,18 +798,20 @@ class TreeView extends View
   onDragEnter: (e) ->
     e.stopPropagation()
 
-    e.target.classList.add('dragenter')
+    e.currentTarget.parentNode.classList.add('selected')
 
   onDragLeave: (e) ->
     e.stopPropagation()
 
-    e.target.classList.remove('dragenter')
+    e.currentTarget.parentNode.classList.remove('selected')
 
   # Public: Handle entry name object dragstart event
   #
   # returns noop
   onDragStart: (e) ->
-    initialPath = $(e.target).data("path")
+    e.stopPropagation()
+
+    initialPath = $(e.currentTarget).find(".name").data("path")
 
     e.originalEvent.dataTransfer.effectAllowed = "move"
     e.originalEvent.dataTransfer.setData("initialPath", initialPath)
@@ -829,14 +831,13 @@ class TreeView extends View
     e.stopPropagation()
 
     entry = e.currentTarget
+    return unless entry instanceof DirectoryView
 
-    if entry instanceof DirectoryView
-      initialPath = e.originalEvent.dataTransfer.getData("initialPath")
-      newDirectoryPath = $(e.target).data("path")
+    initialPath = e.originalEvent.dataTransfer.getData("initialPath")
+    newDirectoryPath = $(entry).find(".name").data("path")
 
-      e.target.classList.remove('dragenter')
+    entry.classList.remove('dragenter')
 
-      unless newDirectoryPath
-        return false
+    return false unless newDirectoryPath
 
-      @moveEntry(initialPath, newDirectoryPath)
+    @moveEntry(initialPath, newDirectoryPath)
--- a/lib/tree-view.coffee
+++ b/lib/tree-view.coffee
@@ -3,7 +3,7 @@ shell = require 'shell'
 
 _ = require 'underscore-plus'
 {BufferedProcess, CompositeDisposable} = require 'atom'
-{repoForPath, relativizePath} = require "./helpers"
+{repoForPath, relativizePath, getStyleObject} = require "./helpers"
 {$, View} = require 'atom-space-pen-views'
 fs = require 'fs-plus'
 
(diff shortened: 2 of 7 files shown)


saschagehlich:
FYI: There's one little bug that I was not able to fix:

When you drag a file onto its own parent directory, the background color has a gap on the left. I haven't tested it with the default themes, so this might be caused by the Material UI theme I'm using.

saschagehlich:
@mnquintana Ping.

Related: #138

lib/tree-view.coffee
```diff
@@ -95,6 +96,16 @@ class TreeView extends View
...
+    @on 'dragstart', '.entry', (e) => @onDragStart(e)
+
+    @on 'dragenter', '.entry.directory > .header', (e) => @onDragEnter(e)
+
+    @on 'dragleave', '.entry.directory > .header', (e) => @onDragLeave(e)
+
+    @on 'dragover', '.entry', (e) => @onDragOver(e)
+
+    @on 'drop', '.entry', (e) => @onDrop(e)
+
```
thomasjo: You should probably remove all the extra whitespace (blank lines) above here.

thomasjo:
At first glance this looks really promising! Great work.

mnquintana:
This is looking pretty good :smile:

Here are a few things I noticed:
- When you first start dragging a file, the cursor briefly changes to the `copy` cursor before changing back to `default`
- If a file name is cut off (ie. because the tree-view width is too small to show the full name), then it remains cut off when you drag:

  screenshot_7_30_15__5_48_pm

- Dragging an open folder also visually drags along all of its children - this is okay for folders with only a few children, but could be kinda bad for folders with a lot of children
- If an expanded folder is taller than the viewport height, then nothing is displayed when you drag it

styles/tree-view.less
```diff
@@ -45,10 +45,20 @@
...
-  .entry:before {
-    content: '';
-    position: absolute;
+  .entry {
+    .name {
+      pointer-events: none;
+    }
+
+    &.dragenter {
+      background-color: rgba(255, 255, 255, 0.1);
```
mnquintana: It would be great if this used something from `ui-variables` so that this color can be explicitly changed by themes

saschagehlich: The .dragenter class is not used anymore, since I'm just adding the `selected` class to the DirectoryView - I'll remove that

saschagehlich:
@mnquintana I think I'll need to take another look at my old version of this feature, since I don't have any influence on how the dragged item is being displayed (it's all handled by HTML5's drag-and-drop feature). Do you think that the issues you named are critical?

lib/tree-view.coffee
saschagehlich: Absolutely. I'd also move this to a separate method. I just copied it from the old PR and since nobody mentioned this issue, I just left it as it was.

mnquintana:
@saschagehlich I think you should have some control over it with [`setDragImage()`](https://developer.mozilla.org/en-US/docs/Web/API/DataTransfer#setDragImage.28.29), which should let you customize exactly what's seen as you're dragging. This might involve adding some custom styling to prevent the visual artifacts I was describing, though.

saschagehlich:
@mnquintana Didn't know about this function yet - neat! But it expects an image or canvas element - we only have the DOM elements now. I read that it's possible to render an SVG element containing a `<foreignObject>` with plain HTML, but this would also need to include inline styles. This could get kind of tricky.

mnquintana:
@saschagehlich The docs aren't the most clear about this, but it actually accepts any DOM node - the only catch is that it has to be visible.

saschagehlich:
@mnquintana Oh lord... I've been playing around with it for 30 minutes now until I found this post:

http://www.kryogenix.org/code/browser/custom-drag-image.html

TL;DR: The DOM node needs to be visible on screen, in full size. If its container makes it partially invisible (e.g. using overflow: hidden), the drag image will be cut off as well. But, if you put another element on top of it (e.g. a div of the same size in the background color of the editor), the drag image is fine.

So, the only solution would be: Clone the element, put it somewhere where it's not cut off, put something on top of it, making it invisible to the user. Then, set the drag image. Wait for one frame to be rendered, then it can be removed again. Could probably result in a very short (1 frame) flickering somewhere on the screen.

Any other idea how to get this to work?

saschagehlich:
Also, if I want to position it anywhere on the screen / in the editor, I'll probably need to append it to the body temporarily, where it will not be styled by the CSS rules. Meh.

saschagehlich:
Here we go:

https://github.com/saschagehlich/tree-view/commit/0bd4981f13127bb0e107bf0ae970263fa7911b20

tree-view-2

Unfortunately I can't get rid of the copy-cursor that appears for 1 frame. I guess that's an issue with Chrome, hopefully this gets resolved with one of the coming releases.

Final clean up commit (hopefully final) is coming up in the next few minutes :)

saschagehlich:
@mnquintana Maybe we should set the drag cursor to "copy"? The tabs are doing the same, even though they're moving tabs instead of copying them

mnquintana:
> Maybe we should set the drag cursor to "copy"? The tabs are doing the same, even though they're moving tabs instead of copying them

I think we should keep it as `default`, because the `copy` cursor in this context would seem to indicate that dragging files / folders *copies* them, when it actually moves them.

saschagehlich:
@mnquintana Did you review this MR? I'd love to see this feature in atom :)

saschagehlich:
@mnquintana @kevinsawicki Please don't delay this again :(

abm0:
@mnquintana What do you think about using @text-color-highlight text color for elements, that are being hovered while dragging element? 

drag-hover-demo

It will solve the problem with background color of left gap and will add unique visual effect for this action without need of modifying existing ui themes.

lib/tree-view.coffee
```diff
@@ -765,3 +784,60 @@ class TreeView extends View
...
+    e.stopPropagation()
+
+    e.currentTarget.parentNode.classList.add('selected')
+
+  onDragLeave: (e) ->
+    e.stopPropagation()
+
+    e.currentTarget.parentNode.classList.remove('selected')
+
+  # Public: Handle entry name object dragstart event
```
nathansobo: Why are these documented as public? They seem pretty internal.

lib/directory-view.coffee
```diff
@@ -9,6 +9,8 @@ class DirectoryView extends HTMLElement
     @subscriptions.add @directory.onDidDestroy => @subscriptions.dispose()
     @subscribeToDirectory()
 
+    @draggable = true
```
nathansobo: What's the reason for assigning these two booleans?

nathansobo:
One thing I would expect is that when you drag a file A next to file B, it would give you the chance to drop file A and have it become a sibling of file B... in other words, make it possible to drag files to directories without dragging directly over the top of a folder. But that could be a followup PR since this is already an improvement.

lib/tree-view.coffee
```diff
@@ -671,6 +676,24 @@ class TreeView extends View
...
+    try
+      fs.makeTreeSync(newDirectoryPath) unless fs.existsSync(newDirectoryPath)
+      fs.moveSync(initialPath, newPath)
+
+      if repo = repoForPath(newPath)
+        repo.getPathStatus(initialPath)
+        repo.getPathStatus(newPath)
+
+    catch error
+      console.warn("#{error.message}.")
```
nathansobo: This should post a notification similar to [this code](https://github.com/atom/tree-view/blob/fc3c32853735d708dfb8d91e5647992e76349b75/lib/tree-view.coffee#L569) rather than just emitting to the console. I don't think that's visible enough.

nathansobo:
This is looking solid. I have a few inline comments/questions but once those are done I am cool with merging this.

saschagehlich:
Cool, thanks for reviewing! I'll take a look at the changes tomorrow night
:)

izuzak:
> One thing I would expect is that when you drag a file A next to file B, it would give you the chance to drop file A and have it become a sibling of file B... in other words, make it possible to drag files to directories without dragging directly over the top of a folder. But that could be a followup PR since this is already an improvement.

Also noticed this, but agree with @nathansobo that it doesn't need to be a part of this PR. Another thing which might be nice to do in a separate PR is to auto-expand directories when you hover the cursor over a specific directory for a few seconds (when dragging).

Two other things I noticed:

1. The selection highlight for the dragged item at its original position changes width as you move the cursor over different hierarchies:

  

  Notice how the selection highlight for invalid.js changes as I drag the item over different folders.

2. It's possible to scroll the tree view up and down when dragging by moving the cursor close to the edge :sparkles:, but if the cursor moves over the edge the scrolling stops. I expected the scrolling to continue, similar to how you can scroll when drag-selecting lines in the editor, even if the cursor moves outside the borders of the editor.

  

Great work, @saschagehlich! :clap: :sparkles: :tophat:

lib/tree-view.coffee
saschagehlich: I just made them public because all other callbacks were documented as public as well, e.g. `resizeStarted`, `entryClicked`, `onMouseDown`. I can refactor this whole think if you like

lib/directory-view.coffee
saschagehlich: https://developer.mozilla.org/en-US/docs/Web/Guide/HTML/Drag_operations#draggableattribute

saschagehlich:
@izuzak 

> The selection highlight for the dragged item at its original position changes width as you move the cursor over different hierarchies:

This is due to the way the tree's DOM is built - I think we should implement @abm0's suggestion to change the font color on hover (for now)

> I expected the scrolling to continue, similar to how you can scroll when drag-selecting lines in the editor, even if the cursor moves outside the borders of the editor.

I think the problem here is that the editor window is very special when it comes to scrolling and rendering. The tree-view is just a simple div with `overflow: scroll` and scrolling is handled by the browser. I guess this should come with a new PR as well?

lib/tree-view.coffee
saschagehlich: Oh now I see, you mean the actual documentation. I'm gonna fix that

saschagehlich:
@izuzak I fixed the hovering issue, looks nice now :)

@nathansobo I think I fixed everything you noted!

izuzak:
> @izuzak I fixed the hovering issue, looks nice now :)

:sparkles: Thanks, @saschagehlich :bow:

@nathansobo Do you think this is ready to :boat: now?

lib/directory-view.coffee
nathansobo: Ah, of course. For a second I was thinking this was a SpacePen view but it's a custom element. Makes sense.

nathansobo:
Thanks a ton @saschagehlich.

mnquintana:
:rocket: :guitar: 

Thanks so much for this @saschagehlich and @abm0!

theaqua:
@saschagehlich big thanks for your work!

kevinsawicki:
:+1: :fireworks:

maxbrunsfeld:
Wow, this is super awesome @saschagehlich and @abm0!

kevinsawicki:
This is now out in Atom 1.0.7 :ship:

tusharmath:
undo doesn't revert the changes though.

joel:
How can we deactivate this feature?

mattqs:
Ctrl drag and drop does not create a copy instead of just moving the file (as it is in Visual Studio)

mnquintana:
@mattqs Feel free to open a new issue for adding that enhancement.

Since this has been merged for awhile now, I'm going to lock this PR. If you have any suggestions for improvement, feel free to move the conversation to other issues. :sparkles:

PR #550, opened 2015-08-08
Update keybindings in the readme

tomekwi:
The `ctrl-0` seems to have [changed](https://github.com/atom/tree-view/commit/9634a08650ac7676240e9dc20ed593121f32c61f) in Linux and Windows recently.

PR #551, opened 2015-08-08
Update markup for keyboard shortcuts

tomekwi:
If you do merge <https://github.com/atom/tree-view/pull/550>, it’d be better to have these consistent.


kevinsawicki:
Thanks for keeping these consistent :ship:

Issue #555, opened 2015-08-19
v0.183.0 breaks certain context menu items


Galadirith:
Certain context menu items seem to no longer work. For example both ['Markdown Preview'](https://github.com/atom/markdown-preview/blob/42eb71da984232faaba2a5f75d30a1d8e1d138cf/menus/markdown-preview.cson#L19) and ['Beautify File'](https://github.com/Glavin001/atom-beautify/blob/72e4384cf815e633e580ffdad8847f355e3e7c89/menus/atom-beautify.cson#L7) do not show up in the context menu. 

- **`v0.183.0`**
![v0 183 0](https://cloud.githubusercontent.com/assets/1201875/9367621/7c80cad6-46b6-11e5-99b1-de4460798bbd.gif)

- **`v0.182.0`**
![v0 182 0](https://cloud.githubusercontent.com/assets/1201875/9367655/a8c69dd2-46b6-11e5-90c9-cf0592dfc21d.gif)

I have tested that the issue does not appear to exist for `v0.182.0`. From my own experiments the problem seems to be selecting leaf elements of the `.tree-view`. For example `.tree-view .file` appears to work fine.

When running in safe mode 'Markdown Preview' still does not show up in the context menu.

Thanks so much for all the amazing work on Atom :smile:

kevinsawicki:
@saschagehlich @nathansobo this looks to have possibly regressed via the drag and drop work done in #543

Galadirith:
It looks as though the style [`.tree-view .entry .name`](https://github.com/atom/tree-view/blob/6f5ab9eab4f58087c25ad337dc62a2cf62b9157b/styles/tree-view.less#L49-L51) add in 6f5ab9e is the problem. Removing it cures the issue, but I don't know what else that style affects. Thanks @kevinsawicki for the super quick response.

saschagehlich:
I made some changes to the CSS to fix the hovering issue mentioned [here](https://github.com/atom/tree-view/pull/543#issuecomment-129742134). I honestly never used any of the custom context menu items. But I'll look into that now

Galadirith:
@saschagehlich Thanks for taking a look into this, really appreciate it.

saschagehlich:
Pinging @nathansobo @kevinsawicki @mnquintana

cakebake:
+1

Galadirith:
Thank you so much for the super fast response @saschagehlich @nathansobo @kevinsawicki @izuzak, its great to have this back in master.

[...]

PR #556, opened 2015-08-20
Fix issue 555

saschagehlich:
#555 stated that some context menu didn't appear anymore after my drag and drop feature has been added. This was due to the fact that I removed pointer-events from the .name element, for the following reason:

Whenever you hover an entry, a dragEnter event is called. When you hover the .name element inside this entry, it triggers another dragEnter event and a dragLeave event immediately after that, which results in the entry not being highlighted.

I fixed this by counting the events of all entries in a global dragEventCounts hash. The dragLeave event handler only removes the selected class if the count is 0.

### DIFF
--- a/lib/tree-view.coffee
+++ b/lib/tree-view.coffee
@@ -39,6 +39,8 @@ class TreeView extends View
     @selectedPath = null
     @ignoredPatterns = []
 
+    @dragEventCounts = {}
+
     @handleEvents()
 
     process.nextTick =>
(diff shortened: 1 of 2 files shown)


nathansobo:
@saschagehlich What about checking the target of the `dragenter` and `dragleave` events explicitly and making sure it's only happening on the desired element?

saschagehlich:
@nathansobo The problem is that these events are triggered twice. First time when you enter the element and second time when you enter its child. The problem is that the dragleave of the parent element is triggered *after* the dragenter of the child. Unfortunately dragleave events don't have some kind of `destinationElement` property that lets me check whether the cursor has moved to a child element. That's why I need to keep track of the event count.

nathansobo:
After playing with this a bit I see what you're saying now. Is it necessary to bucket the event counts by identifier? Could a single counter be incremented and decremented?

nathansobo:
Nevermind, I see that's not going to work either due to awkward sequencing.

saschagehlich:
Exactly ;)

nathansobo:
Thanks a ton for the prompt followup on this @saschagehlich. I really appreciate it.

I added some minimal test coverage and switched to using a `WeakMap` to track the counts so we don't have to synthesize an identifier. Makes the syntax a bit more awkward but I think it's simpler conceptually. Let me know if you disagree.

saschagehlich:
Looks good, thanks for the merge!

PR #578, opened 2015-09-23
Avoid using a hardcoded display value when forcing a redraw

abe33:
Simply removing the previously defined display is enough to make it
visible again and it won’t override what the stylesheet or themes may
have defined for the tree-view resizer div.


izuzak:
cc @simurai for :eyes:

simurai:
Merging, this seems good.

PR #582, opened 2015-09-29
Add option to always open existing view

forivall:
If the file is already open, by default, if it's not open in the current pane, tree view will open a new view onto that file in the current pane.

This option, when set, will open the existing view of a selected file.

### DIFF
--- a/spec/tree-view-spec.coffee
+++ b/spec/tree-view-spec.coffee
@@ -18,6 +18,20 @@ clickEvent = (properties) ->
   _.extend(event, properties) if properties?
   event
 
+setupPaneFiles = ->
+  rootDirPath = fs.absolute(temp.mkdirSync('tree-view'))
+
+  dirPath = path.join(rootDirPath, "test-dir")
+
+  fs.makeTreeSync(dirPath)
+  [1..9].forEach (index) ->
+    filePath = path.join(dirPath, "test-file-#{index}.txt")
+    fs.writeFileSync(filePath, "#{index}. Some text.")
+
+  return dirPath
+
+getPaneFileName = (index) -> "test-file-#{index}.txt"
+
 describe "TreeView", ->
   [treeView, path1, path2, root1, root2, sampleJs, sampleTxt, workspaceElement] = []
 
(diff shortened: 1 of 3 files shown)


50Wliu:
Please add specs for this.

forivall:
@50Wliu specs added.

spec/tree-view-spec.coffee
```diff
@@ -1105,6 +1119,39 @@ describe "TreeView", ->
             expect(atom.views.getView(pane)).toHaveFocus()
             expect(item.getPath()).toBe atom.project.getDirectories()[0].resolve('tree-view.txt')
 
+  xdescribe "opening existing opened files in existing split panes", ->
```
50Wliu: y reason this is `xdescribe`d?

forivall: It takes a long time, and probably isn't necessary. However, i wanted feedback on it. Sorry, i should have mentioned it.

forivall: @50Wliu  So, should i remove it completely, or just enable it?

50Wliu: I would say enable it.

forivall:
@50Wliu test enabled and passing.

int3h:
@50Wliu any update on merging this in? I would really like to see this feature. It's really annoying to get duplicate buffers, just because I didn't happen to have my cursor in the right pane when I clicked a file in tree view.

I virtually never want to duplicate a buffer without explicitly intending to do so. With the current version of tree view, to prevent duplicates, I have to manually scan the tab bar of all panes to see if the file is already open before clicking a file in tree view. This kind of defeats the purpose of the "switch to a file if it's already open" behavior.

The behavior added by this PR brings Atom in line with how Sublime Text and many other multi-pane/tabbed editors already work.

50Wliu:
@int3h This PR currently has conflicts with `master` so it can't be merged until those conflicts are fixed.

Sorry for the gap in communication though.

int3h:
@50Wliu Would the new behavior added by this PR be accepted if the code was updated? 

If @forivall wants assistance on this, I can pitch in, but it only makes sense to spend that effort if the fundamental idea of the PR is acceptable 😃

50Wliu:
I'm fine with this - /cc @atom/feedback.

lee-dohm:
@int3h The fundamental idea of the PR is acceptable, yes :grinning:

forivall:
@int3h I'll be updating the PR soon (by tonight (PDT) hopefully)

forivall:
Ok, I rebased it, but I'm needing some help updating the specs for the updated master, since the pending pane logic has changed significantly since I originally write this PR.

lee-dohm:
@int3h Do you want to help out with the specs?

forivall:
Figured it out! Changing the directories in the tree view removed the elements I was triggering clicks upon from the DOM. So I just don't do that, and leave the directories alone.

🎉 🎉  🎉 🎈

lee-dohm:
Thanks so much!

Ben3eeE:
Shouldn't there also be a description to this setting?
nodesc

forivall:
Oh yeah! Sorry, forgot to add that 😂

phoenixeliot:
@forivall Looks like that description is still missing — found this thread because I googled it out of confusion ;)

forivall:
Oh, you did it in 0ae4993 :stuck_out_tongue:. Thanks!

PR #596, opened 2015-10-14
Fix instanceof check on doubleclick on an entry

olmokramer:
It would always call `entry.toggleExpansion` because `DirectoryView` is always truthy


kevinsawicki:
Thanks for this :+1:

Issue #343, opened 2015-01-27
Sort files by modification date


damiendr:
An option to sort the files by modification date would be nice. My use case is note-taking, where it is convenient to display the most recently updated notes at the top (like in Notational Velocity).

ObviouslyGreen:
Also an option to sort by file type would be nice too, being able to navigate through code for a large project folder with a bunch of header files, source code files, resources, etc. You could just select and open only the desired files in atom, but then you lose the ability to see all the other files and folders in the project.

guiraldelli:
> Also an option to sort by file type would be nice too
:+1: 

Actually, it is impressive the lack of sorting options. The ideal situations would be something like

    tree-view.sortList = [
        filetype
        modified
        alphabetically
    ]

and it sort the files by file type, modification date or alphabetically, for example.

vjpr:
+1

pawelkuras:
+1

Michael-Thibault:
+1 bajillion There really ought to be a way to sort the tree view however the heck one wants. Or at least a clear way to add the extra functionality yourself.

lukelamar:
+1

piascikj:
+1

standyro:
:+1:

Sporky023:
+1

jigargosar:
+1

petonic:
+1

jwhitmarsh:
+1 if we could sort by 'tag' too this would be great for people using https://github.com/davej/john

luisaceituno:
+1

vjpr:
Until this is implemented, you can use https://atom.io/packages/tree-view-finder.

xu4wang:
＋1

d-max:
+1

jonathan-dejong:
+1

andromedarabbit:
+1 also

Draghi:
+1 Desperately need sort by filetype too.

LeoStaley:
+1 

The [tree-view-finder package](https://atom.io/packages/tree-view-finder) hasn't been updated in 2 years, and is severely lacking in features. Sorting by date should be a core part of the tree-view package. 

I'm not a programmer, but a writer, so I can't do anything to improve it, but if this feature won't be added to the core, could someone fork tree-view-finder to add the needed features?

sevenam:
+1

miguelcg:
is this something in the roadmap @lee-dohm?

[...]

PR #598, opened 2015-10-14
Sort files and folders by name/mtime (etc)

jkgeyti:
Preliminary example of how support (recursive) directory sorting in tree-view.

In response to https://github.com/atom/tree-view/issues/343


billybonks:
should, wait or come with a benchmark test https://github.com/atom/tree-view/pull/616

50Wliu:
As mentioned in #343, additional sort orders in the tree-view isn't something that we want to support.  I think this would be a great package idea though!

Issue #581, opened 2015-09-24
Copy & pasting a file to the same directory will remove the extension


abe33:
Initially reported as atom/atom#8928

If a file is copy/pasted in the Tree View, the extension will be removed.

Examples:

`./style.min.css` will create `./style0.min`
`./script.min.js` will create `./script.min`
`./index.blade.php` will create `./index0.blade`
`./index.blade.php` copied to another directory is properly created as `../otherdir/index.blade.php`

Files like `index.html` are properly copied to `index0.html`

`atom --safe` doesn't help either.

Mac OS X 10.11
Atom 1.0.16

abe33:
I could confirm this on my own computer, the issue only occurs when the file name contains several extensions and that it get pasted in the same directory. Whatever the number of extensions the file is pasted only with the first one.

PR #599, opened 2015-10-14
Fix issue #581

crackedmind:
Simple fix for issue #581


50Wliu:
Please add specs to this to make sure this behavior doesn't regress in the future.

50Wliu:
#581 was fixed by #649.  Thanks for the PR though @crackedmind!

PR #601, opened 2015-10-16
:bug: Disallow moving root directories

### LINKED ISSUE #590: Disallow moving project folder
_From @Kroid on September 30, 2015 12:56_

Drag-ang-drop 'project folder' to another folder - bad feature, because undo is difficult. Maybe do not allow this or use copy instead move?

_Copied from original issue: atom/atom#9008_

Tyriar:
Moving a root directory can cause the tree view to become invalid as the
project remains but does not correctly map to a directory on the file system.
It also provides a bad UX as it's very easy to accidentally move a project
directory.

Fixes #590

### DIFF
--- a/spec/tree-view-spec.coffee
+++ b/spec/tree-view-spec.coffee
@@ -76,7 +76,7 @@ describe "TreeView", ->
       expect(treeView.selectedEntry()).toEqual(treeView.roots[0])
 
     it "makes the root folder non-draggable", ->
-      expect(treeView.roots[0]).not.toHaveClass('draggable')
+      expect(treeView.roots[0].hasAttribute('draggable')).toBe(true)
 
     describe "when the project has no path", ->
       beforeEach ->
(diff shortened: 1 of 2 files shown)


Zireael07:
Thanks!

spec/tree-view-spec.coffee
```diff
@@ -75,6 +75,9 @@ describe "TreeView", ->
     it "selects the root folder", ->
       expect(treeView.selectedEntry()).toEqual(treeView.roots[0])
 
+    it "makes the root folder non-draggable", ->
+      expect(treeView.roots[0].hasAttribute('draggable')).toBe(true)
```
50Wliu: I'm not too familiar with these tests, but shouldn't `draggable` be false here?

Tyriar:
@50Wliu good point, the test wasn't correct - I changed it to only set the `draggable` attribute if it's true.

kevinsawicki:
Thanks for this :ship:

PR #602, opened 2015-10-16
Expand tree-view to full or auto height

### LINKED ISSUE #595: Empty context menu when tree view doesn't take entire vertical space
After https://github.com/atom/tree-view/pull/579 was merged, the `.tree-view` doesn't extend vertically to the bottom of the `.tree-view-scroller` when only a few files/directories are visible. The problem is that context menu entries use the `.tree-view` selector, so when I right-click on space below the tree view the context menu is empty.

image

simurai:
This PR fixes the issue that the context menu didn't work when the tree-view wasn't taking up the whole height.

Fixes #595

### DIFF
--- a/styles/tree-view.less
+++ b/styles/tree-view.less
@@ -41,12 +41,16 @@
 }
 
 .tree-view-scroller {
+  display: flex;
+  flex-direction: column;
   flex: 1;
   width: 100%;
   overflow: auto;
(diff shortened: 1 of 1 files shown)


thomasjo:
:trophy: Nice, one! I guess I could have waited with the revert :laughing:

PR #608, opened 2015-10-17
Add setting to squash directories

### LINKED ISSUE #125: Collapse Java-style empty hierarchies
I occasionally have to work on Java projects for work. Java essentially requires project hierarchies like the following:

```
\project
    \src
        \com
            \example
                \project
                    \app
                        source1.java
                        source2.java
                        source3.java
                    \library-foo
                        source1.java
                        source2.java
                        source3.java
```

Java IDEs have a feature in their tree-view controls to collapse directories that only have one entry in them underneath `src`. So the above would look like this:

```
\project
    \src
        \com.example.project
            \app
                source1.java
                source2.java
                source3.java
            \library-foo
                source1.java
                source2.java
                source3.java
```

I'd like to see Atom tree-view support this feature as well on an opt-in basis.

Tyriar:
A new setting tree-view.squashDirectoryNames that squashes directories
together when a directory only contains a single directory. Take the following
directory structure:

- foo
  - bar
    - file.txt

foo and bar would be merged in to a single DirectoryView:

- foo/bar
  - file.txt

Notes:
- Squashing occurs recursively so multiple directories can be merged together.
- There is a loss of functionality when the squash is enabled since the
  directories on the left are no longer available to right click. The
  DirectoryView is treated as if it's the full directory.
- The style .squashed-dir was added to allow styling of the left portion of the
  squashed directory ('foo/' in the above example).
- This feature is disabled by default.

Fixes #125

### DIFF
--- a/spec/tree-view-spec.coffee
+++ b/spec/tree-view-spec.coffee
@@ -2164,7 +2164,7 @@ describe "TreeView", ->
         expect(zetaEntries).toEqual(["zeta.txt"])
 
       it "squashes two dir names when the first only contains a single dir", ->
-        betaDir = $(treeView.roots[0].entries).find('.directory:contains(alpha/beta):first')
+        betaDir = $(treeView.roots[0].entries).find(".directory:contains(alpha#{path.sep}beta):first")
         betaDir[0].expand()
         betaEntries = [].slice.call(betaDir[0].children[1].children).map (element) ->
           element.innerText
@@ -2172,7 +2172,7 @@ describe "TreeView", ->
         expect(betaEntries).toEqual(["beta.txt"])
 
       it "squashes three dir names when the first and second only contain single dirs", ->
-        epsilonDir = $(treeView.roots[0].entries).find('.directory:contains(gamma/delta/epsilon):first')
+        epsilonDir = $(treeView.roots[0].entries).find(".directory:contains(gamma#{path.sep}delta#{path.sep}epsilon):first")
         epsilonDir[0].expand()
         epsilonEntries = [].slice.call(epsilonDir[0].children[1].children).map (element) ->
           element.innerText
--- a/spec/tree-view-spec.coffee
+++ b/spec/tree-view-spec.coffee
@@ -2110,6 +2110,83 @@ describe "TreeView", ->
       expect(treeView.find('.directory .name:contains(test.js)').length).toBe 1
       expect(treeView.find('.directory .name:contains(test.txt)').length).toBe 1
 
+  describe "the squashedDirectoryName config option", ->
+    beforeEach ->
+      rootDirPath = fs.absolute(temp.mkdirSync('tree-view'))
+
+      zetaDirPath = path.join(rootDirPath, "zeta")
+      zetaFilePath = path.join(zetaDirPath, "zeta.txt")
+
+      alphaDirPath = path.join(rootDirPath, "alpha")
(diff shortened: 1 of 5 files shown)


Tyriar:
With `.squashed-dir { color: #777; }` added to my stylesheet:

squash_dirs

kevinsawicki:
:heart_eyes:

lee-dohm:
So much awesomeness :zap: :godmode: :metal: :zap:

spec/tree-view-spec.coffee
```diff
@@ -2110,6 +2110,83 @@ describe "TreeView", ->
...
+      fs.makeTreeSync(lambdaDirPath)
+      fs.makeTreeSync(iotaDirPath)
+      fs.makeTreeSync(kappaDirPath)
+
+      atom.project.setPaths([rootDirPath])
+
+    it "defaults to disabled", ->
+      expect(atom.config.get("tree-view.squashDirectoryNames")).toBeFalsy()
+
+    describe "when enabled", ->
```
lee-dohm: Could a test be added that would ensure that the platform-specific path separator is used? From what I've read, I can see that it currently does ... but I wouldn't want it to break in the future.

spec/tree-view-spec.coffee
```diff
@@ -2110,6 +2110,83 @@ describe "TreeView", ->
...
+      it "squashes two dir names when the first only contains a single dir", ->
+        betaDir = $(treeView.roots[0].entries).find('.directory:contains(alpha/beta):first')
+        betaDir[0].expand()
+        betaEntries = [].slice.call(betaDir[0].children[1].children).map (element) ->
+          element.innerText
+
+        expect(betaEntries).toEqual(["beta.txt"])
+
+      it "squashes three dir names when the first and second only contain single dirs", ->
+        epsilonDir = $(treeView.roots[0].entries).find('.directory:contains(gamma/delta/epsilon):first')
```
lee-dohm: Would these tests break when run on a Windows machine because of the forward slashes embedded here?

Tyriar: Good catch, changed to use #{path.sep}

spec/tree-view-spec.coffee
Tyriar: I tried this approach to test that but it doesn't seem to work, I think because it breaks other calls to `path` like `path.join` in directory.coffee.

```coffeescript
describe "when path.sep is overridden", ->
  originalPathSep = path.sep

  beforeEach ->
    path.sep = '#'

  afterEach ->
    path.sep = originalPathSep

  it "should use the overridden path.sep in the UI", ->
    atom.project.setPaths([rootDirPath])
    expect($(treeView.roots[0].entries).find('.directory:contains(alpha#beta):first')).toExist()
```

Any ideas without exposing test a test-only function in directory.coffee?

lee-dohm: Actually, thinking more about this I feel like changing the other tests to use `path.sep` covers this. I'd prefer it to be more explicit as a form of documentation, but I also don't feel injecting a dependency just for the path separator here in this one spot is warranted.

lee-dohm:
Thanks for this!

Tyriar:
No problems, excited to see it land!

thomasjo:
Brilliant work! Thanks :heart:

howdoicomputer:
I can't seem to get this to work for my Java application; with the setting enabled, I still get a billion empty directories.

PR #609, opened 2015-10-17
Add filename to title attribute

### LINKED ISSUE #557: Add tooltip/hover to show full filename when filename is too long to fit within treeview pane and is truncated

johanderuijter:
Fixes #557

I'm not sure if this will require any tests, if it does, i'm going to need some guidance. I checked the tree-view-spec and with it's 2.5k+ lines it's not really beginner friendly.

Let me know if any changes are needed :)

### DIFF
--- a/lib/directory-view.coffee
+++ b/lib/directory-view.coffee
@@ -32,6 +32,7 @@ class DirectoryView extends HTMLElement
         iconClass = 'icon-file-submodule' if @directory.submodule
     @directoryName.classList.add(iconClass)
     @directoryName.dataset.name = @directory.name
+    @directoryName.title = @directory.name
     @directoryName.dataset.path = @directory.path
 
     if @directory.squashedName?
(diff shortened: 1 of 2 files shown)


billybonks:
@benogle can we merge this ?

mnquintana:
Sorry for the delay on reviewing this!

I think this is a good change, but what do you think about adding this for directories too? A directory name could definitely also exceed the width of the tree view.

Also, we generally use the [Atom Tooltip API](https://atom.io/docs/api/v1.1.0/TooltipManager) for things like this, instead of native tooltips. What do you think about displaying tooltips that way?

billybonks:
@mnquintana created a version with the tooltip api take a look and let me know what you think https://github.com/atom/tree-view/pull/610

johanderuijter:
Added the title attribute to directories aswell.

I used the title attribute instead of the tooltips as they seemed less intrusive (and it was suggested in the original issue), but the other PR looks good to. I'm good with either.

billybonks:
Ye to use the tooltip api i would still need to
- Create a setting that turns it off.
- Only show tooltips if the name is hidden by the viewport.
- fix the positioning bug.

mnquintana:
I think this is fine for now - it would be more consistent to use Atom tooltips, but also more finicky and obtrusive. We can always change this to Atom tooltips in a subsequent PR. Thanks for adding this!

PR #610, opened 2015-10-17
Add tooltips to files in tree view

billybonks:
tooltips

https://github.com/atom/tree-view/issues/557

Some points where it can be improved:
 - Add config option to disable tooltips
 - Bind tool tip to a different element so that when the side bar is to small the tooltip doesn't float over text editor
- Make tooltip appear on the right hand side, blocked by the previous point


billybonks:
ill close this in favour of https://github.com/atom/tree-view/pull/609 if its merged, its less intrusive then this pr

Issue #564, opened 2015-09-03
Highlight git changes on the level of a project


jagi:
Hi currently Atom is highlighting files in which there are git changes. However when you collapse project tree, a project name isn't highlighted. Because of that I have to unfold each project's directories tree to check whether I've made some changes. I'm working on several connected projects in the same time so that's why I think such feature would be handy. What do you think? :)

It would also be nice to have, next to the project name, an indicator showing how many commits have been made and are waiting to be pushed and what branch I'm currently on. Such feature is present for instance in Eclipse.

![atom](https://cloud.githubusercontent.com/assets/1051229/9665934/94cee2a8-5273-11e5-8cfb-32267f64d59f.png)

mertkahyaoglu:
:+1:

subesokun:
Happily helping out on that one as I've solved it already in my [tree-view-git-status](https://github.com/subesokun/atom-tree-view-git-status) package. This way I could move this root node highlighting logic into the tree-view and remove it from my package :)

> It would also be nice to have, next to the project name, an indicator showing how many commits have been made and are waiting to be pushed and what branch I'm currently on. Such feature is present for instance in Eclipse.

Not sure if this logic should be part of the tree-view itself as it will blow up a bit the functionality but feel free to use my [tree-view-git-status](https://github.com/subesokun/atom-tree-view-git-status) package as I've implemented it there already :)

jagi:
Thanks! It's exactly what I wanted :)

a7madgamal:
@subesokun life saver, thanks!

a7madgamal:
I still think this should be implemented in this package too

bretonics:
This seems to have already been addressed. I believe this issue should be closed.

50Wliu:
Yep, I think this is better off in a package.  Thanks for stepping up @subesokun :heart:.

PR #614, opened 2015-10-18
Reflect git changes at root level

glangford:
Beginner fix for #564. This change seems to work in interactive testing, but `apm test` reports one failure. I don't know how to rectify this. Any suggestions?

```
 TreeView
  Git status decorations
    when the project is a symbolic link to the repository root
      when a file loses its modified status
        it updates its and its parent directories' styles
```

### DIFF
--- a/lib/directory-view.coffee
+++ b/lib/directory-view.coffee
@@ -40,9 +40,8 @@ class DirectoryView extends HTMLElement
 
     if @directory.isRoot
       @classList.add('project-root')
-    else
-      @subscriptions.add @directory.onDidStatusChange => @updateStatus()
-      @updateStatus()
+    @subscriptions.add @directory.onDidStatusChange => @updateStatus()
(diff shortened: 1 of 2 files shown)


glangford:
I started this looking at tree-view, but perhaps a better fix than #614 is to start at the level of the `GitRepository` class in atom;  `getDirectoryStatus` appears to turn a path at root level into just "/". So directoryStatus always returns a status of zero for the root.

```coffee
getDirectoryStatus: (directoryPath)  ->
    directoryPath = "#{@relativize(directoryPath)}/"  #############
    directoryStatus = 0
    for path, status of @statuses
      directoryStatus |= status if path.indexOf(directoryPath) is 0
    directoryStatus
```

<https://github.com/atom/atom/blob/master/src/git-repository.coffee#L305>

50Wliu:
Superseded by #1144.

PR #617, opened 2015-10-22
add option for click to focus editor

jacekkopecky:
Often I use the mouse to navigate between files – I point to the file, click on it, look at it, and expect to be editing it but the focus is on tree-view. This PR adds an option that a single click on a file focuses the new text editor.

Comments welcome – would others like this option?

### DIFF
--- a/lib/tree-view.coffee
+++ b/lib/tree-view.coffee
@@ -216,9 +216,10 @@ class TreeView extends View
     filePath = e.currentTarget.getPath()
     detail = e.originalEvent?.detail ? 1
     alwaysOpenExisting = atom.config.get('tree-view.alwaysOpenExisting')
+    activateOnClick = atom.config.get('tree-view.clickFocusesEditor')
     if detail is 1
       if atom.config.get('core.allowPendingPaneItems')
-        openPromise = atom.workspace.open(filePath, pending: true, activatePane: false, searchAllPanes: alwaysOpenExisting)
+        openPromise = atom.workspace.open(filePath, pending: true, activatePane: activateOnClick, searchAllPanes: alwaysOpenExisting)
         @currentlyOpening.set(filePath, openPromise)
         openPromise.then => @currentlyOpening.delete(filePath)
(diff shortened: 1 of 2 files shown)


jacekkopecky:
Any comments, please?

ewoutp:
This is exactly what I'm looking for.

jacekkopecky:
Updated to v208, don't understand why appveyor build failed.
Please comment.

50Wliu:
AppVeyor was prematurely activated on tree-view, which doesn't have Windows spec compatibility yet.

bronson:
I love this PR except I think it should be the default behavior.  No need for an option.

It appears ternjs is giving me this functionality: https://github.com/tststs/atom-ternjs/issues/264  So that's an easy fix until this PR is merged.

BinaryMuse:
Thank you for the PR, @jacekkopecky!

@bronson, #749 is proof that not everyone agrees with that workflow. ;) We won't be changing the default behavior again.

As for the option, #749 and @lee-dohm's comments in #804 explain why we're loathe to add yet more options that modify the behavior of opening items from the tree view.

If you would like to focus the editor when it's opened via single-click from the tree view, you have a few options currently:

* Double-click the entry, which opens the file as a permanent item instead of a pending item
* Use a package like [single-click-open](https://atom.io/packages/single-click-open)
* Add a bit of glue code to your `init.coffee`:

    ```coffee
    atom.workspace.onDidAddPaneItem ({item, pane}) ->
      pane.setPendingItem(null)
      process.nextTick () => atom.views.getView(item).focus()
    ```

bronson:
That makes sense, thanks for the fantastic explanation @BinaryMuse!

jacekkopecky:
Thank you, yes, single-click-open does the trick.

bronson:
Sadly, not for me: https://github.com/olmokramer/atom-single-click-open/issues/3

Hope the maintainer comes back soon.

Good to see you @jacekkopecky

PR #618, opened 2015-10-22
Update folder git status colour after delete

### LINKED ISSUE #498: Folder git status colour doesn't update after delete
As reported [here](https://github.com/atom/atom/issues/7589):

> Create a new file and the directory hierarchy above will change to be green. If you then delete the file the folder colours do not revert to their original even if no other files have been amended or added.

Before deletion of file2:
screenshot 2015-07-01 14 21 05
After deletion of file2:
screenshot 2015-07-01 14 21 23

This seems to fix itself when you tab away from and back to Atom.

glangford:
Fix #498

### DIFF
--- a/lib/tree-view.coffee
+++ b/lib/tree-view.coffee
@@ -523,6 +523,8 @@ class TreeView extends View
         "Move to Trash": ->
           for selectedPath in selectedPaths
             shell.moveItemToTrash(selectedPath)
+            if repo = repoForPath(selectedPath)
+              repo.getPathStatus(selectedPath)
         "Cancel": null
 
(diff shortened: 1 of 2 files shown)


mnquintana:
Would you mind adding specs for this? Thanks!

glangford:
@mnquintana To write a spec I need to have an unchanged git directory as a starting point. I don't see an obvious way of doing this given the existing tree-view specs. Any suggestions?

glangford:
Specs added.

kevinsawicki:
Thanks for fixing this :+1:

PR #619, opened 2015-10-24
Implement multi-selection move (#232)

### LINKED ISSUE #232: Move multiple files.
I would like to see the possibility to move multiple files at once.

I think it could work like this:
1. Select multiple files in the tree view
2. press m (default move command)
3. The prompt asks you for a new folder (just like now, but obviously without the file name, since you are moving multiple files).
4. All selected files are move to your chosen directory.

thomdixon:
This implements multi-selection move through the move dialog, which resolves #232. In the future, drag-and-drop moves should be implemented.


thomdixon:
@Tyriar nice catch! Added the move item to the context menu for multi-selects.

50Wliu:
Superseded by #1179.

PR #621, opened 2015-10-25
Adds a fast implementation of natural sort

billybonks:
I am, currently using a git url in `package.json`, because, i made some modifications and would prefer to get my pr merged into the master repo, rather then creating a whole new npm package. https://github.com/litejs/natural-compare-lite/pull/7
 
The implementation is only concerned with alphanumeric sorting.

ie.
```
file1.txt
file2.txt
file100.txt
```
over
```
file1.txt
file100.txt
file2.txt
```
The commit also adds an option to use the old sort function, incase
this one doesn't work with other locales, or if users prefer posix style
sorting.

### DIFF
--- a/lib/directory.coffee
+++ b/lib/directory.coffee
@@ -5,7 +5,7 @@ fs = require 'fs-plus'
 PathWatcher = require 'pathwatcher'
 File = require './file'
 {repoForPath} = require './helpers'
-
+naturalCompare = require 'natural-compare-lite'
 realpathCache = {}
 
 module.exports =
@@ -159,8 +159,9 @@ class Directory
       names = fs.readdirSync(@path)
     catch error
       names = []
-    names.sort (name1, name2) -> name1.toLowerCase().localeCompare(name2.toLowerCase())
-    
+
+    names.sort(naturalCompare)
+
     files = []
     directories = []
 
--- a/spec/tree-view-spec.coffee
+++ b/spec/tree-view-spec.coffee
@@ -397,13 +397,13 @@ describe "TreeView", ->
 
       it 'scrolls the selected file into the visible view', ->
         # Open file at bottom
-        waitsForPromise -> atom.workspace.open(path.join(rootDirPath, 'file-9.txt'))
+        waitsForPromise -> atom.workspace.open(path.join(rootDirPath, 'file-20.txt'))
         runs ->
           atom.commands.dispatch(workspaceElement, 'tree-view:reveal-active-file')
           expect(treeView.scrollTop()).toBeGreaterThan 400
 
         # Open file in the middle, should be centered in scroll
-        waitsForPromise -> atom.workspace.open(path.join(rootDirPath, 'file-19.txt'))
+        waitsForPromise -> atom.workspace.open(path.join(rootDirPath, 'file-10.txt'))
         runs ->
           atom.commands.dispatch(workspaceElement, 'tree-view:reveal-active-file')
           expect(treeView.scrollTop()).toBeLessThan 400
@@ -414,7 +414,6 @@ describe "TreeView", ->
         runs ->
           atom.commands.dispatch(workspaceElement, 'tree-view:reveal-active-file')
           expect(treeView.scrollTop()).toEqual 0
-
 
   describe "when tool-panel:unfocus is triggered on the tree view", ->
     it "surrenders focus to the workspace but remains open", ->
--- a/package.json
+++ b/package.json
@@ -14,6 +14,7 @@
     "event-kit": "^1.0.0",
     "fs-plus": "^2.3.0",
     "minimatch": "~0.3.0",
+    "natural-compare-lite": "^1.4.0",
     "pathwatcher": "^6.2",
     "temp": "~0.8.1",
     "underscore-plus": "^1.0.0"


lib/directory.coffee
```diff
@@ -159,8 +159,11 @@ class Directory
       names = fs.readdirSync(@path)
     catch error
       names = []
-    names.sort (name1, name2) -> name1.toLowerCase().localeCompare(name2.toLowerCase())
-    
+    unless atom.config.get('tree-view.posixSort')
+      names.sort(naturalCompare);
```
thomasjo: :fire: the semicolon

lib/directory.coffee
```diff
@@ -159,8 +159,11 @@ class Directory
...
+    else
+     names.sort (name1, name2) -> name1.toLowerCase().localeCompare(name2.toLowerCase())
```
thomasjo: Bad indentation — one space is missing here.

lib/main.coffee
```diff
@@ -22,6 +22,10 @@ module.exports =
       type: 'boolean'
       default: true
       description: 'When listing directory items, list subdirectories before listing files.'
+    posixSort:
```
thomasjo: I think it would make sense to change this to `alphanumericSort` with a default of `true`. Thus in essence flipping the current logic.

I'm pretty sure there's nothing in the POSIX standard relating to sort order of directory listings, so calling it `posixSort` is a bit of a misnomer either way.

thomasjo:
All in all this looks good, but obviously the repo URL should be changed (as you mentioned), so merging this will have to be postponed until some action is taken in that regard.

I've left a few line comments, two of which are related to lint, and one related to changing the `posixSort` option into `alphanumericSort` instead and flipping the logic in the code.

Also, since you've introduced an option for selecting the sort order, we need to add new specs. We have to ensure that both types of sort work as intended to protect against regressions. So basically duplicate the existing sort test, and make sure that it uses the non-default sort option, and that the sort order is as expected (the current implementation).

Thanks for taking this on!

lib/main.coffee
billybonks: no problem ill find better nomenclature :)

lib/directory.coffee
billybonks: the struggles when switching between javascript and coffeescript all day long :)

billybonks:
:+1:  will add the specs :)

p.s sorry for not running the linting specs before hand, having some troubles with my atom dev environment that prevents me from running apm test. i will sort them out asap.

thomasjo:
> p.s sorry for not running the linting specs before hand, having some troubles with my atom dev environment that prevents me from running apm test. i will sort them out asap.

No problem, that's why we have automated checks :smile:

Linting is not actually executed as part of the test suite though, but you can easily lint your changes by doing e.g.

```bash
$(npm bin)/coffeelint .
```

from the repo root.

billybonks:
@thomasjo as it stands when the config is changed the user must collapse and expand the tree to view updated sort order. should we include a didChange listener to the config, to refresh the tree when they toggle the setting ?

lib/directory.coffee
```diff
@@ -159,8 +159,12 @@ class Directory
       names = fs.readdirSync(@path)
     catch error
       names = []
-    names.sort (name1, name2) -> name1.toLowerCase().localeCompare(name2.toLowerCase())
-    
+
+    if atom.config.get('tree-view.alphanumericSort')
```
nathansobo: Should the option be called `naturalSort` instead? Sort is *always* alphanumeric isn't it? The difference here is the natural part.

spec/tree-view-spec.coffee
```diff
@@ -394,27 +394,52 @@ describe "TreeView", ->
         atom.project.setPaths([rootDirPath])
         treeView.height(100)
         jasmine.attachToDOM(workspaceElement)
+      describe "if alphanumericSort is set to true", ->
```
nathansobo: Rather than testing this as a dimension of the `reveal-active-file` command, I think it probably warrants a new top-level test like the following:

```coffee
describe "the tree-view.naturalSort option", ->
  it "determines whether tree view items are sorted naturally or in strict alphanumeric order", ->
    # load up the tree view with an example that sorts differently based on the option
    # start with the option one way, assert the order
    # toggle the option, assert the order again
```

And leave it at that. Feel free to interpret my suggested test titles a bit. It's just a rough example.

lib/main.coffee
```diff
@@ -22,6 +22,10 @@ module.exports =
       type: 'boolean'
       default: true
       description: 'When listing directory items, list subdirectories before listing files.'
+    alphanumericSort:
+      type: 'boolean'
+      default: true
+      description: 'By default files/folders are sorted alphanumerically, ticking this option will sort them based on ascii values, using String.localeCompare.'
```
nathansobo: Again, is there a way to make this description a bit clearer? To me alphanumeric and ASCII sorting are pretty much the same thing. I think it warrants emphasis of the "natural" nature somehow.

```
Sort numbers and versions in natural rather than strict lexical order. For example, `1` would sort before `10` even though it's a shorter string.
```

Feel free to adjust my wording here to cover what natural-sort lite actually does.

nathansobo:
Great job on this @billybonks. One question... is this fast enough now to just always do like we did previously? Then we can dispense with the option and save some complexity. It looks like your PR on natural-sort-lite might have been merged in https://github.com/litejs/natural-compare-lite/commit/eec83eee67cfac84d6db30cdd65363f155673770.

If we're fast enough to drop the option, I almost don't think we need explicit tests of the sort order. The existing tests just exercising the sorting indirectly seem fine to me, and presumably there are already more detailed tests on the library itself.

billybonks:
Sounds, good we are definitely fast enough, to drop the option.

nathansobo:
So, looks like the tests are passing and the code is pretty straightforward. Are you ready for me to merge this?

nathansobo:
Thanks @billybonks! This is going to be nice to have back without the slowness. Good research.

jeancroy:
Some comments:

1. It looks like the `natural-compare-lite` library mostly deal with a-zA-Z0-9 range.
Basically this loose whatever internationalization `localeCompare` brought.

2. Locale compare have a `kn` flag to support number, also `kf` to support case insensitive.

3. For speed Intl.Collator() is to be preferred to `a.localeCompare(b)` . The above flag correspond to numeric and caseFirst options.

````
comparator = (new Intl.Collator()).compare
array.sort(comparator)
````

nathansobo:
Good points. Sounds like we should probably back this out and try again. Always need to remember international support. Thanks @jeancroy.

billybonks:
I forgot i added the locale compare option, incase of broken locales. but it seems, the kn flag will solve both problems, with the same stone. great find :+1: @nathansobo @jeancroy open a pr that does it this way tomorrow.

nathansobo:
@billybonks How we doing here... should I back this out? Seems like breaking international compare isn't worth natural sort for now.

billybonks:
implementing a version of natural sort with localeCompare is easy if we
know the language that the user is using

`['1','100','2'].sort( new Intl.Collator('en',{numeric:true}).compare)`.

i found a node package that i could use to do this. does atom use anything
at the moment ?

jeancroy:
I believe the language specification is optional.

billybonks:
ye, this works `['1','2','100'].sort( new Intl.Collator('kn').compare)`

billybonks:
i lie, it doesnt.

billybonks:
the mdn docs says its optional. but if you specify options. then it seems
to become required. any ideas?

jeancroy:
It looks like a hack but this will give you default locale

(new Intl.Collator()).resolvedOptions().locale

nathansobo:
There is a `navigator.language` feature: https://github.com/atom/electron/pull/363

billybonks:
ok, that seems good.

we can replace the sort with `new Intl.Collator(navigator.language).compare`

nathansobo:
@billybonks: @jeancroy's solution might be better because I'm not sure `navigator.language` gives you a string that seems match what `Intl.Collator` expects. You'll have to investigate.

jeancroy:
Intl.collator expect somthing that is [bcp47](https://tools.ietf.org/html/bcp47)
`navigator.language`provide a compatible string.

`navigator.locale` I cannot find. Is it atom specific ?

nathansobo:
That's weird because I'm getting `"English"` from `navigator.language`. I would have expected something more like `"en-US"`.

jeancroy:
From both chrome and atom dev tool console I get "en-US".
Maybe it's a platform thing? I'm on windows.

billybonks:
i got en-US on mac. what are you using @nathansobo

PR #622, opened 2015-10-25
Select first sub-entry on expand-directory if already expanded

### LINKED ISSUE #561: Select first sub-folder when pressing → (right-arrow) on already expanded folder
Example

Pressing the right-arrow key would select the first (folder) item inside the currently selected folder.

I just noticed that it's not supported....

Tyriar:
Enable faster, more natural navigation by having the tree-view.expand-directory
command select the first sub-entry when the directory is already expanded. This
new behavior applies only when a non-recursive directory expand is triggered
from the tree-view.

Fixes #561

### DIFF
--- a/lib/tree-view.coffee
+++ b/lib/tree-view.coffee
@@ -330,7 +330,7 @@ class TreeView extends View
     @selectEntry(@entryForPath(entryPath))
 
   moveDown: (event) ->
-    event.stopImmediatePropagation()
+    event?.stopImmediatePropagation()
     selectedEntry = @selectedEntry()
     if selectedEntry?
       if selectedEntry instanceof DirectoryView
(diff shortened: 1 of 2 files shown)


Tyriar:
Ping?

lee-dohm:
What happens when I single-click the directory with the mouse? Does it change my selection out from under me?

Tyriar:
@lee-dohm it toggles the directory like the previous behavior.

lee-dohm:
Cool thanks :grinning: I've added this to the queue of things for the team to review.

kuychaco:
@Tyriar - at first glance this looks good :+1:. If you can resolve the conflicts with master I can go ahead and merge this. Thanks for another great PR :)

Tyriar:
@kuychaco don't have the time to fix up the conflicts unfortunately, going to close this off.

PR #623, opened 2015-10-25
Add enable drag and drop option

### LINKED ISSUE #588: Option to Disable Folder Dragging
Just a second ago I accidentally clicked and dragged a folded somewhere and had no clue where. Had my source not been in version control this would have been super annoying! Is there any way we could implement a checkbox for disabling this feature?

Tyriar:
Fixes #588

### DIFF
--- a/spec/tree-view-spec.coffee
+++ b/spec/tree-view-spec.coffee
@@ -2419,6 +2419,39 @@ describe "TreeView", ->
               expect(treeView.list).toHaveClass('full-menu')
               expect(treeView.list).not.toHaveClass('multi-select')
 
+  describe "the enableDragAndDrop config option", ->
+    beforeEach ->
+      rootDirPath = fs.absolute(temp.mkdirSync('tree-view'))
+      alphaFilePath = path.join(rootDirPath, "alpha.txt")
+      alphaDirPath = path.join(rootDirPath, "alpha")
+
+      fs.writeFileSync(alphaFilePath, "doesn't matter")
+      fs.makeTreeSync(alphaDirPath)
+      atom.project.setPaths([rootDirPath])
+
+    it "defaults to set", ->
+      expect(atom.config.get("tree-view.enableDragAndDrop")).toBeTruthy()
+
+    describe "when enabled", ->
+      beforeEach ->
+        atom.config.set("tree-view.enableDragAndDrop", true)
+
+      it "makes FileView draggable", ->
+        expect(treeView.roots[0].entries.children[0].getAttribute('draggable')).toBe("true")
+
+      it "makes DirectoryView draggable", ->
+        expect(treeView.roots[0].entries.children[1].getAttribute('draggable')).toBe("true")
+
+    describe "when disabled", ->
+      beforeEach ->
+        atom.config.set("tree-view.enableDragAndDrop", false)
+
+      it "makes FileView non-draggable", ->
+        expect(treeView.roots[0].entries.children[0].getAttribute('draggable')).toBe('false')
+
+      it "makes DirectoryView non-draggable", ->
+        expect(treeView.roots[0].entries.children[1].getAttribute('draggable')).toBe('false')
+
   describe "the sortFoldersBeforeFiles config option", ->
     [dirView, fileView, dirView2, fileView2, fileView3, rootDirPath, dirPath, filePath, dirPath2, filePath2, filePath3] = []
 
--- a/spec/tree-view-spec.coffee
+++ b/spec/tree-view-spec.coffee
@@ -2429,7 +2429,7 @@ describe "TreeView", ->
       fs.makeTreeSync(alphaDirPath)
       atom.project.setPaths([rootDirPath])
 
-    it "defaults to set", ->
+    it "defaults to true", ->
       expect(atom.config.get("tree-view.enableDragAndDrop")).toBeTruthy()
 
     describe "when enabled", ->
--- a/lib/directory-view.coffee
+++ b/lib/directory-view.coffee
@@ -9,7 +9,7 @@ class DirectoryView extends HTMLElement
     @subscriptions.add @directory.onDidDestroy => @subscriptions.dispose()
     @subscribeToDirectory()
 
-    @draggable = true
+    @draggable = atom.config.get('tree-view.enableDragAndDrop')
 
     @classList.add('directory', 'entry',  'list-nested-item',  'collapsed')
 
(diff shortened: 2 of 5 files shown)


spec/tree-view-spec.coffee
```diff
@@ -2419,6 +2419,39 @@ describe "TreeView", ->
...
+    beforeEach ->
+      rootDirPath = fs.absolute(temp.mkdirSync('tree-view'))
+      alphaFilePath = path.join(rootDirPath, "alpha.txt")
+      alphaDirPath = path.join(rootDirPath, "alpha")
+
+      fs.writeFileSync(alphaFilePath, "doesn't matter")
+      fs.makeTreeSync(alphaDirPath)
+      atom.project.setPaths([rootDirPath])
+
+    it "defaults to set", ->
```
50Wliu: `it "defaults to true"`

50Wliu:
Looks good to me!  /cc @atom/feedback

benogle:
I feel like the solution is not in adding yet another checkbox, but in making it harder to screw up during drag and drop. Maybe it's too sensitive? Maybe there should be some kind of confirmation when moving a folder?

benogle:
So, I'm :-1: on this for now.

thedaniel:
I'm also :-1:, and not just because I'm allergic to adding checkboxes in general. I think we can fix DnD problems with improved UX. Perhaps expanding to and highlighting the dropped item after its move so you can't lose it?

benogle:
Closing as a wontfix

ryanpalmerweb:
Perfect - thanks

I was going to drop using atom and go back to sublime until I found this.

It's a massive issue, shame on the atom team for not opening up a way to fix this with the core.

garfieldbanks:
Steps to disable drag and drop (for linux):
1. Find the version of tree-view you are currently using in atom (mine was 0.198.0)
2. Go to https://github.com/atom/tree-view/releases and find the commit that corresponds to your version of tree-view (for me it was 01ad2a9)
3. Run these commands:
git clone https://github.com/atom/tree-view.git
cd tree-view
git checkout 01ad2a9
apm link
4. Replace all instances of @draggable = true with @draggable = false
5. Run more commands:
apm link
npm install
npm link
apm rebuild
6. Every time you make a change to tree-view you have to run these again:
npm install
apm rebuild

ponyesteves:
Hello, I see this PR was closed because there is a better wat of solving it, yet de D&D problem is still a problem. Is this topic being debates elsewhere, so I can follow? 

Thanks

ryanpalmerweb:
Hi Jose,

I modified the tree view package to do this.

I've attached the package. 

I'm on OSX, it lives in /Users/{my username}/.atom/packages

Thanks,
Ryan

[tree-view-nodnd.zip](https://github.com/atom/tree-view/files/199432/tree-view-nodnd.zip)

kewball:
How about merging this PR until the UX actually gets fixed? Don't let the perfect solution be the enemy of something that's merely good.

strindhaug:
Why is this wontfix? I hate drag and drop, it's almost impossible to click in the tree view without accidentally dragging and dropping some times, and there is not even any way to undo it or confirming the move.

Being able to disable this is a MUST HAVE feature! 
Is there at least any way of fixing this bug with a user script? I don't care if it's a checkbox or not, as long as I can permanently disable this misfeature. (if I want to move files I move them in my terminal)

ryanpalmerweb:
@strindhaug 

There are various ways to fix it highlighted above (such as the modified tree view package I posted earlier in the thread).

Agreed that this should be part of the core however.

strindhaug:
@ryanpalmerweb: It didn't work for me, it somehow just disabled the tree view from working completely until i removed it.

The real problem is that there is NO drag-and-drop threshold at all, move one pixel while clicking and you're allready dragging, move just a few px more depending on where you clicked and you've moved a folder somewehere else, instantly and without any confirmation or visual clue where it went.

Even if they insist on not allowing us to disable drag and drop completely; there NEED to be a drag and drop threshold setting (with no upper limit so that I can effectively disable it by setting it to 500px or whatever). 1 pixel threshold is ridiculous! That's less than 0.1mm movement both on screen and with my physical mouse. I'm having problems not accidentally dragging stuff when trying to click, and I'm not even 40 years old; imagine how anyone with even a little tremor would do...

ponyesteves:
@ryanpalmerweb: thank's for sharing I'll try it and let you know.

I think like @strindhaug, this is really a problem that must be solved or by incrementing threshold or adding the option on settings.

ryanpalmerweb:
Hi Ponyesteves,

I've forked from a more recent version of tree-view.

Add the contents of the attached zip to ~/.atom/packages/tree-view/

Remember that new versions of the original tree-view version will overwrite it when you run an update, so it does make sense to rename the package.

Thanks,
Ryan

[tree-view.zip](https://github.com/atom/tree-view/files/411145/tree-view.zip)

ponyesteves:
thanks for the advice Ryan!

warren-bank:
Sorry to bump an old thread, but for anyone still interested in this topic (as I was).. I just added a [repo](https://github.com/warren-bank/atom-tree-view-nodnd) that includes 2 bash scripts.
* The 1st: downloads a (configurable) version of "tree-view" from github, unzips it, applies a few patches that make inplace edits to replicate (most of) the [changes suggested by Tyriar](https://github.com/atom/tree-view/pull/623/commits/6614d3417b5ac31e8fc00550092fa0947c2df068), then zips up a modified archive.
* The 2nd: copies the modified archive to Atom's packages directory, unzips it, then installs it.

If the core dev team has no intention to implement this kind of configuration option.. to disable a feature that's unwanted by many, then having available an (innocuous) script that can universally patch any (current) release seems like a decent alternate method of delivery.

ahfarmer:
I think like a lot of the other commenters on this thread, I don't use drag-and-drop in Atom's tree view at all. To me it is only a source of pain. Would love to be able to disable it.

@benogle would you consider re-opening? The solution you envisioned hasn't happened in almost 2 years and there is no sign that it ever will. Improving drag-and-drop only makes sense if users actually want it - and as you can see above, many do not.

ryanpalmerweb:
@benogle I'm crying out for this, still. 

Multiple issues have been caused in my code by inadvertently dragging and dropping files and not realising. 

I have to patch every single update to the tree-view module. 

I don't have this issue anywhere else, on any application or operating system. 

Why can't you admit there is a problem here, just because you are not affected personally?

benyanke:
Considering there's still been no movement on an actual fix here, and the official stance has been "It adds complexity, and we want to fix it right by making the UI better," is there now any openness to adding a simple checkbox *until* the UI is improved? 

A great many have no desire to move files within their editor. That's what a terminal and file manager are for.

iolsen:
Given that we're not going to take on significant work on the tree view any time soon, adding this checkbox, while less than ideal, feels like the right thing now.

I can't re-open this PR because the source repo is gone, but we'd accept a good PR adding this. I'll reopen #588.

PR #624, opened 2015-10-26
Use asynchronous repo

thedaniel:
tree-view is the best candidate for exercising the new async repo class we’re working on in https://github.com/atom/atom/pull/9213

This depends on that branch of atom/atom. The plan is to merge the async repo after it proves sufficient for tree-view’s consumption, then merge this.

### DIFF
--- a/lib/directory-view.coffee
+++ b/lib/directory-view.coffee
@@ -113,26 +113,46 @@ class DirectoryView extends HTMLElement
     if @isExpanded then @collapse(isRecursive) else @expand(isRecursive)
 
   expand: (isRecursive=false) ->
+    expansionPromise = null
     unless @isExpanded
       @isExpanded = true
       @classList.add('expanded')
       @classList.remove('collapsed')
-      @directory.expand()
+      expansionPromise = @directory.expand()
 
     if isRecursive
-      for entry in @entries.children when entry instanceof DirectoryView
-        entry.expand(true)
+      children = (entry for entry in @entries.children when entry instanceof DirectoryView)
+      firstEntry = children.shift()
+      if firstEntry
+        initial = if expansionPromise
+          expansionPromise.then => first.expandAsync()
+        else
+          first.expandAsync()
+        reducer = (curr, next) =>
+          curr.then =>
+            next.expand(true)
+        children.reduce reducer, initial
+      else
+        initial = if expansionPromise
+          expansionPromise.then => @expandAsync()
+        else
+          @expandAsync()
 
+        initial.then (children) =>
+          for child in children when child instanceof DirectoryView
+            child.expand(true)
     false
 
   # Non-recursive ::expand that returns the promise from the model, for use in
   # recursively revealing the active file
   expandAsync: ->
-    unless @isExpanded
+    if @isExpanded
+      return Promise.resolve(@entries.children)
+    else
       @isExpanded = true
       @classList.add('expanded')
       @classList.remove('collapsed')
-      return @directory.expand()
+      return @directory.expand().then => @entries.children
 
   collapse: (isRecursive=false) ->
     @isExpanded = false
--- a/lib/directory-view.coffee
+++ b/lib/directory-view.coffee
@@ -30,9 +30,12 @@ class DirectoryView extends HTMLElement
     else
       iconClass = 'icon-file-directory'
       if @directory.isRoot
-        iconClass = 'icon-repo' if repoForPath(@directory.path)?.isProjectAtRoot()
+        repoForPath(@directory.path)?.isProjectAtRoot().then (projectAtRoot) =>
+          @directoryName.classList.add('icon-repo') if projectAtRoot
       else
-        iconClass = 'icon-file-submodule' if @directory.submodule
+        repoForPath(@directory.path)?.isSubmodule(@directory.path).then (isSubmodule) =>
+          @directoryName.classList.add('icon-file-submodule') if isSubmodule
+
     @directoryName.classList.add(iconClass)
     @directoryName.textContent = @directory.name
     @directoryName.dataset.name = @directory.name
--- a/lib/directory-view.coffee
+++ b/lib/directory-view.coffee
@@ -44,14 +44,17 @@ class DirectoryView extends HTMLElement
     if @directory.isRoot
       @classList.add('project-root')
     else
-      @subscriptions.add @directory.onDidStatusChange => @updateStatus()
+      @subscriptions.add @directory.onDidStatusChange => @updateStatus(arguments[0])
       @updateStatus()
 
     @expand() if @directory.expansionState.isExpanded
 
-  updateStatus: ->
+  updateStatus: (status) ->
     @classList.remove('status-ignored', 'status-modified', 'status-added')
-    @classList.add("status-#{@directory.status}") if @directory.status?
+    if status?
+      @classList.add("status-#{status}")
+    else
+      @classList.add("status-#{@directory.status}") if @directory.status?
 
   subscribeToDirectory: ->
     @subscriptions.add @directory.onDidAddEntries (addedEntries) =>
--- a/lib/directory-view.coffee
+++ b/lib/directory-view.coffee
@@ -125,6 +125,15 @@ class DirectoryView extends HTMLElement
 
     false
 
+  # Non-recursive ::expand that returns the promise from the model, for use in
+  # recursively revealing the active file
+  expandAsync: ->
+    unless @isExpanded
+      @isExpanded = true
+      @classList.add('expanded')
+      @classList.remove('collapsed')
+      return @directory.expand()
+
   collapse: (isRecursive=false) ->
     @isExpanded = false
 
--- a/lib/directory-view.coffee
+++ b/lib/directory-view.coffee
@@ -113,32 +113,26 @@ class DirectoryView extends HTMLElement
     if @isExpanded then @collapse(isRecursive) else @expand(isRecursive)
 
   expand: (isRecursive=false) ->
-    expansionPromise = null
+    initial = Promise.resolve(true)
     unless @isExpanded
       @isExpanded = true
       @classList.add('expanded')
       @classList.remove('collapsed')
-      expansionPromise = @directory.expand()
+      initial = initial.then => @directory.expand()
 
     if isRecursive
       children = (entry for entry in @entries.children when entry instanceof DirectoryView)
       firstEntry = children.shift()
       if firstEntry
-        initial = if expansionPromise
-          expansionPromise.then => first.expandAsync()
-        else
-          first.expandAsync()
+        initial.then => first.expandAsync()
         reducer = (curr, next) =>
           curr.then =>
             next.expand(true)
         children.reduce reducer, initial
       else
-        initial = if expansionPromise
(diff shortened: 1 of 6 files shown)


thedaniel:
Hm, I only have 2 (seemingly unrelated) failures on my machine, but there are a bunch of :red_circle: on travis. In any case this is starting to come together nicely.

thedaniel:
Of course, it's never as simple as 'just make git decorations async' - this refactor ended up touching a bunch of parts of the tree-view (which is fine, more async is better, right?) but it's been a slog. I'm hoping most of these remaining red dots are just specs that need to be updated to wait on asynchronous directory expansion, not bugs or races w/ the async git repo or within this package's views.

screen shot 2015-12-24 at 5 49 42 pm

thedaniel:
This is green on my machine :tada: with this change to atom - https://github.com/atom/atom/pull/10352 though it seems that PR needs a little more attention before it can merge.

screen shot 2016-01-09 at 12 45 13 pm

The specs are deeply littered with waitsfor/runs grossness, but it seemed to be the most direct route to getting them green with this big refactor. In the future we should consider breaking up the spec sections into separate files and porting them to ES6 one by one.

Also there are a few opportunities to asyncify this further but given the original intention was to just start using the async git stuff, i figured more refactoring should be done in a separate PR rather than mixing it in here.

cc @atom/feedback - this isn't quite ready to merge but let's have a review anyway :smile_cat:

lib/tree-view.coffee
```diff
@@ -688,6 +728,8 @@ class TreeView extends View
       fs.moveSync(initialPath, newPath)
 
       if repo = repoForPath(newPath)
+        # The return values of this are thrown away, they are called for emiiter
+        # side effects only.
```
joshaber: `GitRepositoryAsync` has `.refreshStatusForPath` which does this, just with a clearer name.

joshaber:
:metal: Just :one: :memo:

thedaniel:
cc: @atom/feedback @atom/core 

Items for review / notable changes:

`tree-view.coffee`

- See big comment block at top of `initialize()` explaining use of a mutation observer that makes me feel not great.
- I serialized promises using Array.reduce in `initialize()` as we needed to expand recursively until we find (or don't find) a view for the path that's supposed to be selected on open.

`directory-view.coffee`

- I refactored the synchronous `expand()` to use a helper method called `expandAsync()` rather than doing the promise chaining inline, and realized that `expandAsync()` inside `waitsForPromise` is nice in tests. The old `expand()` returned false when done, and I kept that even though it is no longer sync, in case there were callers out there expecting that return value. Not sure if this will cause problems or if I should change the return value to a promise there as well.

`directory.coffee`

- `updateStatus()` used a lot of sync methods and got a lot more complicated using the async repo. It could use a sanity check.
- `getEntries()` really blew up in complexity. It was a very simple "loop over files and call sync status checks on each one" and I ultimately chose to build an array of promises checking statuses and then resolving them with Promise.all and sorting the last step of the promise chain. It seems fine, but the method doubled in size so there might be a clearer way to do it.
- I just changed `reload()` from sync to async, assuming that since tree-view doesn't really have a public API #itsfine.
- same with `expand()`

`file.coffee`

- `updateStatus()` became a 3-promise-chain async method with a bunch of conditionals, due to use of async staus checking methods

`tree-view-spec.coffee`

- So much ugly and repetitive conversion waitsFor and waitsForPromise :/ . In a dream world I'd love to break this up into 10 different spec files and rewrite them in ES6 using async/await but that's out of scope esp since it seems we're going to need to rewrite tree-view anyway to provide some kind of API to package developers.

lib/tree-view.coffee
```diff
@@ -53,7 +53,34 @@ class TreeView extends View
     @updateRoots(state.directoryExpansionStates)
     @selectEntry(@roots[0])
 
-    @selectEntryForPath(state.selectedPath) if state.selectedPath
+    # Now that directory expansion is done asyncly combining 2 different
```
nathansobo: :grimacing: This is why it would be really fantastic to have a view model.

lib/tree-view.coffee
```diff
@@ -301,15 +328,29 @@ class TreeView extends View
...
+    reducer = (currentItemPromise, nextItem) =>
+      return currentItemPromise.then (currentPath) =>
+        processPathComponent(nextItem, currentPath)
+
+    # Porting git status to async means that some of the internal tree-view APIs
+    # needed to go async as well. However, certain parts of the codebase (including
+    # this function) iterate over lists of values - generally we can just use
+    # Promise.all in those cases, but this one works its way sequentially through
+    # the list, so we use Array::reduce to serialize the promises.
+    activePathComponents.reduce reducer, Promise.resolve(rootPath)
```
nathansobo: Interesting use of `reduce` for sure. More wishful thinking, would be elegantly expressed in Babel with a for loop wrapping `await`. Oh well.

lib/file.coffee
```diff
@@ -35,32 +35,33 @@ class File
...
     repo = repoForPath(@path)
-    return unless repo?
-
-    newStatus = null
-    if repo.isPathIgnored(@path)
-      newStatus = 'ignored'
-    else
-      status = repo.getCachedPathStatus(@path)
-      if repo.isStatusModified(status)
+    repo?.isPathIgnored(@path).then (isIgnored) =>
```
nathansobo: This all makes sense to me.

lib/directory.coffee
```diff
@@ -30,8 +32,6 @@ class Directory
     @status = null
     @entries = {}
 
-    @submodule = repoForPath(@path)?.isSubmodule(@path)
```
nathansobo: Are we eliminating the `submodule` distinction with the new repository API or has it just moved elsewhere?

lib/directory.coffee
```diff
@@ -163,34 +165,63 @@ class Directory
...
-          # track the insertion index for the created views
-          files.push(name)
-        else
-          files.push(new File({name, fullPath, symlink, realpathCache}))
-
-    @sortEntries(directories.concat(files))
+      # throw this in a closure so we don't have scoping issues
+      # async/await would make this more pleasant when this is
+      # inevitably ported to ES6
+      f = =>
```
nathansobo: Would Coffee's `do` syntax help here at all for an anonymous immediately-invoked function?

lib/directory.coffee
```diff
@@ -163,34 +165,63 @@ class Directory
...
+          return if isIgnored
+
+          stat = fs.lstatSyncNoException(fullPath)
+          symlink = stat.isSymbolicLink?()
+          stat = fs.statSyncNoException(fullPath) if symlink
+          if stat.isDirectory?()
+            if @entries.hasOwnProperty(localName)
+              # push a placeholder since this entry already exists but this helps
+              # track the insertion index for the created views
+              return [localName, 'directory']
```
nathansobo: I'm a bit lost here and might be confused, but could you just insert the existing directory object itself rather than the placeholder?

lib/directory.coffee
```diff
@@ -163,34 +165,63 @@ class Directory
...
+                  fullPath: fullPath,
+                  symlink: symlink,
+                  expansionState: expansionState,
+                  ignoredPatterns: @ignoredPatterns
+                  })]
+          else if stat.isFile?()
+            if @entries.hasOwnProperty(localName)
+              # push a placeholder since this entry already exists but this helps
+              # track the insertion index for the created views
+              return [localName, 'file']
```
nathansobo: Same here... could you use the object from the `@entries` map instead of a string?

lib/directory-view.coffee
```diff
@@ -106,18 +113,41 @@ class DirectoryView extends HTMLElement
...
       @classList.add('expanded')
       @classList.remove('collapsed')
-      @directory.expand()
+      initial = initial.then => @directory.expand()
 
     if isRecursive
-      for entry in @entries.children when entry instanceof DirectoryView
-        entry.expand(true)
-
+      children = (entry for entry in @entries.children when entry instanceof DirectoryView)
```
nathansobo: I'm not understanding why you're special casing on the existence of `firstEntry`... Is it to check for the completion the view's directory's expansion? If so could you just always wait on the promise, which might already be resolved and do work immediately?

nathansobo:
Okay, I looked at everything. It seems as reasonable as it can be for the fact that this code is woefully in need of a view model layer to separate the async complexity from DOM concerns. Asked a couple questions and suggested a couple things.

lib/directory.coffee
maxbrunsfeld: Since you're building up a corresponding array of promises and need a closure for scoping, it might make sense to use `Array.prototype.map`.

lib/tree-view.coffee
```diff
@@ -53,7 +53,34 @@ class TreeView extends View
...
+    # expansion states. So, we could make the expansion state construct know
+    # about when it's been fully walked, or we could just watch the DOM for the
+    # item in question to appear. I'm still a little concerned about hitting an
+    # edge case where this doesn't get ::disconnected() though.
+    #
+    # 1: Directories were expanded synchronously, but this was changed to work
+    #    with the new async git repo interaction. However, we also use event
+    #    subscription with Directory::onDidAddEntries to build up a tree view as
+    #    well, meaning we can't just tag on to the end of the Directory::expand()
+    #    promise chain :/
```
maxbrunsfeld: Naively, I would think we could have `updateRoots` return a promise that resolves after all of the directories are expanded, and do something like

```coffee
@updateRoots(state.directoryExpansionStates).then =>
  if state.selectedPath
    @selectEntryForPath(state.selectedPath)
  else
    @selectEntry(@roots[0])
```

I assume I'm missing something; how does the `Directory::onDidAddEntries` code path relate to this?

lib/tree-view.coffee
thedaniel: I wish we could figure out a way to rewrite tree-view incrementally without making it a giant project...

lib/directory.coffee
thedaniel: Good call - `do` would be cleaner, but `map` is the right thing to do here, I think, I prefer using natural JS stuff over CS sugar unless it's really messy to do so.

lib/directory.coffee
thedaniel: I'm not sure. I wondered the same thing (it's the same way in the sync code) but I didn't fully grok why the placeholder was used at the time and just kept the existing behavior to avoid more breakage. Now that things are green again it's a better time to mess with it.

thedaniel: Naïvely swapping in `@entries[localName]` lights the specs on fire - I'll spend a small amount of time digging into it before committing to the cargo cult

thedaniel: Interestingly passing the `File` entry down is fine, it just causes problems when the `Directory` is passed down instead of `'directory'`. I wonder if this has to do with async expanding of the tree...

alexandernst:
Almost 2 months without any work/news on this. Is this dead?

thedaniel:
@alexandernst not dead, just got set aside while we did some other work. The async repo support only just landed in stable a week ago so we'll need to pick this back up again soon.

50Wliu:
I don't believe we're going with the route that this PR is taking, considering nodegit is no more in Atom core.

PR #631, opened 2015-10-28
Select files after paste operation

### LINKED ISSUE #573: select new file after copy and paste
- You usually want to do something (rename, delete) with the new file, not the old one. 
- New file would be more easy to find. 
- It is a standard behavior in tree views and file managers.

glangford:
Fix #573

### DIFF
--- a/lib/tree-view.coffee
+++ b/lib/tree-view.coffee
@@ -562,10 +562,12 @@ class TreeView extends View
     cutPaths = if LocalStorage['tree-view:cutPath'] then JSON.parse(LocalStorage['tree-view:cutPath']) else null
     copiedPaths = if LocalStorage['tree-view:copyPath'] then JSON.parse(LocalStorage['tree-view:copyPath']) else null
     initialPaths = copiedPaths or cutPaths
+    newPaths = []
 
-    catchAndShowFileErrors = (operation) ->
+    catchAndShowFileErrors = (operation, newPath) ->
       try
         operation()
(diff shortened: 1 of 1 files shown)


50Wliu:
Hi @glangford, can you please add specs to this?  Thanks!

glangford:
Specs updated.

glangford:
select-after-paste

glangford:
### Additional commentary on Git status

Note that this PR also fixes the latent problem of git status being incorrectly shown on paste. The solution is included as part of `selectOnCreate()` at the moment. However, the proper fix is likely in files.coffee; note that in updateStatus(), the git status is retrieved from the cache. I have verified that pasted files will have no entry in the `GitRepository` cache, so a status of null is returned when they are initially constructed and displayed in the tree view.

Because the file status in cache is null, the parent directory status is also incorrect. (Repro in master by copying and pasting files into a directory, change focus to another app, then change focus back).

So an additional commit would be to change `updateStatus` for class `File` to include
```coffee
status = repo.getCachedPathStatus(@path) ? repo.getPathStatus(@path)
```
rather than just
```coffee
status = repo.getCachedPathStatus(@path)
```

Let me know if this sounds ok.

### References

<https://github.com/atom/tree-view/blob/master/lib/file.coffee#L55>

<https://github.com/atom/atom/blob/7d743827820969871e8724e6f353887a3ddfc105/src/git-repository.coffee#L338>

PR #643, opened 2015-11-07
:bug: Natural sort should be I18n compatible

billybonks:
I initially added a faster implementation of natural sort https://github.com/atom/tree-view/pull/621, but it was not I18n compatible.

the users locale is derived with `(new Intl.Collator()).resolvedOptions().locale` because electron's implementation of `navigator.langauge` https://github.com/atom/electron/pull/363 does not return a BCP 47 language tag.

this implementation is also faster then `natural-sort-lite` and `String.localeCompare`

ping @nathansobo @jeancroy

### DIFF
--- a/lib/directory.coffee
+++ b/lib/directory.coffee
@@ -5,7 +5,6 @@ fs = require 'fs-plus'
 PathWatcher = require 'pathwatcher'
 File = require './file'
 {repoForPath} = require './helpers'
-naturalCompare = require 'natural-compare-lite'
 realpathCache = {}
 
 module.exports =
@@ -162,8 +161,7 @@ class Directory
       names = fs.readdirSync(@path)
     catch error
       names = []
-
-    names.sort(naturalCompare)
+    names.sort(new Intl.Collator(undefined, {numeric: true, sensitivity: "base"}).compare)
 
     files = []
     directories = []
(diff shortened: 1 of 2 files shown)


billybonks:
looking at a sort lines commit passing undefined seems to work https://github.com/atom/sort-lines/pull/8/files#diff-d6704175dc178121f16be36010281bccR46. i will have to test it out

jeancroy:
Sensitivity: base looks user friendly too.

billybonks:
@jeancroy if i understand the Swedish alphabet issue correctly, the user doesn't want `ä` to be treated as `a` it come after `z`. sensitivity base would definitely treat it like an `a`

jeancroy:
@billybonks 
In any french dictionary `é` is basically the same letter as `e`
But I just checked and :
````
"é".localeCompare("e","fr") == 1
"é".localeCompare("f","fr") == -1
````

so between e and f seems like OK too.

billybonks:
so i added sensivity base, as it seems to sort Swedish correctly as well.

jeancroy:
unfortunately I do not know enough Swedish  to make a statement.
maybe @thomasjo can comment on the ordering of such alphabet

billybonks:
haha neither do i, based it on how wikipedia sorts it :smile:

nathansobo:
Thanks for the followup. Learned a lot watching this.

thomasjo:
> unfortunately I do not know enough Swedish to make a statement.
> maybe @thomasjo can comment on the ordering of such alphabet

The `äöå` does indeed come after `z`, same with the Norwegian and Danish variations `æøå`. Good job guys!

PR #649, opened 2015-11-14
Fix copy/pasting files under certain scenarios

### LINKED ISSUE #397: unexpected naming a copied file
_From @kirrie on March 5, 2015 1:12_

in the tree menu on left window, i copied a file with ctrl + c / ctrl + v.

<b>a.js -> a0.js</b> is ok.
but, <b>a.b.js -> a0.b</b> is not what i expected. i expected 'a.b0.js'.

i'm using 0.186.0 version of atom.

_Copied from original issue: atom/atom#5848_

### LINKED ISSUE #464: Having a dot inside the directory path breaks copy/paste of file
Having a dot inside the directory path breaks copy/paste of files.  It seems atom tries to append a number before the first dot it finds.  This should not be done with the full path but just the file name and also the last and not first dot should be used.

    Error: EACCES: permission denied, mkdir '/home/f0.ludwig'
        at Error (native)
        at Object.fs.mkdirSync (fs.js:791:18)
        at sync (/usr/share/atom/resources/app.asar/node_modules/fs-plus/node_modules/mkdirp/index.js:55:12)
        at sync (/usr/share/atom/resources/app.asar/node_modules/fs-plus/node_modules/mkdirp/index.js:61:24)
        at sync (/usr/share/atom/resources/app.asar/node_modules/fs-plus/node_modules/mkdirp/index.js:61:24)
        at sync (/usr/share/atom/resources/app.asar/node_modules/fs-plus/node_modules/mkdirp/index.js:61:24)
        at sync (/usr/share/atom/resources/app.asar/node_modules/fs-plus/node_modules/mkdirp/index.js:61:24)
        at sync (/usr/share/atom/resources/app.asar/node_modules/fs-plus/node_modules/mkdirp/index.js:61:24)
        at sync (/usr/share/atom/resources/app.asar/node_modules/fs-plus/node_modules/mkdirp/index.js:61:24)
        at Function.sync (/usr/share/atom/resources/app.asar/node_modules/fs-plus/node_modules/mkdirp/index.js:61:24)
        at Object.fsPlus.writeFileSync (/usr/share/atom/resources/app.asar/node_modules/fs-plus/lib/fs-plus.js:278:14)
        at TreeView.module.exports.TreeView.pasteEntries (/usr/share/atom/resources/app.asar/node_modules/tree-view/lib/tree-view.js:965:32)
        at space-pen-div.atom.commands.add.tree-view:paste (/usr/share/atom/resources/app.asar/node_modules/tree-view/lib/tree-view.js:277:26)
        at CommandRegistry.module.exports.CommandRegistry.handleCommandEvent (/usr/share/atom/resources/app.asar/src/command-registry.js:238:29)
        at CommandRegistry.handleCommandEvent (/usr/share/atom/resources/app.asar/src/command-registry.js:3:61)
        at CommandRegistry.module.exports.CommandRegistry.dispatch (/usr/share/atom/resources/app.asar/src/command-registry.js:153:19)

### LINKED ISSUE #581: Copy & pasting a file to the same directory will remove the extension

### LINKED ISSUE #648: Copying and pasting a file with no extension adds 'undefined' extension.
Normally when copy-pasting `fileName.txt`, we see this pattern:
```
fileName.txt
fileName0.txt
fileName1.txt
...
```

If the file being copied has no extension, it instead creates files with a new extension:
```
fileName
fileName0.undefined
fileName1.undefined
...
```

50Wliu:
This probably should have 2 separate commits, but oh well.

The two issues that have been fixed:
1) If the absolute filepath you were copying had additional periods in it (eg `/Users/John.Doe/test.txt`), the Tree View would interpret the file to be pasted as `/Users/John0.Doe`.  This would usually fail (for pretty obvious reasons, you're trying to create a new user), especially on Windows.
2) If the file you were copying did not have an extension (eg `test`), the pasted file would be `test0.undefined` because the extension was undefined, and Tree View didn't care.

The fix for 1) involved moving away from `.split()` and switching to a similar regex that captures only the *last* period instead of the first.  This naturally led to 2): by adding a check to make sure there are no file separators after the period, the captured regex will always be undefined if the filename doesn't have an extension.  Tree View now detects this and acts appropriately.

~~This *somewhat* addresses #581 in that the last extension will no longer be stripped as per 1).  HOWEVER, the second-to-last extension and not the true filename will have the file counter appended to it: `style.min.css` -> `style.min0.css`.  On the other hand, #397 (which I just realized #581 is a duplicate of) is asking for the functionality implemented in this PR. :confused:~~
This PR now fixes #581 in the manner requested which I feel is more natural than #397.

Specs might take me a while.  I'll see what I can do.

Fixes #464, fixes #648, fixes #581, closes #397

### DIFF
--- a/lib/tree-view.coffee
+++ b/lib/tree-view.coffee
@@ -588,8 +588,15 @@ class TreeView extends View
             if initialPathIsDirectory
               newPath = "#{originalNewPath}#{fileCounter.toString()}"
             else
-              fileArr = originalNewPath.split('.')
-              newPath = "#{fileArr[0]}#{fileCounter.toString()}.#{fileArr[1]}"
+              if process.platform is 'win32'
+                fileArr = /(.+)\.(?!.*\\)(.+)/.exec(originalNewPath)
+              else
+                fileArr = /(.+)\.(?!.*\/)(.+)/.exec(originalNewPath)
+
+              if fileArr?
+                newPath = "#{fileArr[1]}#{fileCounter.toString()}.#{fileArr[2]}"
+              else # File didn't have an extension so don't attempt to append a nonexistent one
+                newPath = "#{originalNewPath}#{fileCounter.toString()}"
             fileCounter += 1
 
           if fs.isDirectorySync(initialPath)
--- a/lib/tree-view.coffee
+++ b/lib/tree-view.coffee
@@ -611,7 +611,7 @@ class TreeView extends View
               newPath = "#{originalNewPath}#{fileCounter.toString()}"
             else
               extension = getFullExtension(originalNewPath)
-              filePath = path.join(path.dirname(originalNewPath), path.basename(originalNewPath, fullExtension))
+              filePath = path.join(path.dirname(originalNewPath), path.basename(originalNewPath, extension))
               newPath = "#{filePath}#{fileCounter.toString()}#{extension}"
             fileCounter += 1
 
--- a/lib/tree-view.coffee
+++ b/lib/tree-view.coffee
@@ -3,7 +3,7 @@ path = require 'path'
 
 _ = require 'underscore-plus'
 {BufferedProcess, CompositeDisposable} = require 'atom'
-{repoForPath, getStyleObject} = require "./helpers"
+{repoForPath, getStyleObject, getFullExtension} = require "./helpers"
 {$, View} = require 'atom-space-pen-views'
 fs = require 'fs-plus'
 
@@ -610,15 +610,9 @@ class TreeView extends View
             if initialPathIsDirectory
               newPath = "#{originalNewPath}#{fileCounter.toString()}"
             else
-              nextExtension = path.extname(originalNewPath)
-              fullExtension = ''
-              while nextExtension isnt '' # This is for files with multiple extensions since extname only returns the last extension
-                fullExtension = nextExtension.concat(fullExtension)
-                nextExtension = path.extname(path.basename(originalNewPath, fullExtension))
-
-              filePath = path.dirname(originalNewPath) + path.sep + path.basename(originalNewPath, fullExtension)
-
-              newPath = "#{filePath}#{fileCounter.toString()}#{fullExtension}"
+              extension = getFullExtension(originalNewPath)
+              filePath = path.dirname(originalNewPath) + path.sep + path.basename(originalNewPath, extension)
+              newPath = "#{filePath}#{fileCounter.toString()}#{extension}"
             fileCounter += 1
 
           if fs.isDirectorySync(initialPath)
--- a/lib/tree-view.coffee
+++ b/lib/tree-view.coffee
@@ -625,9 +625,9 @@ class TreeView extends View
               newPath = "#{originalNewPath}#{fileCounter.toString()}"
             else
               if process.platform is 'win32'
-                fileArr = /(.+)\.(?!.*\\)(.+)/.exec(originalNewPath)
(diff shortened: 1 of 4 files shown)


lib/tree-view.coffee
```diff
@@ -588,8 +588,15 @@ class TreeView extends View
             if initialPathIsDirectory
               newPath = "#{originalNewPath}#{fileCounter.toString()}"
             else
-              fileArr = originalNewPath.split('.')
-              newPath = "#{fileArr[0]}#{fileCounter.toString()}.#{fileArr[1]}"
+              if process.platform is 'win32'
+                fileArr = /(.+)\.(?!.*\\)(.+)/.exec(originalNewPath)
```
50Wliu: I initially tried to use an extended regex pattern here (the `///` provided by Coffeescript), but that led to some nasty escaping issues with `#{path.sep}` because apparently interpolation happens before the regex is parsed.

prrrnd:
+1, came here to report the issue when copying/pasting files with dots in their absolute path.

50Wliu:
Hmm, that reminds me that I still need to add specs to this.

hedefalk:
+1.

This also happens for me when trying to copy the file 

~/.teamocil/sj.yml

Instead of getting

~/.teamocil/sj.0.yml

I get another folder and a file without extension:

~/0.teamocil/sj

This on 1.3.3. Pretty please, can this be fixed into master?

50Wliu:
> Pretty please, can this be fixed into master?

I still need to fix the specs.

50Wliu:
The remaining specs are severely stumping me, but I'm probably just missing something really obvious.

When run normally through the specs window, `$(treeView.roots[0].entries).find('.directory:contains(test\\.dir)')` fails to find anything.  However, if I set a breakpoint and run the exact same code in the console, a match is found and the test passes (even though I didn't assign it to anything)?  :confused:

Anyone have any ideas?

50Wliu:
> Specs might take me a while.

A while being 6 months :laughing:...:frowning:.  The new specs now pass on my machine - just waiting on Travis to confirm that they're good to go :watch: :hourglass_flowing_sand:.

I also decided to go with #581 and not #397 as the base filename should always be incremented and not any of the extensions.

I'm worried about the regex performance though - matching long filenames takes a considerable number of steps (147 for `C:\Users\John.Doe\Documents\Github\atom\tree-view\lib\test.min.js`), and rejecting filenames without extensions takes more than 10x as many steps.  I'll try finding a cleaner way, but for now, this is ready for review! :tada:

sftd:
I encountered 2 "wrong" behaviors while copying (Atom 1.7.3, Windows 10 ). Are they covered in here?

1. Copying `project-path/.dir-with-dot/file-name.ext` into `project-path/.dir-with-dot` results in `project-path/0.dir-with-dot/file-name` (pretty sure this one is).

2.
atom-bug
(not sure about this one).

I would appreciate answer.

as-cii:
Thanks, @50Wliu!

Ajedi32:
Fixes #674

Issue #407, opened 2015-03-22
Use Ctrl+f in tree view context for call "Search in Directory" if directory selected


xlaywan:
Now it is just "Find in Buffer"

benogle:
This is a really good idea.

chick-chains:
Shouldn't this be handled by the find and replace package?

jsumnersmith:
I've been digging into this, and there's some baked in functionality that could handle this (namely, the command: `project-find:show-in-current-directory`), but it needs some help figuring out which menu item is selected (usually, it has the mouse click event to help with the selected item).

bronson:
I love this idea.  Changing the File/Directory Pattern in Find in Project is tedious and leaves you in an unfortunate mode.  If you search again, and forget to delete the pattern, you're going to be confused by the lack of results...  at least, that's happened to me a bunch.

peterp:
@benogle I've added a pull-request that adds the required functionality. I'm just not entire sure how I should go about testing that the search dialog is visible?

judygabeskiria:
what if user is not selecting a directory and wants to look for in all directories?

elliotaplant:
@peterp Which pull request was that? Did it get merged? If not, I'd like to tackle this one.

peterp:
@elliotaplant it wasn't merged, they decided to go a different route, but here' s the PR: https://github.com/atom/tree-view/pull/679

elliotaplant:
Thanks @peterp! It looks like they added context menu options, but no hotkey. I guess this would be a pretty simple addition to my keybindings file. @benogle, should we close this issue since you can right-click a directory in the tree view to search in that directory, or is it still worth making a hotkey that command?

PR #679, opened 2015-12-22
add contextual search for selected item

peterp:
for issue #407, still trying to figure out how to test this, any tips?

I originally thought I could target a detailed selector with a keymap: `.tree-view li.directory,` but I couldn't get that to work. I need to dive a bit deeper into `atom-keymap` to figure out why.

### DIFF
--- a/spec/tree-view-spec.coffee
+++ b/spec/tree-view-spec.coffee
@@ -3126,21 +3126,21 @@ describe "TreeView", ->
       fs.writeFileSync(filePath, "doesn't matter")
 
       atom.project.setPaths([rootDirPath])
-      spyOn(atom.commands, 'dispatch').andCallThrough();
+      spyOn(atom.commands, 'dispatch').andCallThrough()
 
     describe 'when a file is selected', ->
       it 'dispatches the command to show the `find and replace` dialog', ->
         fileView = findFileContainingText(treeView.roots[0], 'a-file.txt')
         fileView.click()
         atom.commands.dispatch(treeView.element, 'tree-view:search-selected-entry')
-        expect(atom.commands.dispatch.calls[1].args[1]).toEqual('find-and-replace:show');
+        expect(atom.commands.dispatch.calls[1].args[1]).toEqual('find-and-replace:show')
 
     describe 'when a directory is selected', ->
       it 'dispatches the command to show the `find in project` dialog', ->
         dirView = findDirectoryContainingText(treeView.roots[0], 'a-directory')
         dirView.click()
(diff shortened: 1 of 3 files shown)


benogle:
Can you write specs for this PR?

peterp:
@benogle I can, I just need some direction. My initial was to test for the visibility of a ".search/ .search-project" class, but further inspection into the specs made me realise that this isn't loaded into tree-view's specs. (And maybe shouldn't be?)

Is there a precedent for writing specs that dispatch atom commands?

lee-dohm:
I would start with checking for the visibility like you were first planning to do and see if you can make that robust. If that doesn't work or doesn't work reliably, let us know!

peterp:
@lee-dohm I've added the tests that check for visibility, but the .find-and-replace and .find-project elements are null, do I have to activate the package?

lee-dohm:
Yes, you have to manually activate any packages you want to work in specs.

peterp:
@lee-dohm I've now added the `find-and-replace` package, but the tests still don't find the class names.

spec/tree-view-spec.coffee
```diff
@@ -2530,6 +2534,43 @@ describe "TreeView", ->
               expect(treeView.list).toHaveClass('full-menu')
               expect(treeView.list).not.toHaveClass('multi-select')
 
+  fdescribe "search based on selected items", ->
```
liuderchi: Typo: `fdescribe`

tenorok:
Hello, is there any chance for merging?

peterp:
@lee-dohm this is ready for review again, an unrelated test is failing, but passing locally.

lee-dohm:
Thanks @peterp! I'm swamped today and tomorrow, but I'll take a look Wednesday :+1:

lee-dohm:
@peterp Sorry I haven't gotten back to you on this yet. It is still on my radar. I'm working on an improvement for it as well. I'll touch base on it early next week.

lee-dohm:
@peterp In digging deeper on this, I found that the [find-and-replace package has context menu items that it adds to the tree-view](https://github.com/atom/find-and-replace/blob/master/menus/find-and-replace.cson#L24-L27). A similar approach could be used to add the keybindings you're wanting here and I bet the change would be much simpler.

What do you think?

peterp:
@lee-dohm Sorry, I've had a bit of a vacation, that sounds like a much better approach, and we can keep the functionality where it belongs. I'll close this off and get back to you.

lee-dohm:
Sounds good. Let me know if you need any help with it.

PR #682, opened 2015-12-24
Open file in pending state on single click

kuychaco:
Single-clicking opens a file in pending state and double-clicking opens a file regularly (in a non-pending state).

See corresponding changes in [atom/atom/pull/10178](https://github.com/atom/atom/pull/10178)
Addresses atom/atom/issues/9165

### DIFF
--- a/lib/tree-view.coffee
+++ b/lib/tree-view.coffee
@@ -211,6 +211,7 @@ class TreeView extends View
       when 2
         if entry instanceof FileView
           atom.workspace.getActivePaneItem().terminatePendingState?()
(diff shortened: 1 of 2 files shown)


kuychaco:
TODO: Fix failing test (see image).

Not sure why the spec shows the Text Editor having focus rather than the Tree View. When I'm manually testing by double-clicking a directory the Tree View retains focus. 

Also, getting uncaught errors from Pane::activate. Will trouble shoot more...

screen shot 2015-12-24 at 2 30 10 pm

PR #695, opened 2016-01-12
Make 'Move in trash' more verbose on failure, add a note for Linux

Hyrrmadr:
Add an alert to warn the user when a file couldn't be moved to trash correctly instead of silently failing.

Add a note for Linux to ask the user to check if 'gvfs-trash' is installed (the command-line utility used by electron to move files to trash).

screenshot_2016-01-12_11-31-43

### DIFF
--- a/lib/tree-view.coffee
+++ b/lib/tree-view.coffee
@@ -509,7 +509,7 @@ class TreeView extends View
     else if activePath = @getActivePath()
       selectedPaths = [activePath]
 
-    return unless selectedPaths
+    return unless selectedPaths and selectedPaths.length > 0
 
     for root in @roots
       if root.getPath() in selectedPaths
@@ -524,7 +524,10 @@ class TreeView extends View
       buttons:
         "Move to Trash": ->
           for selectedPath in selectedPaths
-            shell.moveItemToTrash(selectedPath)
+            if not shell.moveItemToTrash(selectedPath)
+              atom.notifications.addError "The file couldn't be removed#{if process.platform is 'linux' then " (is `gvfs-trash` installed?)" else ""}",
+                detail: "Error while moving '#{selectedPath}' to trash"
+                dismissable: true
             if repo = repoForPath(selectedPath)
               repo.getPathStatus(selectedPath)
         "Cancel": null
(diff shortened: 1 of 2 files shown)


lee-dohm:
I think this would be better implemented as a [notification](https://github.com/atom/notifications). Also, could you add a spec for this behavior?

Hyrrmadr:
Thanks for your feedback! I followed your advice and implemented it as a notification.

screen shot 2016-01-14 at 16 05 28

Sorry for the lack of spec, I couldn't find a way to trigger the bug without `gvfs-trash` missing under Linux, but I figured a way now.

I also fixed a small 'bug' where atom would display the "Move to trash" confirm popup when trying to delete with no file selected (I deselected the current file with ⌘). I added a spec to test the behavior.

lee-dohm:
One question on this ... what if I have 15 files selected and they all run into an error? Am I going to get 15 notifications?

Hyrrmadr:
With these changes, yes.
Do you think I should make a single notification with the list of the failed deletions?
But it could be cumbersome too.

lee-dohm:
Yes, it would be cumbersome. But I tend to see it as one failure rather than fifteen. Also, if it was 1015 instead of 15, I wouldn't want my editor thrown into some crazy notification loop or something.

I'd appreciate it if you could make the change. Thanks!

Hyrrmadr:
I mean obviously you are right, one notification is better than 10 000. I was just wandering if there was any better solution. Maybe a way to collapse it like the exception notification.

Here is how it looks now:
screen shot 2016-01-14 at 22 59 16

50Wliu:
Should `gvfs-trash` just be made a dependency to install Atom so that we don't need the notification to install it?

Hyrrmadr:
We could just remove the note in the notification about `gvfs-trash`, I just found it pretty frustrating that I couldn't remove anything in Atom because I was missing this package.
Anyway, I think it's a good idea to warn the user about a failure instead of just doing nothing, obviously it would be better if we had any idea why the deletion failed.

lee-dohm:
I agree that this change is good to have rather than failing silently. We can have a separate issue or PR for the gvfs dependency.

Hyrrmadr:
So, do you want me to remove the note about `gvfs-trash` ?

lee-dohm:
@Hyrrmadr Thanks for the ping. No, I think the note about gvfs-trash is fine for now.

50Wliu:
@lee-dohm are changes still required for this PR?

lee-dohm:
Thanks for this! And thanks for the ping @50Wliu :grinning:

Issue #128, opened 2014-05-01
Add package level ignoredNames


kevinsawicki:
This would be similar to `fuzzy-finder.ignoredNames`, a separate list of ignores that only pertain to the tree view

adambuczynski:
+1 for this feature please, although the last activity has been a year and a half ago.

My use case is the following: I don't need to see `.DS_store` files in my tree view, but I *would* like to see the `node_modules` folder so that I can easily browse it if I need to. However, when doing a search in files or using the fuzzy finder, I want Atom to *ignore* the `node_modules` folder.

So to accomplish the latter, I currently have to specify the `node_modules` folder in my list of ignored files. Tree view however, uses that same list to determine which files to hide, therefore I cannot accomplish the former.

If tree view would have a separate list to dictate which files to hide, that would solve the problem. To maintain some kind of backwards compatibility, it could have a default fallback to the ignored files list.

adambuczynski:
I've created a PR for this, but haven't been able to test it locally as Atom seems to be having issues with my dev package folder being a symlink to my fork.

I think I've ported all of the logic from the fuzzy-finder's implementation to tree-view though, including test. If someone could please have a look and let me know if anything needs changing, that's be great.

Also, this is my first time writing code in coffeescript (-.-) so bear with me.

adambuczynski:
I've added the merge toggle, but I think this is an overarching issue which should perhaps be solved in a more appropriate way. 

For example, a more elegant way would be to allow support for negated globs, e.g. `!node_modules`. Then, merging could be made a bit smarter to filter out these globs after the merge. Currently, if you specify `!node_modules` in the ignoredNames list, only the `node_modules` folder will be shown.

PR #698, opened 2016-01-13
Add package level ignoredNames

adambuczynski:
See #128
This PR adds a package level ignoredNames configuration setting which gets merged in with the core ignoredNames setting, akin to what’s done in the fuzzy finder package.


adambuczynski:
Updated to put config settings in `package.json`

Issue #459, opened 2015-05-31
Don't focus the tree view with "reveal-active-file


mehcode:

vraravam:
Sorry - I dont have enough knowledge of coffeescript/atom API to get this working. Any help would be appreciated. I used a "hack" in my init script, but that has stopped working in the past week or so.

```
atom.workspace.observeTextEditors (editor) ->
  editor.onDidSave ->
    atom.commands.dispatch(atom.workspaceView.element, 'tree-view:reveal-active-file')
    atom.views.getView(editor).focus()
  editor.onDidChangePath ->
    atom.commands.dispatch(atom.workspaceView.element, 'tree-view:reveal-active-file')
    atom.views.getView(editor).focus()
```

Issue #188, opened 2014-07-15
Reveal in tree view does not highlight file if project folder is collapsed


JLoppert:
### Repro steps:
1. Open the `image-view` repository in safe mode `atom --safe`
1. Open the `spec/fixtures` folder
1. Open `binary-file.png`
1. Collapse the `spec` folder
1. Collapse the root folder (`image-view`)
1. Right click on the image and select `Reveal in Tree View` from the context menu

**Expected:**
`binary-file.png` to be highlighted because it was revealed

**Actual:**
The `spec` folder is expanded but the file is not selected

![reveal root](https://cloud.githubusercontent.com/assets/1058982/19150598/f7c24948-8bc7-11e6-838e-fba97de42127.gif)

----

Repro Steps:
1. Open file inside a folder that visible in the tree view
2. Collapse folder with file
3. In file, right click and select Reveal in Tree View

The desired behavior is that the file is highlighted and not the parent folder. The desired behavior happens if the folder is expanded. 

![croppercapture 1](https://cloud.githubusercontent.com/assets/1620641/3587219/1f4d0fba-0c37-11e4-9416-89d406ae7046.Gif)

Windows 7, Atom Version 0.115.0

50Wliu:
Also apparently happens under Mac.

benogle:
I am unable to reproduce on latest master

Ben3eeE:
I can still reproduce this if the project folder is collapsed. With all other folders it seems that this bug has been fixed.

I am on Windows 7, Atom 1.10.2

![reveal root](https://cloud.githubusercontent.com/assets/1058982/19150598/f7c24948-8bc7-11e6-838e-fba97de42127.gif)

PR #715, opened 2016-01-30
feat: add two new reveal options for the tree-view

kumarharsh:
1. AutoReveal - control the revealing of files automatically as they
    are switched in the panes.
2. FocusOnReveal - control whether the focus on the editor panes is
    transferred to the tree-view pane on calling 'reveal-active-file'

This adds a small behaviour which is more in line with what
SublimeText offers. This PR also addresses other open issues/PRs
like: #247 #459 #188

### DIFF
--- a/lib/main.coffee
+++ b/lib/main.coffee
@@ -27,6 +27,14 @@ module.exports =
       type: 'boolean'
       default: true
       description: 'When listing directory items, list subdirectories before listing files.'
+    focusOnReveal:
+      type: 'boolean'
+      default: true
+      description: 'When a file is revealed in the sidebar, switch the focus to the sidebar'
+    autoReveal:
+      type: 'boolean'
+      default: false
+      description: 'Reveal files as they are opened or switched-to in the editor panes'
 
   treeView: null
 
--- a/lib/tree-view.coffee
+++ b/lib/tree-view.coffee
@@ -137,6 +137,9 @@ class TreeView extends View
 
     @disposables.add atom.workspace.onDidChangeActivePaneItem =>
       @selectActiveFile()
+      if atom.config.get('tree-view.autoReveal')
+        @revealActiveFile()
+
     @disposables.add atom.project.onDidChangePaths =>
       @updateRoots()
     @disposables.add atom.config.onDidChange 'tree-view.hideVcsIgnoredFiles', =>
@@ -293,7 +296,8 @@ class TreeView extends View
     return if _.isEmpty(atom.project.getPaths())
 
     @attach()
-    @focus()
+    if atom.config.get('tree-view.focusOnReveal')
+      @focus()
 
     return unless activeFilePath = @getActivePath()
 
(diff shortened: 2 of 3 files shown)


kumarharsh:
@kuychaco can someone review this PR?

benogle:
Can you split these out into 2 PRs, one for each option? The focus option is less controversial, and looks like we'll be able to merge that.

kumarharsh:
Meaning the `autoreveal` is controversial? Even if it's configurable

benogle:
I think we need to think about that one a little more.

kumarharsh:
What are the concerns if any, if I may know.

benogle:
It's yet another option. Do we really need it? We'd really like to limit the number of options if we can. Maybe it can be the default behavior? Does it have to do with the focus option? Like is the behavior different when the focus option is checked? Basically someone needs to look at it a little more deeply. So by splitting the PR, it allows us the merge the one that seems like a no brainer, and think about the other.

kumarharsh:
Yeah sure, I'd do that tomorrow. Although both options play well with each other and everything else. I thought to keep the `autoreveal` as an option because some people might like the default functionality (since it's been given with atom for so long)

laughedelic:
@benogle 

> It's yet another option. Do we really need it? 

There are at least 3 packages that I know with this functionality:

- https://atom.io/packages/auto-reveal-active-file

  ```
auto-reveal-active-file
├── 0.3.1
├── https://github.com/sfaedo/auto-reveal-active-file
├── Auto reveal the current active file in the tree view
├── 289 downloads
└── 2 stars
  ```
- https://atom.io/packages/auto-reveal-in-sidebar

  ```
auto-reveal-in-sidebar
├── 0.8.0
├── https://github.com/DeanNonatomic/atom-auto-reveal-in-sidebar
├── Calls 'tree-view:reveal-in-sidebar' whenever a file is opened or brought into focus
├── 1264 downloads
└── 3 stars
  ```

- https://atom.io/packages/synced-sidebar

  ```
synced-sidebar
├── 0.4.3
├── https://github.com/peterdotjs/atom-synced-sidebar
├── Based on Sublime Text SyncedSideBar plugin. Active tab & sidebar are synced - sidebar entry automatically scrolls into view.
├── 6017 downloads
└── 54 stars
  ```

Looking at the downloads I would say that people use them. But none of them is perfect, so having it as a native option which plays well with all the others would be much better.

@kumarharsh :+1:

kumarharsh:
@laughedelic thanks. Didn't know these many were already there... Actually didn't even search for it :stuck_out_tongue_closed_eyes: 

This is a small init script I wrote before deciding to contribute here:

```coffee-script
atom.workspace.observeActivePane (pane) ->
  pane.onDidChangeActiveItem ->
    workspace = atom.views.getView(atom.workspace)
    # Reveal the current file in sidebar when switched to that tab
    atom.commands.dispatch(workspace, 'tree-view:reveal-active-file')
```

lee-dohm:
Since it seems this was split into two separate PRs, should this one be closed @kumarharsh?

kumarharsh:
yes! forgot to do that :smile:

PR #739, opened 2016-02-16
Feature Proposal: Advanced file add mode

mrapogee:
I constantly mess-up the `A` and `a`, creating folders when I should be creating files and vice versa. Am I alone? It's also hard to create multiple files with the current add file/folder functionality.

This provides a single mode, currently bound to n, to create both files and folders. Entering the following example into the dialog:

`start.js, src/server.js, src/client/, src/user/ , ./.babelrc`

The following will be created:
- start.js file in the selected directory
- src/server.js file in the selected dir
- src/client/ folder in the selected dir
- src/user/ folder in selected dir
- .babelrc file in the root of the project

It needs testing and finalization, but I figured I'd gather some feedback before finishing. Currently using and has made my life much easier.

It also leaves the `A` and `a` bindings the same.


lee-dohm:
This appears to duplicate some or all of the functionality of the [advanced-open-file package](https://atom.io/packages/advanced-open-file). Perhaps you might want to submit some of this as a pull request there?

mrapogee:
I would, but with those you can't select a folder in the sidebar and then easily list `a.js, b.js, c.js`.

Could you expose to other plugins what the currently selected file in the sidebar is?

lee-dohm:
> Could you expose to other plugins what the currently selected file in the sidebar is?

I think this would be a great idea for a service that the tree-view could provide. I'd be much more likely to merge a PR for something like that. You may want to look at the documentation on services:

https://atom.io/docs/latest/behind-atom-interacting-with-other-packages-via-services

50Wliu:
tree-view now provides [an API] which allows you to query the currently selected files' paths using `selectedPaths()`.  Therefore, it seems like this PR can now be implemented in community packages, so I'm closing this out.

PR #741, opened 2016-02-17
On second click, open without `pending` option

kuychaco:
Merge once Atom v1.7.0 is released

With https://github.com/atom/atom/pull/10872 it’s no longer necessary to 
explicitly `terminatePendingState` here


BinaryMuse:
Now depends on atom/atom#10959

kuychaco:
This is superseded by https://github.com/atom/tree-view/pull/801

PR #743, opened 2016-02-18
Don't close dialog unless Atom has focus

50Wliu:
See atom/atom#1918 for context.  Now switching windows or activating the dev tools will not cause the dialog (new file/folder, duplicate, rename, etc.) to disappear.


lee-dohm:
Thanks for the ping and the contribution!

PR #744, opened 2016-02-21
Add option for disabling VCS colors

zenwalker:
2016-02-21 01-50-01 settings users maxim projects simply simply-backend

### DIFF
--- a/lib/directory.coffee
+++ b/lib/directory.coffee
@@ -12,7 +12,6 @@ class Directory
   constructor: ({@name, fullPath, @symlink, @expansionState, @isRoot, @ignoredPatterns}) ->
     @destroyed = false
     @emitter = new Emitter()
-    @subscriptions = new CompositeDisposable()
 
     if atom.config.get('tree-view.squashDirectoryNames')
       fullPath = @squashDirectoryNames(fullPath)
@@ -32,14 +31,13 @@ class Directory
 
     @submodule = repoForPath(@path)?.isSubmodule(@path)
 
-    @subscribeToRepo()
-    @updateStatus()
+    @loadStatus()
     @loadRealPath()
 
   destroy: ->
     @destroyed = true
     @unwatch()
-    @subscriptions.dispose()
+    @unsubscribeFromRepo()
     @emitter.emit('did-destroy')
 
   onDidDestroy: (callback) ->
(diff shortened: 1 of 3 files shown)


lee-dohm:
Rather than all this code, that tests would need to be written for, I feel it would be much less risky to implement something like this:

First, add this to the stylesheet in the package:

```less
// remove Git colors from tree-view
.tree-view.vcs-disabled {
  .status-modified,
  .status-added,
  .status-renamed,
  .status-removed {
    color: inherit !important; // remove diff color

    &.directory > .list-item {
      color: inherit !important; // remove diff color
    }
    &.directory.selected > .list-item {
      color: @text-color-selected !important; // add selected color
    }
  }
}
```

Second, add code that monitors the configuration and either adds or removes the `.vcs-disabled` class to the top tree view `ol` element:

screen shot 2016-02-21 at 12 56 40 pm

I tested this manually and it seems to work pretty well.

zenwalker:
Your solution is more safe but dirty (`!important` is anti-pattern) and incompatible with some of atom themes. For sample: native-ui theme [use pseudo elements][1] for displaying status.

Currently I just remove subscription from git repository to disabling `updateStatus` calls.
As alternative solution I may suggest don't add status class [here][2] if option is disabled. What do you think?

[1]: https://github.com/fv0/native-ui/blob/1f065db338475372b62185b58b51583c642f72dd/styles/tree-view.less#L7
[2]: https://github.com/atom/tree-view/blob/3b99ca0f0b13879a3d604433caa2d252b6060146/lib%2Fdirectory-view.coffee#L59

lee-dohm:
By not adding the status classes, you're mixing function with presentation. I feel that the functionality should remain exactly the same and only the presentation should change. This is what leads me to a CSS-only (or very nearly so) solution.

zenwalker:
This does't contradict for separating functionality and representation.
[See implementation][1] this option in the tabs package. It is the same that I suggest.

[1]: https://github.com/atom/tabs/blob/4a95521a9284f6b4963815a1c384c5dbaa441fd2/lib/tab-view.coffee#L258

lee-dohm:
I can have some other people from the team take a look once this has automated tests to help us ensure that it doesn't get broken in a future update.

garygreen:
@lee-dohm imo your suggestion doesn't really turn off VCS colours, the classes are still being added which unnecessarily impacts performance.

50Wliu:
Closing until tests are added.

Issue #566, opened 2015-09-08
Poor UX with Drag and Drop Files Folders


eccegordo:
I think this is the right place to report this issue. If not please let me know.

As of Atom 1.0.7 Drag and Drop folders and files and were added as a feature.

While the drag drop is a nice feature, there is a subtle jankiness to current implementation. In my day to day experience, I find it far too easy to drag and drop some folder into another folder and not realize I have done so. Haplessly moving stuff around in my project.

I think part of the problem is the lack of visual indication. OS X solves this problem I believe with spring loaded folders and usually a little green plus circle to indicate you are intending to copy something. 

I am surprised at how hair trigger the drag and drop is. Probably needs some careful review of the event handling and behaviors.

At least two things needed:

1.) More obvious visual indication that a move or copy is about to happen.

2.) Some way to cancel mid flight. I believe this is often handled by dragging away and release of mouse button.

Also, how do you disable this feature currently?

Thanks!

saschagehlich:
> 1.) More obvious visual indication that a move or copy is about to happen.

We can indeed change the icon that is displayed while dragging. I already tried to implement copying files - but for this you'd need to get a keydown event (CMD on Mac, I believe CTRL in Windows) while dragging. And the current version of Chrome doesn't emit any keyboard events while dragging

> 2.) Some way to cancel mid flight. I believe this is often handled by dragging away and release of mouse button.

This works on my end. Pressing ESC also works for me.

danhigham:
+1 for disabling drag and drop in the tree, I have moved folders accidentally too many times to count!

billybonks:
currently the drag and drop feature, holds the icon at the top left. changing it to hold the file from the bottom right could be a good improvement.

it's also the same colour as the rest of the folders so gets lost if you aren't paying attention. we could change the colour to make it more obvious.

i don't mind adding a config to disable it or require a key down to initiate the event. as long as its not the only way, i like it as it is now.

sergiu-paraschiv:
+1 for the disable option

Zireael07:
What do you mean it's the same color as the rest of the folders?

billybonks:
![screen shot 2015-10-19 at 7 35 40 pm](https://cloud.githubusercontent.com/assets/1640136/10576805/a90fdbc0-7698-11e5-834f-d9e1a3bc6adf.png)
@Zireael07. i am currently dragging a file in the screenshot. :)

stgogm:
[...]

PR #748, opened 2016-02-24
Issue #566 - Using configuration parameter to allowDragAndDrop of files and folders

KenBridgeman:
This is my very first Javascript as well as OpenSource Pull request.  I think I've done OK, but if not please be kind.  

**package.json:** Added allowDragAndDrop attribute with a default value of **true** which ensures backward compatibility.

**directory-view.coffee** and **file-view.coffee**:Simple one-liner to get config item and set draggable attribute in DOM structure.

**tree-view.coffee**: One-liner to ensure that changes to the config attribute are propagated to the tree-view immediately.  Without this line, the user would have to exit the app and restart.

**tree-view-spec.coffee**: Had the hardest time figuring this out.  I seemed to have both closure issues as well as inter-process issues with changing of the config item; but really not sure.  In the end, I've got both allowDragAndDrop = true as well as false tests.

### DIFF
--- a/lib/tree-view.coffee
+++ b/lib/tree-view.coffee
@@ -96,11 +96,22 @@ class TreeView extends View
     @on 'mousedown', '.entry', (e) =>
       @onMouseDown(e)
     @on 'mousedown', '.tree-view-resize-handle', (e) => @resizeStarted(e)
-    @on 'dragstart', '.entry', (e) => @onDragStart(e)
-    @on 'dragenter', '.entry.directory > .header', (e) => @onDragEnter(e)
-    @on 'dragleave', '.entry.directory > .header', (e) => @onDragLeave(e)
-    @on 'dragover', '.entry', (e) => @onDragOver(e)
-    @on 'drop', '.entry', (e) => @onDrop(e)
+    @on 'dragstart', '.entry', (e) =>
+      if atom.config.get('tree-view.allowDragAndDrop')
+        @onDragStart(e)
+
+    @on 'dragenter', '.entry.directory > .header', (e) =>
+      if atom.config.get('tree-view.allowDragAndDrop')
+        @onDragEnter(e)
+    @on 'dragleave', '.entry.directory > .header', (e) =>
+      if atom.config.get('tree-view.allowDragAndDrop')
+        @onDragLeave(e)
+    @on 'dragover', '.entry', (e) =>
+      if atom.config.get('tree-view.allowDragAndDrop')
+        @onDragOver(e)
+    @on 'drop', '.entry', (e) =>
+      if atom.config.get('tree-view.allowDragAndDrop')
+        @onDrop(e)
 
     atom.commands.add @element,
      'core:move-up': @moveUp.bind(this)
--- a/lib/tree-view.coffee
+++ b/lib/tree-view.coffee
@@ -97,20 +97,14 @@ class TreeView extends View
       @onMouseDown(e)
     @on 'mousedown', '.tree-view-resize-handle', (e) => @resizeStarted(e)
     @on 'dragstart', '.entry', (e) =>
-      if atom.config.get('tree-view.allowDragAndDrop')
         @onDragStart(e)
-
     @on 'dragenter', '.entry.directory > .header', (e) =>
-      if atom.config.get('tree-view.allowDragAndDrop')
         @onDragEnter(e)
     @on 'dragleave', '.entry.directory > .header', (e) =>
-      if atom.config.get('tree-view.allowDragAndDrop')
         @onDragLeave(e)
     @on 'dragover', '.entry', (e) =>
-      if atom.config.get('tree-view.allowDragAndDrop')
         @onDragOver(e)
     @on 'drop', '.entry', (e) =>
-      if atom.config.get('tree-view.allowDragAndDrop')
         @onDrop(e)
 
     atom.commands.add @element,
@@ -153,6 +147,8 @@ class TreeView extends View
     @disposables.add atom.config.onDidChange 'tree-view.hideVcsIgnoredFiles', =>
       @updateRoots()
     @disposables.add atom.config.onDidChange 'tree-view.hideIgnoredNames', =>
+      @updateRoots()
+    @disposables.add atom.config.onDidChange 'tree-view.allowDragAndDrop', =>
       @updateRoots()
     @disposables.add atom.config.onDidChange 'core.ignoredNames', =>
       @updateRoots() if atom.config.get('tree-view.hideIgnoredNames')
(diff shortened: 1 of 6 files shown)


50Wliu:
So this was discussed before in #623 and ultimately closed as a wontfix, the consensus being that the better solution would to improve the UX of dragging-and-dropping.

Sorry for not catching this earlier and warning you in #566.

/cc @atom/feedback

KenBridgeman:
That's OK, I learned a ton.  

Would you mind very much doing a very quick review and providing a comment as this was my first foray.

spec/tree-view-spec.coffee
```diff
@@ -663,7 +674,10 @@ describe "TreeView", ->
 
         children = root1.find('.directory')
         expect(children.length).toBeGreaterThan 0
-        children.each (index, child) -> expect(child).toHaveClass 'expanded'
+        children.each (index, child) ->
+          expect(child).toHaveClass 'expanded'
+          expect(child.hasAttribute('draggable')).toBe(true)
+          expect(child.getAttribute('draggable')).toBe('true')
```
50Wliu: Why do we have two assertions of this?

maxbrunsfeld:
> Would you mind very much doing a very quick review and providing a comment as this was my first foray.

Of course!

spec/tree-view-spec.coffee
```diff
@@ -2462,6 +2476,44 @@ describe "TreeView", ->
...
+      atom.project.setPaths([rootDirPath])
+
+      dirView = $(treeView.roots[0].entries).find('.directory:contains(test-dir)')
+      dirView[0].expand()
+      fileView1 = treeView.find('.file:contains(test-file1.txt)')
+      fileView2 = treeView.find('.file:contains(test-file2.txt)')
+      fileView3 = treeView.find('.file:contains(test-file3.txt)')
+      treeView.updateRoots()
+
+      it 'makes expanded directory and files draggable', ->
```
50Wliu: Shouldn't setting `allowDragAndDrop` to false *disable* drag-and-drop?

spec/tree-view-spec.coffee
KenBridgeman: I could be wrong about this, but I interpret hasAttribute as being true regardless of whether draggable's value is true or false; it has the attribute and it's value is 'x'.  

I noticed that at the root in DirectoryView that @draggable is not created and therefore, for root I was thinking that it will not have the attribute at all.  

I didn't want the test to error out if the attribute doesn't exist.  

Of course, I'm brand new and don't know what I'm doing. ;)

lib/tree-view.coffee
```diff
@@ -96,11 +96,16 @@ class TreeView extends View
...
+    @on 'dragstart', '.entry', (e) =>
+        @onDragStart(e)
+    @on 'dragenter', '.entry.directory > .header', (e) =>
+        @onDragEnter(e)
+    @on 'dragleave', '.entry.directory > .header', (e) =>
+        @onDragLeave(e)
+    @on 'dragover', '.entry', (e) =>
+        @onDragOver(e)
+    @on 'drop', '.entry', (e) =>
+        @onDrop(e)
```
maxbrunsfeld: It looks like this is just a formatting change. In general with PRs, it's good to avoid extraneous changes like this, in order to ensure that the change merges easily, and also because the maintainers may prefer their existing formatting.

spec/tree-view-spec.coffee
maxbrunsfeld: I think the `getAttribute` assertion would be sufficient. In our test framework, Jasmine, failures don't halt the test, so it will still proceed to execute the next assertion anyway.

spec/tree-view-spec.coffee
```diff
@@ -2462,6 +2476,44 @@ describe "TreeView", ->
...
+        console.log(dirView[0].getAttribute('draggable'))
```
maxbrunsfeld: Stray `console.log` in there.

spec/tree-view-spec.coffee
```diff
@@ -2486,6 +2538,19 @@ describe "TreeView", ->
...
+        expect(dirView[0].getAttribute('draggable')).toBe('true')
+        expect(fileView1[0].hasAttribute('draggable')).toBe(true)
+        expect(fileView1[0].getAttribute('draggable')).toBe('true')
+        expect(fileView2[0].hasAttribute('draggable')).toBe(true)
+        expect(fileView2[0].getAttribute('draggable')).toBe('true')
+        expect(fileView3[0].hasAttribute('draggable')).toBe(true)
+        expect(fileView3[0].getAttribute('draggable')).toBe('true')
+
+
+
```
maxbrunsfeld: Extra whitespace here. We just need one blank line.

spec/tree-view-spec.coffee
50Wliu: Nevermind - misread `getAttribute` vs `hasAttribute`.

maxbrunsfeld: But your comment is still valid: the test description should be something like:

```coffee
it "prevents directories and files from being dragged", ->
```

lib/tree-view.coffee
KenBridgeman: Yeah, I had a hack in there actually that was stupid and I removed it.

What I really needed to do was refresh that file, but I didn't.  I will do that next time.

maxbrunsfeld:
Left various minor comments. Other than that, this PR is good! Unfortunately, as @50Wliu mentioned, we have decided *not* to add this option at this time. In general, we like to avoid a proliferation of settings, where there are alternatives. Thanks for delving into the Atom code, @KenBridgeman; I hope to see more work by you!

spec/tree-view-spec.coffee
KenBridgeman: Noted.

spec/tree-view-spec.coffee
KenBridgeman: Noted.

KenBridgeman: Thanks very much for taking the time to give me a quick review.

strindhaug:
@maxbrunsfeld What "alternatives" are you talking about? Are you saying there is a way to disable this damned drag and drop misfeature (that doesn't involve forking the code and commenting out the code)?

This is a bug you've ignored since 2015 and this is the second (the first was #623) pull request (that tries to fix the bug of not being able to disable drag and drop completely) you've effectively wontfixed... All in the name of "avoiding configuration" for some reason... WTF? This is an app for developers, we can handle a having few configurations; in fact we generally love having configurations. So why won't you ever allow us to fix this bug!?

50Wliu:
As the Atom maintainers have agreed that we will not be adding a configuration setting for drag-and-drop, I'm going to close this PR.  Thanks again for the contributions @KenBridgeman 😀.

lee-dohm:
@strindhaug your comment was deleted as a violation of the [Atom Code of Conduct](https://github.com/atom/atom/blob/master/CODE_OF_CONDUCT.md) as is insulting or derogatory. You may consider this an official warning.

PR #751, opened 2016-02-27
feat(option): added an option to auto-reveal files on changing them

kumarharsh:
this option controls the revealing of files automatically as they
are switched in the editor panes.

PR split as requested by @benogle in #715


nathansobo:
For the record, this PR was actually merged with some changes in #754. Many thanks!

Issue #426, opened 2015-04-13
Expand to show current file in tree view


alexjurkiewicz:
Sometimes I open a file in my project with cmd-t or via the "project find results" tab. It would be nice if the tree view auto-expanded to and highlighted the same file in these cases.

PR #754, opened 2016-03-04
Add option to auto-reveal tree view entries when they become the active pane item

nathansobo:
Closes #751

This PR applies some minor cleanup to #751. Thanks to @kumarharsh for the contribution!

@kumarharsh: In case you're interested:

* I moved your test and went ahead and combined two tests into one, which tested the basic behavior as well as parent directory expansion.
* In the future, we *prefer* commit messages and PR titles to be phrased in the imperative voice. "Add option" rather than "Added option". That said, I'd rather have the contribution than worry too much about this so I'll go ahead and merge the commit as-is.

Thanks again!!

### DIFF
--- a/lib/tree-view.coffee
+++ b/lib/tree-view.coffee
@@ -137,6 +137,7 @@ class TreeView extends View
 
     @disposables.add atom.workspace.onDidChangeActivePaneItem =>
       @selectActiveFile()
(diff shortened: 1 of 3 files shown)


kumarharsh:
@nathansobo Thanks :) and yes, sorry about that commit message, point taken!

Ben3eeE:
Closes https://github.com/atom/tree-view/issues/426?

PR #756, opened 2016-03-05
Fix #732: Tree-view API is not being called on startup

### LINKED ISSUE #732: Tree-view API is not being called on startup
I am trying to use https://github.com/atom/tree-view#api in my https://github.com/azawawi/atom-perl6-editor-tools/blob/master/lib/file-icons.js#L12 to display custom file icons for Perl 6 file type. The problem is that `iconClassForPath(path)` is called only after a collapse/expand folder event and not on atom environment startup.

I have been trying to work around it via the following code but it is not working either.

```js
atom.workspace.getLeftPanels()[0].emitter.emit("tree-view:collapse-directory")
atom.workspace.getLeftPanels()[0].emitter.emit("tree-view:expand-directory")
```

Your help is appreciated.

azawawi:
This is fixes #732: Tree-view API is not being called on startup courtesy of ssorallen++.

### DIFF
--- a/lib/main.coffee
+++ b/lib/main.coffee
@@ -34,6 +34,7 @@ module.exports =
   consumeFileIcons: (service) ->
     FileIcons.setService(service)
     @fileIconsDisposable = service.onWillDeactivate -> FileIcons.resetService()
+    @treeView?.updateRoots()
 
   serialize: ->
     if @treeView?


lib/main.coffee
```diff
@@ -34,6 +34,7 @@ module.exports =
   consumeFileIcons: (service) ->
     FileIcons.setService(service)
     @fileIconsDisposable = service.onWillDeactivate -> FileIcons.resetService()
```
acontreras89: Shouldn't we call `@treeView?.updateRoots()` from within the `service.onWillDeactivate` callback as well?

PR #768, opened 2016-03-12
Fix exception when double clicking opened file after activation

### LINKED ISSUE #474: Unable to open tree view
Since Atom commit https://github.com/atom/atom/commit/4300ba6bb3155eb4340eb9bda42a5ab171bd24c4  the treeview doesn't show up anymore. Trying to open it (command "Tree View: Show") appears to do nothing. Commit https://github.com/atom/atom/commit/57610518f7edbc1f16c9ed8ed08dce457b85f34e still works.
Removing `~/.atom` does not change anything.
Running on Ubuntu Linux 15.04 kernel 3.16.0-36-generic x86_64.
Node v0.10.33, npm v2.5.1.

Startup messages (for good and bad commit):
```
$ atom -n -d -f -w
[7409:0609/132223:ERROR:browser_main_loop.cc(170)] Running without the SUID sandbox! See https://code.google.com/p/chromium/wiki/LinuxSUIDSandboxDevelopment for more information on developing with the sandbox on.
App load time: 230ms
[7409:0609/132224:INFO:CONSOLE(49)] "Window load time: 800ms", source: file:///[...]/share/atom/resources/app.asar/static/index.js (49)
[7409:0609/132225:WARNING:accelerator_util.cc(181)] Invalid accelerator token: 
[7409:0609/132225:WARNING:accelerator_util.cc(181)] Invalid accelerator token: 
[7409:0609/132232:WARNING:accelerator_util.cc(181)] Invalid accelerator token: 
```

BinaryMuse:
The incorrect behavior only exhibits itself on the very first interaction with tree-view after activation, and only if the currently active item is the same as the item being opened.

Thanks to @m133225 for the fix!

/cc @kuychaco 

Fixes #474

### DIFF
--- a/lib/tree-view.coffee
+++ b/lib/tree-view.coffee
@@ -213,7 +213,7 @@ class TreeView extends View
         else if entry instanceof DirectoryView
           entry.toggleExpansion(isRecursive)
       when 2
-        if entry instanceof FileView & @openedItem?
+        if entry instanceof FileView and @openedItem?
           @openedItem.then (item) ->
             activePane = atom.workspace.getActivePane()
             if activePane?.getPendingItem?
--- a/lib/tree-view.coffee
+++ b/lib/tree-view.coffee
@@ -213,7 +213,7 @@ class TreeView extends View
         else if entry instanceof DirectoryView
           entry.toggleExpansion(isRecursive)
       when 2
-        if entry instanceof FileView
+        if entry instanceof FileView & @openedItem?
           @openedItem.then (item) ->
             activePane = atom.workspace.getActivePane()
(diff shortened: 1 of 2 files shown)


BinaryMuse:
Ah, it looks like the proposed fix doesn't handle the case where the file being double-click right after package activation is a pending tab (in which case it should be made permanent). I've added a test for this, but it won't run on CI because CI tests against Atom stable.

@kuychaco, do you see any problems with the new fix I've proposed in the latest commit? The tests all pass but I'm not sure of any edge case behavior I may be missing from when you originally implemented the feature.

PR #769, opened 2016-03-13
Move away from deprecated Electron require syntax

50Wliu:
Required for atom/atom#10983.

Will merge once CI passes.


50Wliu:
:green_heart:

PR #771, opened 2016-03-15
Delete empty file without confirmation

### LINKED ISSUE #693: Enhancement: Don't prompt/confirm when deleting empty files
This is one subtle, but really nice behavior of Sublime. When creating files sometimes I get the name wrong and, for whatever reason, I'd rather delete the file and re-create than just rename it. Atom adds friction by prompting "Are you sure you want to delete the selected item?" – this gets old. There's no real consequence for deleting an empty file, so the prompt would be great to remove. Adjusting the behavior would also be great when removing files you don't end up needing after scaffolding a project.

TLDR: the prompt gets to be a bit much.

(moved from https://github.com/atom/atom/issues/10313)

m133225:
Fixes #693

Simple change to delete without prompting when the only item being deleted is an empty file.

### DIFF
--- a/lib/tree-view.coffee
+++ b/lib/tree-view.coffee
@@ -560,13 +560,13 @@ class TreeView extends View
           moveToTrash(selectedPaths)
         "Cancel": null
 
-  moveToTrash: (paths) ->
+  moveToTrash: (pathsToDel) ->
     failedDeletions = []
-    for path in paths
-      if not shell.moveItemToTrash(path)
-        failedDeletions.push "#{path}"
-      if repo = repoForPath(path)
-        repo.getPathStatus(path)
+    for pathToDel in pathsToDel
+      if not shell.moveItemToTrash(pathToDel)
+        failedDeletions.push "#{pathToDel}"
+      if repo = repoForPath(pathToDel)
+        repo.getPathStatus(pathToDel)
     if failedDeletions.length > 0
       atom.notifications.addError "The following #{if failedDeletions.length > 1 then 'files' else 'file'} couldn't be moved to trash#{if process.platform is 'linux' then " (is `gvfs-trash` installed?)" else ""}",
         detail: "#{failedDeletions.join('\n')}"
--- a/lib/tree-view.coffee
+++ b/lib/tree-view.coffee
@@ -546,22 +546,31 @@ class TreeView extends View
           buttons: ['OK']
         return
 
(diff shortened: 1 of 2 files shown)


lib/tree-view.coffee
```diff
@@ -546,23 +546,32 @@ class TreeView extends View
           buttons: ['OK']
         return
 
+    if selectedPaths.length is 1 and fs.isFileSync(selectedPaths[0])
+      if fs.getSizeSync(selectedPaths[0]) is 0
+        @moveToTrash(selectedPaths)
+        return
+
+    moveToTrash = @moveToTrash
```
thomasjo: Not necessary. Make the arrow in the `atom.confirm` button callback below a "fat arrow" to capture the `this` scope, then delete this line.

lib/tree-view.coffee
```diff
@@ -546,23 +546,32 @@ class TreeView extends View
...
     atom.confirm
       message: "Are you sure you want to delete the selected #{if selectedPaths.length > 1 then 'items' else 'item'}?"
       detailedMessage: "You are deleting:\n#{selectedPaths.join('\n')}"
       buttons:
         "Move to Trash": ->
```
thomasjo: Make this a fat arrow; `=>` to avoid the unnecessary `moveToTrash` variable.

lib/tree-view.coffee
```diff
@@ -546,23 +546,32 @@ class TreeView extends View
...
-          if failedDeletions.length > 0
-            atom.notifications.addError "The following #{if failedDeletions.length > 1 then 'files' else 'file'} couldn't be moved to trash#{if process.platform is 'linux' then " (is `gvfs-trash` installed?)" else ""}",
-              detail: "#{failedDeletions.join('\n')}"
-              dismissable: true
+          moveToTrash(selectedPaths)
         "Cancel": null
 
+  moveToTrash: (pathsToDel) ->
+    failedDeletions = []
+    for pathToDel in pathsToDel
```
thomasjo: Rename these variables to `pathToDelete` and `pathsToDelete`.

m133225:
@thomasjo Made the suggested changes!

jerone:
Should empty files that are under source control be excluded from this feature?

lee-dohm:
Upon further review, we would rather not change this behavior. Thanks very much for the effort that was put into this PR.

PR #789, opened 2016-03-28
Use sync fs for testing

joshaber:
This is needed to support https://github.com/atom/atom/pull/11280.

So it seems like these tests were always racing: we depended on loading the `realpath` before emitting change events. It’s just that we usually won that race. Now we use synchronous FS during tests to *ensure* we win the race.

I don’t love this. It means testing isn’t quite doing the same thing as the Real Version. But I also don’t want to make dramatic changes and make https://github.com/atom/tree-view/pull/624 more difficult to merge.

/cc @thedaniel to make sure I’m not being crazy here.

### DIFF
--- a/lib/directory.coffee
+++ b/lib/directory.coffee
@@ -9,7 +9,7 @@ realpathCache = {}
 
 module.exports =
 class Directory
-  constructor: ({@name, fullPath, @symlink, @expansionState, @isRoot, @ignoredPatterns}) ->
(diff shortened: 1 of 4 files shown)


thedaniel:
I don't love it either, but maybe we can make a point to refactor it out after landing #624?

joshaber:
I haven't looked at #624 but I wouldn't be surprised if you already ran into this and hopefully fixed it.

PR #791, opened 2016-03-31
Fix specs on Windows

50Wliu:
This PR fixes some of the specs that currently fail on Windows - see each individual commit for more details.

I'm still working on this - many of the remaining failures on my machine either have to deal with problems out of the scope of this PR (eg #649) or are flaky enough that I can't reproduce them at all when focused.

### DIFF
--- a/spec/default-file-icons-spec.coffee
+++ b/spec/default-file-icons-spec.coffee
@@ -41,7 +41,7 @@ describe 'DefaultFileIcons', ->
       filePath = path.join(tempDir, 'foo.bar')
       linkPath = path.join(tempDir, 'link.bar')
       fs.writeFileSync(filePath, '')
-      fs.symlinkSync(filePath, linkPath)
+      fs.symlinkSync(filePath, linkPath, 'junction')
 
       expect(fileIcons.iconClassForPath(linkPath)).toEqual('icon-file-symlink-file')
 
@@ -49,6 +49,6 @@ describe 'DefaultFileIcons', ->
       filePath = path.join(tempDir, 'foo.zip')
       linkPath = path.join(tempDir, 'link.zip')
       fs.writeFileSync(filePath, '')
-      fs.symlinkSync(filePath, linkPath)
+      fs.symlinkSync(filePath, linkPath, 'junction')
 
       expect(fileIcons.iconClassForPath(linkPath)).toEqual('icon-file-symlink-file')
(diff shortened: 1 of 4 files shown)


50Wliu:
At least in my experience `symlinkSync` has to have the `'junction'` option in order for it to work on Windows - `file` and `directory` have always returned `ENOENT` errors for me.  If that's the general case it would be great to add `junction` everywhere where `symlinkSync` exists.

50Wliu:
I get 101 failures (that's a 8.8% failure rate) when running tests through the command prompt, the majority of which complain about `Error: now has already been spied upon (file:\C:\Users\username\Documents\GitHub\atom\spec\spec-helper.coffee:69:3)`.  However, when running specs through the GUI, I only get 11 failures.  8 of those will be fixed by #649.

Two of the remaining failures are related to Git symlinks:
`Git status decorations -> when the project is a symbolic link to the repository root -> when a file loses its modified status -> it updates its and its parent directories' styles`
```
Expected '<two dom elements*> not to have class 'status-modified'.
```
\* Where one of the DOM elements represents a file and the other a directory structure

The final one is the weirdest...
`when other panels are added -> it should resize normally on the right side -> Expected 133 to be 150`.
This **only** passes when no other specs fail.  Don't believe me?  Try adding a bogus spec (eg `expect(true).toBe false` and focus those two.

50Wliu:
With the specs window in the background and not focused, I get 52 failures.

50Wliu:
After @damieng commented that tree-view specs also crashed on his end, I decided to run them over and over again to see if there were any patterns...here are the results.
```
24x5
37x3
38x2
48x4
62
66x4
73
84
90
91
```
(so for example, spec 24 crashed 5 times.)  Spec 24 was also the one that crashed on AppVeyor.

It's getting late over here and it's time for me to get some :zzz:, but I'll be looking later to see what those specs have in common to try to narrow things down.

spec/default-file-icons-spec.coffee
```diff
@@ -42,24 +41,14 @@ describe 'DefaultFileIcons', ->
       filePath = path.join(tempDir, 'foo.bar')
       linkPath = path.join(tempDir, 'link.bar')
       fs.writeFileSync(filePath, '')
-      try
```
damieng: I put these in because they fail unless you're an Admin - probably want to keep them in as AppVeyor etc. do not run as admin

50Wliu:
The crashes are being caused by any `:contains` selector such as `.find('.directory:contains(dir1)')`.

50Wliu:
Well, looks like nothing has changed between July and now :/.

damieng:
Let's get these merged in and see whats left.

PR #797, opened 2016-04-06
Create new layer to reduce work on high DPI displays

ssorallen:
## TL;DR

This change improves typing performance when the tree-view is visible by decreasing the amount of work done in Chromium on each keypress.

Keypresses in `<atom-text-editor>` force layouts and layer tree updates. On high resolution displays (ex. Macbooks w/ Retina screens), Chromium inadvertently creates a duplicate layer in tree-view that causes layer tree updates to take 3x longer than needed.

## Investigation

Force a new stacking context to prevent a large, duplicate paint layer
from being created for tree-view's scrolling contents that can make the
cost of layer tree updates scale at 3x the size of the layer rather than
the optimal 1x.

On high resolution displays, Chromium handles layers for scrolling
content differently and inadvertently creates a duplicate paint layer
the size of .tree-view-scroller because descendants of the scroller
overlap the auto-created layer.

#### Example of duplicate layers created, no changes applied

Duplicate layers:
screen shot 2016-04-06 at 10 38 58 am

Layer tree update ~60ms (occurs on every keypress in `<atom-text-editor>`):
screen shot 2016-04-06 at 10 38 28 am

#### Example with isolation, new layer

Only one layer for scrolling contents:
screen shot 2016-04-06 at 10 39 54 am

Layer tree update only ~20ms (occurs on every keypress in `<atom-text-editor>`):
screen shot 2016-04-06 at 10 40 36 am

#### w/ duplicate layers

Layer tree update ~1.5ms (occurs on every keypress in `<atom-text-editor>`):
screen shot 2016-04-06 at 10 45 40 am

#### w/ isolation, single layer

Layer tree update ~0.5ms (occurs on every keypress in `<atom-text-editor>`):
screen shot 2016-04-06 at 10 45 05 am

cc @nathansobo @hellendag

### DIFF
--- a/styles/tree-view.less
+++ b/styles/tree-view.less
@@ -51,6 +51,18 @@
 .tree-view {
   flex-grow: 1;
   flex-shrink: 0;
+  /*
+   * Force a new stacking context to prevent a large, duplicate paint layer from
+   * being created for tree-view's scrolling contents that can make the cost of
+   * layer tree updates scale at 3x the size of the layer rather than the
+   * optimal 1x.
+   *
+   * On high resolution displays, Chromium handles layers for scrolling content
+   * differently and inadvertently creates a duplicate paint layer the size of
+   * .tree-view-scroller because descendants of the scroller overlap the
+   * auto-created layer.
+   */
+  isolation: isolate;
   min-width: -webkit-min-content;
   min-height: 100%;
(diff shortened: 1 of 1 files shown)


nathansobo:
Thanks for the help @ssorallen. This was a simple solution to a complex problem.

Issue #703, opened 2016-01-18
Ctrl/Cmd-Drag and drop should copy the selected file / directory


mattqs:
Ctrl drag and drop in file tree does not create a copy of the file/directory (as it is in Visual Studio) instead it just moves the file.

bananastalktome:
:+1: would be very useful to have!

ahnbizcad:
Yes, currently, you can select multiple files and folders, but once you dry to move them by click dragging, it undoes all your effort and selects only the file you clicked, rather than the files you selected.

yamsellem:
I guess `alt-drag` should copy on mac, to keep a consistent behavior with the os, don't you think?

bhj:
Yes, when alt/option-dragging a file from Finder, the green plus sign indicates the file is being copied but is moved when released onto tree-view. I think this should be considered a bug. Furthermore the operation is not undo-able in Finder (not sure if that's possible to implement, but it does compound the issue)

jcollum:
It looks like this was never merged in? #816 was closed for being a dupe. 

PR #800 appears to be unmerged.

50Wliu:
That's correct.  #800 has not been merged yet.

Zelgadis87:
Any updates on this?

egeland:
This would be really great to get merged in - I keep expecting it to just work, and instead it just moves my file...

DeividasK:
+1 for merge

TigersWay:
Any hope for this one?
Hoping to be able to copy files from explorer (windows) to Atom editor/tree-view.

fonziemedia:
Please, someone fix/merge this..

PR #800, opened 2016-04-11
Fixes https://github.com/atom/tree-view/issues/703

nabeelomer:
Fixes https://github.com/atom/tree-view/issues/703 and changes the behavior of drag and drop.


damieng:
I think this just needs the move specs updated?

a7madgamal:
@nabeelomer almost done :medal_sports:

Issue #749, opened 2016-02-25
Clicking on file in tree view that isn't open sets focus on editor


adambuczynski:
When I click a file in tree view that isn't open yet, focus is shifted to the editor (file contents). This is problematic, as I relied on clicking on a file for the focus to stay on the tree view, so that I could hit `m` to rename the file.

I suspect this may have to do with the refactoring around files not being opened as temporary files anymore?

Atom version 1.6.0-beta5, OS X 10.11.2.

lee-dohm:
This is a duplicate of https://github.com/atom/tree-view/issues/199. Please see the comments and suggested packages there. Let us know if you have any questions!

adambuczynski:
@lee-dohm no, it is not a duplicate of that issue.

That issue wants to prevent the files from opening, I don't mind the files being opened, I just want the _focus_ to go to the editor. 

This didn't use to happen in Atom 1.4.x and 1.5.x and earlier, it only started in 1.6.x.

lee-dohm:
My mistake. Thanks for the correction :grinning:

As an aside, including [repro steps](https://github.com/atom/atom/blob/master/CONTRIBUTING.md#how-do-i-submit-a-good-bug-report) can help clarify these differences.

adambuczynski:
No worries, sorry I thought it was clear. Let me specify:

1. Open a project, close all file tabs
2. Single click on a file in the tree view
3. Press `m` to rename the file

Expected: the file rename dialog pops up

Actual: you type the letter `m` in the opened file on top, because after clicking the file focus is shifted to the editor window.

adambuczynski:
@lee-dohm have you had some time to look into this by any chance? It's impairing my productivity a lot, as I keep hitting `m` and having to erase the character and make sure I click the file name twice (not double click) before I rename it.

billyjanitsch:
Based on https://github.com/atom/tree-view/issues/740#issuecomment-185518512 it looks like this is actually intended behavior (introduced in e93148d5ec981ae69df788a3ab9dab6dcd389c0e, see also #742 and #682). The rationale seems to be "you can click on a file and immediately start editing it". However, I agree with @adambuczynski that this is undesirable.

The old behavior (single click to gain focus/preview, double click to open and transfer focus to the editor) aligns with Windows/OSX file explorers, most email clients, Dropbox/Drive, Sublime, etc. It's strange for Atom to diverge from such a common UX standard.

[...]

PR #801, opened 2016-04-12
Keep focus on single click and update to use 1.6stable pending items api

kuychaco:
/cc @BinaryMuse 

Fixes https://github.com/atom/tree-view/issues/749


BinaryMuse:
👍, :shipit:

Issue #199, opened 2014-07-30
tree view automatically opens the file i click on. how to disable?


ninjajs:
tree view automatically opens the file i click on. how do I disable this? i only want files to open when i double click.

liquidvisual:
This is what you need: [https://atom.io/packages/chary-tree-view](Charity Tree View package)

janmarek:
:+1:

jbosse:
Yes, this frustrates me a great deal. I cannot find an option to disable this behavior. +1

ceyhunkerti:
+1

arielpontes:
:+1:  I agree. This should totally be "disableable".

Mothil93:
I'm joining this: +1!
It's quite frustrating to look at some files and end up with 10 open ones, when the only thing I wanted to do was to check them out. Making them only respond to double click would be so nice.

janmarek:
@Mothil93 It helps but it's not the same. Normal behavior is to select the file in tree view and do nothing else.

Mothil93:
@janmarek I agree. I didn't notice until after I edited it, but still... It's a lot better. Ideally it should be as you explained, though.

rondonjon:
+1 for a settings option to change the default behaviour.
I would prefer to "select" on single click and "open" on double click.

liquidvisual:
Agree +1

andela-anandaa:
Sure, will be nice to have it as preview (`Sublime Text` way), but not opening the file. +1 this!

erik-landvall:
+1

andrewnester:
+1

victor-torres:
Hey, you should try to configure sublime-tabs plugin, which seems to be installed by default in recent Atom builds.

1. Settings
2. Packages
3. Installed Packages
4. Search for 'tabs' - it's a core package
5. In its settings check 'Use preview tabs'

mattfrey:
+1 Terrible feature, makes the editor a huge PITA for performing file system operations - creating new files, folders, deleting files/folders etc. All these things become multi-step processes including closing a window, and in the case of deleting my db files between iterations, having to dismiss a 'large file may be slow in preview...' dialog.

Zireael07:
@mattfrey, did you see the tabs -> use preview tabs option?

mattfrey:
[...]

PR #803, opened 2016-04-12
Don't open file on single-click unless core.allowPendingPaneItems=true

kuychaco:
Also cleaned up specs for when files are clicked. Removed unnecessary tests that cover behavior implemented in Atom core.

/cc @BinaryMuse

### DIFF
--- a/lib/tree-view.coffee
+++ b/lib/tree-view.coffee
@@ -211,7 +211,8 @@ class TreeView extends View
     else if entry instanceof FileView
       detail = e.originalEvent?.detail ? 1
       if detail is 1
-        atom.workspace.open(entry.getPath(), pending: true, activatePane: false)
+        if atom.config.get('core.allowPendingPaneItems')
+          atom.workspace.open(entry.getPath(), pending: true, activatePane: false)
       else if detail is 2
         atom.workspace.open(entry.getPath())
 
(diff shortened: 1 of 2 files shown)


Alhadis:
May I ask the reasons why this was done?

I barely upgraded 10 minutes ago and the need to double-click everything all of a sudden is already driving me crazy...

kuychaco:
@Alhadis #199 contains the discussion leading to this change

Alhadis:
Ah, I see.

Would you consider having a user-setting to keep the old behaviour? I'm happy to submit the PR. =)

vemv:
* I want keep opening files with a single click, as I've done with Atom for a couple years
* I don't want to set `allowPendingPaneItems = true`, it's not a feature of my liking (it was recently introduced - it's not like it's a traditional Atom feature)
* Furthermore two totally different things (allowPendingPaneItems, open-on-single-click) shouldn't be conflated. Please add a separate option instead: openOnSingleClick.

**Please stop breaking everyone's workflow**. Atom does this a few times a year. Pretty much barbaric.

Alhadis:
@vemv Couldn't agree more, which is why I [submitted a PR for this](https://github.com/atom/tree-view/pull/804).

PR #825, opened 2016-04-29
Allow file-icons service to set multiple class names

Alhadis:
Packages that provide the `atom.file-icons` service are currently limited to setting only one class name per file. This is because any attempt to add a DOMTokenList item with a whitespace character will trigger an exception, as per the spec.

However, this is particularly limiting for packages that need flexibility. For example, the [File Icons](https://github.com/DanBrooker/file-icons) package will be adding two different classes: one for the icon, and another for its colour.

I've amended the service consumption routine to allow a list of classes to be specified as either an array or a whitespace-delimited string (analogous to jQuery's approach).

So this:
```js
service.iconClassForPath("gear-icon medium-orange");
```

will now lead to this:
```html
<span class="name icon gear-icon medium-orange">
```

... instead of, well, this:

### DIFF
--- a/lib/file-view.coffee
+++ b/lib/file-view.coffee
@@ -19,7 +19,7 @@ class FileView extends HTMLElement
     @fileName.dataset.name = @file.name
     @fileName.dataset.path = @file.path
 
-    iconClass = FileIcons.getService().iconClassForPath(@file.path)
+    iconClass = FileIcons.getService().iconClassForPath(@file.path, "tree-view")
     if iconClass
       unless Array.isArray iconClass
         iconClass = iconClass.toString().split(/\s+/g)
--- a/lib/file-view.coffee
+++ b/lib/file-view.coffee
@@ -19,7 +19,7 @@ class FileView extends HTMLElement
     @fileName.dataset.name = @file.name
     @fileName.dataset.path = @file.path
 
-    iconClass = FileIcons.getService().iconClassForPath(@file.path, @file.getMetadata())
+    iconClass = FileIcons.getService().iconClassForPath(@file.path)
     if iconClass
       unless Array.isArray iconClass
         iconClass = iconClass.toString().split(/\s+/g)
--- a/lib/file-view.coffee
+++ b/lib/file-view.coffee
@@ -19,7 +19,7 @@ class FileView extends HTMLElement
     @fileName.dataset.name = @file.name
     @fileName.dataset.path = @file.path
 
-    iconClass = FileIcons.getService().iconClassForPath(@file.path, @)
+    iconClass = FileIcons.getService().iconClassForPath(@file.path, this)
     if iconClass
       unless Array.isArray iconClass
         iconClass = iconClass.toString().split(/\s+/g)
--- a/lib/file-view.coffee
+++ b/lib/file-view.coffee
@@ -19,7 +19,7 @@ class FileView extends HTMLElement
     @fileName.dataset.name = @file.name
     @fileName.dataset.path = @file.path
 
-    iconClass = FileIcons.getService().iconClassForPath(@file.path, @file)
+    iconClass = FileIcons.getService().iconClassForPath(@file.path, @)
     if iconClass
       unless Array.isArray iconClass
         iconClass = iconClass.toString().split(/\s+/g)
--- a/lib/file-view.coffee
+++ b/lib/file-view.coffee
@@ -19,7 +19,11 @@ class FileView extends HTMLElement
     @fileName.dataset.name = @file.name
     @fileName.dataset.path = @file.path
 
-    @fileName.classList.add(FileIcons.getService().iconClassForPath(@file.path))
+    iconClass = FileIcons.getService().iconClassForPath(@file.path)
+    if iconClass
+      unless Array.isArray iconClass
+        iconClass = iconClass.toString().split(/\s+/g)
+      @fileName.classList.add(iconClass...)
 
     @subscriptions.add @file.onDidStatusChange => @updateStatus()
     @updateStatus()
--- a/lib/file-view.coffee
+++ b/lib/file-view.coffee
@@ -19,7 +19,7 @@ class FileView extends HTMLElement
     @fileName.dataset.name = @file.name
     @fileName.dataset.path = @file.path
 
-    iconClass = FileIcons.getService().iconClassForPath(@file.path, this)
+    iconClass = FileIcons.getService().iconClassForPath(@file.path, @file.getMetadata())
     if iconClass
       unless Array.isArray iconClass
         iconClass = iconClass.toString().split(/\s+/g)
--- a/lib/file-view.coffee
+++ b/lib/file-view.coffee
@@ -19,7 +19,7 @@ class FileView extends HTMLElement
     @fileName.dataset.name = @file.name
     @fileName.dataset.path = @file.path
 
-    iconClass = FileIcons.getService().iconClassForPath(@file.path, "tree-view")
+    iconClass = FileIcons.getService().iconClassForPath(@file.path)
     if iconClass
       unless Array.isArray iconClass
         iconClass = iconClass.toString().split(/\s+/g)
--- a/lib/file-view.coffee
+++ b/lib/file-view.coffee
@@ -19,7 +19,7 @@ class FileView extends HTMLElement
     @fileName.dataset.name = @file.name
     @fileName.dataset.path = @file.path
 
-    iconClass = FileIcons.getService().iconClassForPath(@file.path)
+    iconClass = FileIcons.getService().iconClassForPath(@file.path, @file)
     if iconClass
       unless Array.isArray iconClass
         iconClass = iconClass.toString().split(/\s+/g)
--- a/lib/file.coffee
+++ b/lib/file.coffee
@@ -36,7 +36,7 @@ class File
   onDidStatusChange: (callback) ->
     @emitter.on('did-status-change', callback)
 
-  # Subscribe to the project' repo for changes to the Git status of this file.
+  # Subscribe to the project's repo for changes to the Git status of this file.
   subscribeToRepo: ->
     repo = repoForPath(@path)
     return unless repo?
@@ -67,6 +67,3 @@ class File
 
   isPathEqual: (pathToCompare) ->
     @path is pathToCompare or @realPath is pathToCompare
-
-  # Return a snapshot of the instance's filesystem properties for API consumption
-  getMetadata: -> {@symlink, @realPath, @status, @stats}
--- a/lib/file.coffee
+++ b/lib/file.coffee
@@ -67,3 +67,6 @@ class File
 
   isPathEqual: (pathToCompare) ->
     @path is pathToCompare or @realPath is pathToCompare
+
+  # Return a snapshot of the instance's filesystem properties for API consumption
+  getMetadata: -> {@symlink, @realPath, @status, @stats}
--- a/spec/file-icons-spec.coffee
+++ b/spec/file-icons-spec.coffee
@@ -1,3 +1,7 @@
+fs = require 'fs-plus'
+temp = require('temp').track()
+path = require 'path'
+
 DefaultFileIcons = require '../lib/default-file-icons'
 FileIcons = require '../lib/file-icons'
 
@@ -21,3 +25,45 @@ describe 'FileIcons', ->
     FileIcons.resetService()
 
     expect(FileIcons.getService()).not.toBe(service)
+
+  
+  describe 'Class handling', ->
+    [workspaceElement, treeView, files] = []
+    
+    beforeEach ->
+      rootDirPath = fs.absolute(temp.mkdirSync('tree-view-root1'))
+      
+      for i in [1..3]
+        filepath = path.join(rootDirPath, "file-#{i}.txt")
+        fs.writeFileSync(filepath, "Nah")
+      
+      atom.project.setPaths([rootDirPath])
+      workspaceElement = atom.views.getView(atom.workspace)
+      jasmine.attachToDOM(workspaceElement)
+
+      FileIcons.setService
+        iconClassForPath: (path, file) ->
+          [name, id] = path.match(/file-(\d+)\.txt$/)
+          switch id
+            when "1" then 'first second'
+            when "2" then ['first', 'second']
+            when "3" then file.constructor.name
+
+      waitsForPromise ->
+        atom.packages.activatePackage('tree-view')
+      
+      runs ->
+        treeView = atom.packages.getActivePackage("tree-view").mainModule.createView()
+        files = workspaceElement.querySelectorAll('li[is="tree-view-file"]')
+      
+    afterEach ->
+      temp.cleanup()
+  
+    it 'allows multiple classes to be passed', ->
+      expect(files[0].fileName.className).toBe('name icon first second')
+
+    it 'allows an array of classes to be passed', ->
+      expect(files[1].fileName.className).toBe('name icon first second')
+
+    it 'passes a file reference as iconClassForPath\'s second argument', ->
+      expect(files[2].fileName.className).toBe('name icon File')
--- a/spec/file-icons-spec.coffee
+++ b/spec/file-icons-spec.coffee
@@ -65,5 +65,5 @@ describe 'FileIcons', ->
     it 'allows an array of classes to be passed', ->
       expect(files[1].fileName.className).toBe('name icon first second')
 
-    it 'passes a file reference as iconClassForPath\'s second argument', ->
-      expect(files[2].fileName.className).toBe('name icon File')
+    it 'passes a FileView reference as iconClassForPath\'s second argument', ->
+      expect(files[2].fileName.className).toBe('name icon tree-view-file')
--- a/spec/file-icons-spec.coffee
+++ b/spec/file-icons-spec.coffee
@@ -1,7 +1,3 @@
-fs = require 'fs-plus'
-temp = require('temp').track()
-path = require 'path'
-
 DefaultFileIcons = require '../lib/default-file-icons'
 FileIcons = require '../lib/file-icons'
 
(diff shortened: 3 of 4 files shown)


Alhadis:
@lee-dohm Anything I can do to move this along?

Alhadis:
@50Wliu, any progress?

I don't mean to come across as impatient, but two months of solid silence *does* tend to leave one feeling uncomfortable.

50Wliu:
I'm not the person to ask about progress generally.  All I do is add the appropriate labels and move on.

Alhadis:
Ah, right. Sorry to bother then...

I guess I'll just keep waiting then..

lib/file-view.coffee
```diff
@@ -19,7 +19,11 @@ class FileView extends HTMLElement
     @fileName.dataset.name = @file.name
     @fileName.dataset.path = @file.path
 
-    @fileName.classList.add(FileIcons.getService().iconClassForPath(@file.path))
+    iconClass = FileIcons.getService().iconClassForPath(@file.path, this)
```
as-cii: Could you explain why we need to pass the current `FileView` instance to the file icons service? Maybe it's not a huge deal, but seems like the service shouldn't need it in order to provide an icon. What am I missing? 💭

lib/file-view.coffee
```diff
@@ -19,7 +19,11 @@ class FileView extends HTMLElement
...
+    if iconClass
+      unless Array.isArray iconClass
```
as-cii: 💡 What do you think about slightly changing this into: 

```coffee
if Array.isArray(iconClass)
  @fileName.classList.add(iconClass…)
else
  @fileName.classList.add(iconClass)
```

spec/file-icons-spec.coffee
```diff
@@ -21,3 +25,45 @@ describe 'FileIcons', ->
     FileIcons.resetService()
 
     expect(FileIcons.getService()).not.toBe(service)
+
```
as-cii: Seems like we could remove this newline 🔥

lib/file-view.coffee
Alhadis: It allows the provider to understand the context of where the icon is being requested. To give a concrete example, [File-Icons](https://github.com/DanBrooker/file-icons) has a *"Show in tabs"* setting that should determine whether an icon gets displayed in a tab or not.

spec/file-icons-spec.coffee
```diff
@@ -21,3 +25,45 @@ describe 'FileIcons', ->
...
+  
+  describe 'Class handling', ->
```
as-cii: What about moving this spec into `tree-view-spec.coffee`? Testing the interaction between `FileView` and the file icon service seems to belong more to that file. 💡

lib/file-view.coffee
Alhadis: That'll try coercing the array into a string.

Alhadis: In other words, `["first", "second"]` will end up being concatenated and added as a single class. E.g.:

```html
<element class="first,second">
```

lib/file-view.coffee
as-cii: Interesting, I wonder if we should put that setting on the individual packages instead of customizing every package UI via `File-Icons`. Both tree-view and tabs could expose a `disableCustomIcons` setting which, if enabled, wouldn't call the file icons service at all, keeping its default icons.

What do you think? Seems like this would be more customizable and would allow `File-Icons` to not implement a setting for every package that uses it. 💭

spec/file-icons-spec.coffee
Alhadis: Erm, why does the `file-icons-spec` exist in a separate file?

lib/file-view.coffee
as-cii: Are you referring to `@fileName.classList.add(iconClass…)`? I think the trailing `…` will work fine, adding `first` and `second` appropriately, without commas. As for when `iconClass` is a string with spaces, the code I proposed will throw an exception like the following:

```
Uncaught DOMException: Failed to execute 'add' on 'DOMTokenList': The token provided ('foo bar') contains HTML space characters, which are not valid in tokens.
```

But I think this was the case also before this pull-request, and it seems nice to let `classList.add` enforce its default preconditions. Thoughts?

lib/file-view.coffee
Alhadis: Not a bad idea, actually. =) Want me to tackle that?

as-cii: That would be 🆒! How about avoiding to pass `this` in the current pull-request and open a new one with the new config parameter?

spec/file-icons-spec.coffee
as-cii: I think that file is meant to test `FileIcons` without involving the UI, and just assert that it can be configured properly. This one seems more like an integration test that verifies that the tree-view package is able to deal with arrays (and strings) coming from a service.

lib/file-view.coffee
Alhadis: I think it's safe to assume many authors (particularly those used to working with jQuery) might want to pass multiple classes as a space-separated string. E.g., they wouldn't expect `first second` to throw an error.

lib/file-view.coffee
Alhadis: There's more to it than just knowing context. Additional information about *where* the icon will be displayed needs to be provided. In my case, I need to know if the target is a directory, a symlink, or already has existing classes applied. [You can see for yourself here](https://github.com/DanBrooker/file-icons/blob/29218e42fbdbd9d877ec9eb5c3a76848f282e016/lib/service/icon-service.coffee#L146).

spec/file-icons-spec.coffee
Alhadis: Alright. Did you want me to merge it into `tree-view-spec`? :)

Alhadis: Alright. Did you want me to merge it into `tree-view-spec`? :)

lib/file-view.coffee
Alhadis: Basically, I feel the most sustainable solution in long-term is to give service providers an opportunity to query the calling object for additional information.

spec/file-icons-spec.coffee
Alhadis: I will, just let me know if I should move the other specs as well. =)

lib/file-view.coffee
Alhadis: @as-cii Do you think a cleaner approach would be to pass an object like this as the second argument...?

```
{
	context: "tab",
	target: this
}
```

spec/file-icons-spec.coffee
Alhadis: Alright, on it. =)

Hopefully that [flaky issue](https://github.com/atom/tree-view/pull/859#issuecomment-234780696) with the random spec-failure won't bite me...

lib/file-view.coffee
as-cii: I feel like passing extra information to the file icon service needs to be done in a more generic manner, without tying such state to a specific package: I believe we should design services to be general enough to work with other packages that may not necessarily fit into the `tab` and `tree-view` categories. Passing `this` looks somewhat suspicious too, because I am still unclear on the interface that `target` should comply to in order to be passed to `iconClassForPath`.

After taking a look at https://github.com/DanBrooker/file-icons/blob/29218e42fbdbd9d877ec9eb5c3a76848f282e016/lib/service/icon-service.coffee#L146, I feel like we could circumvent this in two ways:

1. For symlinks, query the filesystem in the package itself. That is, call https://nodejs.org/api/fs.html#fs_class_fs_stats and check whether the file is a symlink or not.
2. For the `Show In Tabs` setting, put a configuration setting into each package.

Fundamentally I think there isn't any knowledge (that isn't also available elsewhere) that each package needs to supply to file icon services, and passing state around looks more error-prone than letting the service retrieve pieces of information for itself. Later, when we'll have a better abstraction for file system nodes, we might expose a method called `iconClassForFileSystemNode(node)`, and `node` might have a bunch of facility utilities (such as `isSymlink`) that will allow us to remove some duplication. 

For now, however, it seems like the above proposal should provide a high degree of freedom in the implementation of the service, which makes this design choice more appealing to me.

lib/file-view.coffee
as-cii: > I think it's safe to assume many authors (particularly those used to working with jQuery) might want to pass multiple classes as a space-separated string.

I wonder if we could just deprecate the string option altogether, and just call this method `iconClassesForPath`, which always returns an array...

Anyways, I think it's okay to be less strict on this method and accept both space-separated strings as well as arrays, let's keep this as it is.

lib/file-view.coffee
Alhadis: Thing is, I want to avoid extra filesystem calls if I can help it. This information is already accessible on the object at the time of the request... I agree that passing state around can introduce potential complications, but I still feel extra metadata about the file itself should be supplied.

Damn well agreed with the `Show icons in {tabs/tree-view}`, though.

Alhadis: Could we (perhaps just for now) pass the relevant `File` instance, without passing a reference to the calling object?

as-cii: > Thing is, I want to avoid extra filesystem calls if I can help it.

Yeah I totally understand that, but that said I don't think neither the tree-view nor tabs should be the sources of truth about that kind of information, and enforcing this pattern even we see its flaws right now seems the wrong thing to do. Atom could be where we query and process file stats, but we don't have this abstraction yet unfortunately.

Did you benchmark how file icons performs when issuing fs calls? How better it performs after putting a cache behind it? Maybe you could warm the cache in the background on a separate process? Alternatively, since this is an IO-bound operation, we might think to make the method asynchronous and apply the classes to the DOM node when the operation completes.

Alhadis: > Did you benchmark how file icons performs when issuing fs calls? How better it performs after putting a cache behind it?

It's not just that. Even if we cache the results for filesystem calls, we still have to monitor the paths to respond to filesystem changes. For instance, a path marked as a symlink might change, and be replaced with something that *isn't* symlinked. Caching is one thing, but having so many different listeners for separate files is another. And given the effort one would go to to reliably ascertain this sort of information, it feels kinda cruel to force package authors down that road.

Maybe just a flat object of bare minimum information, no references to package-defined classes, or anything?

spec/file-icons-spec.coffee
Alhadis: I'm having a hard time getting these specs to work in the `tree-view-spec` file. :( They're failing and I can't figure out why and where. I haven't even changed any of the code.

lib/file-view.coffee
as-cii: > It's not just that. Even if we cache the results for filesystem calls, we still have to monitor the paths to respond to filesystem changes.

That's a great point and I agree it's both unfortunate and less than ideal. What I am worried about is inconsistency, though: for example, what should we pass from the `tabs` package? After a quick glance, it doesn't seem to have any reference to a `File` object, and therefore would make the icon service report an incorrect information. Maybe we could let `tabs` conform to the same interfaces as `tree-view` (i.e. by querying the filesystem), but wouldn't that lead to the same performance considerations we're making for the `file-icons` package?

I think having a better abstraction in Atom core (where we provide a central facility that collects information about the filesystem and watches files as they change over time) would be the optimal scenario, but in the meantime it seems like the only alternative to return a correct result that is also consistent across calls is to bake fs queries into the file icons service package. What do you think? 💭

Alhadis: > After a quick glance, it doesn't seem to have any reference to a File object,

Here:
```coffee
TabView.item?.buffer.file
```

> What I am worried about is inconsistency, though

To be honest, inconsistency is a trivial sacrifice to make until Atom has a formal API for interacting with the filesystem. And yes, you're 100% right. Furthermore, I believe *that* API should be the one consuming the icon-class service... that way, there's only one point of info, and services can deal with that without having to care about where the class is being used.

Alhadis: And no, I think it's a harsh burden that's being placed on package authors, particularly those who have to consider scalability.

spec/file-icons-spec.coffee
Alhadis: @as-cii I'm having an avalanche of issues trying to get these specs to run in `tree-view-spec.coffee` and I don't know why...

Alhadis: Okay, moving it out of the top-level `describe` seemed to do the trick... no idea why, but at least it's in the one file now, as per your request.

lib/file-view.coffee
as-cii: I guess we could hand some metadata about the file to the service as a stopgap measure, like:

```js
service.iconClassForPath(path, {symlink: false})
```

Thoughts?

Alhadis: I'm happy with that. I'll even see if I can tie it in with what #859 uses.

FYI, that IS an addition I took delicate measures to shield against complications.

Alhadis: @as-cii Okay, it's done! :D You'll notice I've passed a [`@stats` property](https://github.com/atom/tree-view/pull/825/files#diff-fe4fb3eaf876e081a80d3f01512c120aR72) from the newly-added `File` method, which will be smoothly taken care of by #859.

The metadata includes atomic values with no dangling references. I only included stuff that was strictly filesystem-related, and not in any way related to the tree-view itself (which I believe was your point in abstracting the consumer's logic from the provider's).

lib/file.coffee
```diff
@@ -67,3 +67,6 @@ class File
 
   isPathEqual: (pathToCompare) ->
     @path is pathToCompare or @realPath is pathToCompare
+
+  # Return a snapshot of the instance's filesystem properties for API consumption
+  getMetadata: -> {@symlink, @realPath, @status, @stats}
```
as-cii: Again, I am a little worried about consistency. In this package we have access to `stats` and `status`, but what about the `tabs` package?

Alhadis: `*cough*` UH, well... at the moment, it's returning a reference to the calling [`TabView`](https://github.com/atom/tabs/blob/master/lib/tab-view.coffee#L215) element, which I'll obviously tend to once we're done with this PR

Access to the tab's `File` instance is available through `@item.buffer.file`, but that's an instance of [one of these](https://atom.io/docs/api/v1.8.0/File), which doesn't have the cool extras of filesystem stats the Tree-View's `File` class has.

Alhadis: I'm aware it's not 100% consistent, but how about storing a reference to the buffer's `File` instance on the second argument? `stats` won't be available, but the source file can be keyed to `bufferFile` or just `file`.

```
tree-view: { symlink, realPath, status, stats }
tabs: { symlink, realPath, sourceFile }
```

To be honest, I think we can probably justify having an extra filesystem call to return the missing info to stay consistent with `tree-view`'s handling. Let's face it: there won't be nearly as many icon-displaying tabs as there will be icon-displaying file listings.

Alhadis:
@as-cii Actually, I'm beginning to think having a `disableCustomIcons` setting probably isn't a good idea. Might be convenient for package authors, but it threatens to make a mess of user configuration. What if a user prefers one package show icons in tabs, and their theme's icons in the tree-view? (Or vice versa...).

There'd be no logical in-between, and package authors would (again) be reduced to using ugly hackery. I think in the long-term, the setting would introduce more complications than it'd solve. My gut feeling says *"no, don't…"* and my guts are usually pretty good about this kinda stuff.

lib/file.coffee
Alhadis: @as-cii Should I file a PR to address the tabs issue?

Alhadis: @as-cii What's happening here? The sudden silence has me rather worried.

as-cii:
> Actually, I'm beginning to think having a disableCustomIcons setting probably isn't a good idea. Might be convenient for package authors, but it threatens to make a mess of user configuration. What if a user prefers one package show icons in tabs, and their theme's icons in the tree-view? (Or vice versa...).

Could you explain why having a setting on each package wouldn't solve this? Like, users could set `tree-view.enableCustomIcons = false` and have `tabs.enableCustomIcons = true`, which seems reasonable to me. What am I missing?

lib/file.coffee
as-cii: > To be honest, I think we can probably justify having an extra filesystem call to return the missing info to stay consistent with tree-view's handling. Let's face it: there won't be nearly as many icon-displaying tabs as there will be icon-displaying file listings.

That's fair, although this feels like a clear indicator that this interface is inconsistent and more complex than it should need to be. I think we should try to design services in an elegant manner, so I am 👎 on passing different arguments based on the package, and I think we should issue `fs.stat` calls directly in the `FileIcons` service, delegating to it all the concerns about determining if a file is symlink and whatnot.

The only reasonable alternative I could think of (that doesn't involve a major redesign in Atom core) is to provide an alternative `iconClassForPath` method that returns a `Promise`, so that the UI thread never blocks and we can do whatever I/O is needed on that function. It would make the calling code a little more complex, but it's probably worth a try if performance-wise the file-icons package becomes a problem.

Alhadis:
Well, a package or theme (like Seti) might offer users a way to disable custom icons specifically to allow other packages to display their icons instead. Seti and File-Icons [have had problems](https://github.com/jesseweed/seti-ui/pull/216) in the past with overlapping icon rules, and this is precisely one of the issues the icon-service exists to solve.

lib/file.coffee
Alhadis: So what's wrong with passing different arguments as a temporary solution until a formal API is developed? When it is, you can just use Grim to deprecate usage of the second argument so package authors know to use the more elegant approach.

Alhadis: Actually never mind. I've lost patience during all this bikeshedding. I'll kill the second argument and write whatever craploads more code I need to write in order to get things off the ground. Seriously.

Alhadis:
I've removed the second parameter, but if I were to suggest a solution to the "disabling icons per package" issue, it'd be to simply pass the package's name as an argument:

```coffee
iconClassForPath(path, "tree-view");
```

That way, packages can decide for themselves whether they want to provide an icon class or not.

Alhadis:
@as-cii I've implemented my last suggestion... hope that's okay with you, mate.

I'll make the same change to the `tabs` package too.

as-cii:
Thanks for the changes, @Alhadis! 🙇 

> Well, a package or theme (like Seti) might offer users a way to disable custom icons specifically to allow other packages to display their icons instead. Seti and File-Icons have had problems in the past with overlapping icon rules, and this is precisely one of the issues the icon-service exists to solve.

I see, so if for instance we have a `enableCustomIcons` setting for tree-view but users have the Seti-UI theme installed they would need to:

1. Set `tree-view.enableCustomIcons = true`.
2. Set `seti-ui.fileIcons = false`

Or, if they want only standard icons for the tree-view, they would need to:

1. Set `tree-view.enableCustomIcons = false`.
2. Set `seti-ui.fileIcons = false`

Is that accurate? If so, it still seems like having a per-package setting provides a high degree of customization while still being pretty easy to change, also allowing new consumers of the file-icons service to pass just a path. Then from the tree-view all we have to do is (pseudocode):

```js
if (atom.config.get('tree-view.enableCustomIcons')) {
  someElement.classList.add(fileIconsService.iconForPath(path))
}
```

I guess you're more concerned about not having the `enableCustomIcons` settings all in one place? I also wonder: do we currently have any usage of disabling custom icons only for only a part of the UI? Seems like it could make the UX between tabs and tree-view inconsistent, right?

That said, I am not totally opposed to pass the context, but seems worth exploring other solutions that make the file-icons service highly reusable as well as keeping the API consistent and clean.

Thanks for your patience on this! If in the meantime you'd like to change this pull-request to introduce only the change you mentioned in the title (i.e. allowing multiple class names to be passed), I think we can merge that right away! 👍

Alhadis:
> …but seems worth exploring other solutions that make the file-icons service highly reusable as well as keeping the API consistent and clean.

The issue here is that you're trying to predict and solve issues on behalf of package authors, instead of entrusting them to tailor their package's logic. There is *no* logical reason to obfuscate the context of the API's call for the sake of keeping things "clean".

Alhadis:
I'll remove the context from this PR temporarily so this bullshit bug can be fixed. But I'll be adding it in a follow-up PR.

PR #826, opened 2016-05-06
rename file from tab

danielmurray:
Addresses feature requests

...atom/atom-11226
https://github.com/atom/atom/issues/11226

... and atom/tabs-251
https://github.com/atom/tabs/issues/251

### DIFF
--- a/menus/tree-view.cson
+++ b/menus/tree-view.cson
@@ -40,7 +40,7 @@
     {'label': 'Copy Project Path', 'command': 'tree-view:copy-project-path'}
     {'label': 'Open In New Window', 'command': 'tree-view:open-in-new-window'}
   ]
-  
+
   '.tree-view.full-menu [is="tree-view-file"]': [
     {'label': 'Split Up', 'command': 'tree-view:open-selected-entry-up'}
     {'label': 'Split Down', 'command': 'tree-view:open-selected-entry-down'}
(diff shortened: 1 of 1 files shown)


danielmurray:
@50Wliu does the `needs-testing` label means I need to write tests, or someone needs to validate that what I implemented does what is says it does?

50Wliu:
Someone needs to test your PR out.  I'll do so tonight!

danielmurray:
Awesome! Looking forward to it!

50Wliu:
Just tested this out and I only see one small problem: attempting to rename an inactive tab doesn't work (eg there's no context menu option for that).  Though I noticed that `Reveal in Tree View` doesn't show up either.

Can you also clarify what the difference is between `.item-views` and `.tab-active`?  I'm guessing the latter is referring to the active tab, but I'm not too sure what the former is.

danielmurray:
The former is the div defining the workspace of the file, however that feature is not defined the aforementioned tickets, so I will remove it.
As for not being able to edit the file name when not active, I figured this feature should be connected to the other tree-view function. If users request it then we can change it.

danielmurray:
\

danielmurray:
PR look good?

50Wliu:
Thanks!

PR #833, opened 2016-05-11
Emit events when directories are toggled

Alhadis:
Packages currently have no reliable way of gauging when a folder's been opened (or closed). One may be tempted to simply attach `click` handlers to each directory entry, but this rules out the possibility of tabbed/keyboard navigation (as well as recursive expansion for subdirectories).

This PR adds two event emissions to the `Directory` class, each for when a directory is collapsed and expanded.

Please advise if you wish for a spec to be included.

### DIFF
--- a/lib/directory.coffee
+++ b/lib/directory.coffee
@@ -54,6 +54,12 @@ class Directory
   onDidRemoveEntries: (callback) ->
     @emitter.on('did-remove-entries', callback)
 
+  onDidCollapse: (callback) ->
+    @emitter.on('did-collapse', callback)
+
+  onDidExpand: (callback) ->
+    @emitter.on('did-expand', callback)
+
   loadRealPath: ->
     if @useSyncFS
       @realPath = fs.realpathSync(@path)
@@ -251,6 +257,7 @@ class Directory
     @expansionState.isExpanded = false
     @expansionState = @serializeExpansionState()
     @unwatch()
+    @emitter.emit('did-collapse')
 
   # Public: Expand this directory, load its children, and start watching it for
   # changes.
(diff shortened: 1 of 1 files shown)


as-cii:
This looks good to me, although I am wondering: is the `Directory` object public? Seems like a private method of this package, which makes me somewhat uncomfortable. Maybe we should expose a service from this package? What am I missing?

Alhadis:
The `Directory` instances are accessible through the `li` element's `directory` property. That's what I'm using to access required metadata.

I'm wondering if it shouldn't be merged with the existing `Directory` class for simplicity's sake...

Alhadis:
To be honest, I feel like a lot of the tree-view's logic should be integrated more seamlessly into Atom's existing APIs. For instance, Atom exposes two public interfaces:

* [File](https://atom.io/docs/api/v1.8.0/File)
* [Directory](https://atom.io/docs/api/v1.8.0/Directory#instance-getEntries)

Meanwhile, the TreeView has two interfaces of the same names, which define completely different properties and methods:

* [File](https://github.com/atom/tree-view/blob/master/lib/file.coffee)
* [Directory](https://github.com/atom/tree-view/blob/master/lib/directory.coffee)

Personally, I'd confine all File-related logic to the File class, and have stuff like `expansionState` defined only by the view-handler... but I'm sure there'd be a more logical reason why such a thing wouldn't be so easy. =)

as-cii:
> The Directory instances are accessible through the li element's directory property. That's what I'm using to access required metadata.

👍

> Personally, I'd confine all File-related logic to the File class, and have stuff like expansionState defined only by the view-handler

Agreed, although for the sake of these changes I think it's okay to keep the duplication between the two objects. Thanks for this @Alhadis, 🚢.

Alhadis:
No, thank YOU! :D

PR #839, opened 2016-05-17
Use getFullExtension instead of path.extname in dialogs

### LINKED ISSUE #838: Provide a more accurate file extension matching through grammars
Recently I've started to refactor a lot of Blade template files, which have extension `.blade.php`. Renaming such files is slightly inconvenient because if I were to rename a file called `somefile.blade.php`, it would automatically select `somefile.blade` for me, I would quickly start typing and then lose the `.blade` suffix -- this forces me to override the suggested selection or re-type it.

I figured it would be rather neat if we could attempt matching the longest file extension from the available grammars and select the base file name based on that first. If there aren't any matches, then revert to old behavior.

Currently I am leaving this here as an idea, but I might eventually submit a pull request, unless someone else wants to do this.

Ingramz:
Closes #838 

If we need a test for this, then I need some guidance regarding how to approach. And also the current test will only work because the file names used for testing already give the full extension using `path.extname`.

### DIFF
--- a/lib/dialog.coffee
+++ b/lib/dialog.coffee
@@ -1,5 +1,6 @@
 {$, TextEditorView, View} = require 'atom-space-pen-views'
 path = require 'path'
+{getFullExtension} = require "./helpers"
 
 module.exports =
 class Dialog extends View
@@ -19,7 +20,7 @@ class Dialog extends View
     @miniEditor.getModel().setText(initialPath)
 
     if select
-      extension = path.extname(initialPath)
+      extension = getFullExtension(initialPath)
       baseName = path.basename(initialPath)
       if baseName is extension
         selectionEnd = initialPath.length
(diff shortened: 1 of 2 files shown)


50Wliu:
For specs, you could create a related spec that tests with a multi-extension file.  For example,
```coffee
it "recognizes multi-extension files", ->
  dotFilePath = path.join(dirPath, "test.file.txt")
  expect(moveDialog).toExist()
  expect(moveDialog.miniEditor.getText()).toBe(atom.project.relativize(dotFilePath))
  expect(moveDialog.miniEditor.getModel().getSelectedText()).toBe "test"
```

Ingramz:
@50Wliu alright, thanks. I'll work on this and also add related tests for `getFullExtension`.

50Wliu:
> and also add related tests for `getFullExtension`.

Don't worry about this unless you really have nothing else to do :).  I added that helper, I should probably be the person writing tests for it.

50Wliu:
@Ingramz are you still interested in finishing this PR?  All it needs is that extra spec.

Ingramz:
@50Wliu I'll probably get to it sometime next week.

PR #847, opened 2016-05-25
Fix bug where duplicate tabs open for a given file

kuychaco:
Fixes https://github.com/atom/atom/issues/11391

In atom/atom#11391 a race condition in opening pending items caused
duplicate items to be opened in the pane. The issue was that the file
was not yet opened in the pane from the first click by the time the second
click triggered a check for the item in the pane. With this commit, the
second call to `atom.workspace.open` is called only after the promise for
the first call is fulfilled and the item has successfully opened in the pane

Signed-off-by: Michelle Tilley <binarymuse@github.com>

### DIFF
--- a/lib/tree-view.coffee
+++ b/lib/tree-view.coffee
@@ -222,10 +222,13 @@ class TreeView extends View
         @currentlyOpening.set(filePath, openPromise)
         openPromise.then => @currentlyOpening.delete(filePath)
     else if detail is 2
-      if promise = @currentlyOpening.get(filePath)
-        promise.then => atom.workspace.open(filePath, searchAllPanes: alwaysOpenExisting)
-      else
-        atom.workspace.open(filePath, searchAllPanes: alwaysOpenExisting)
+      @openAfterPromise(filePath, searchAllPanes: alwaysOpenExisting)
+
+  openAfterPromise: (uri, options) ->
+    if promise = @currentlyOpening.get(uri)
+      promise.then => atom.workspace.open(uri, options)
+    else
+      atom.workspace.open(uri, options)
 
   resizeStarted: =>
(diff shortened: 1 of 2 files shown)


Ben3eeE:
I looked over this and it seems that it only guards when `fileViewEntryClicked` gets called twice caused by double clicking a file in the tree-view. This is only part of the problem reported in https://github.com/atom/atom/issues/11391 that was introduced(discovered?) in Atom 1.7.0.
The original issue was before Atom 1.7.0 (and very stupid and low priority) that was reproduced by first clicking and then pressing enter. This would cause `fileViewEntryClicked` to get called and then `openSelectedEntry` and `openSelectedEntry` has it's own call to `atom.workspace.open` that does not have the same guard?

I'm just guessing I tried to test and had no success. `tree-view` fails to load after I did `apm link --dev` I have no idea what I am doing.

kuychaco:
@Ben3eeE ah yes, good point. Thanks for catching that! Implemented a fix ☝️

WaxMikhail:
I was experiencing this issue in 1.7.3 on OS X  (10.15.1). After upgrading to 1.8 I am still experiencing the issue with just as much frequency.

50Wliu:
@WaxMikhail This issue should be fixed in Atom 1.9.0 I believe.

Ben3eeE:
@50Wliu Latest tree-view is 0.208.0 and was released before this was merged I think so it's not even fixed on master yet? Or am I missing something?

50Wliu:
Ah ok, then it's most likely targeted for 1.10.0.

thelamborghinistory:
any chance of get that into 1.9.0? It's really bugging me, I (and I guess a whole lot of more people) must be continuously aware of not double clicking a file from the first time.

PR #793, opened 2016-04-01
Drag multiple items in tree-view

anderoonies:
this is working toward #232, just one bug that needs to be worked out
it functions correctly, as can be seen: 
tree-multi

PR #851, opened 2016-05-26
Reset drag counts when dropping

anderoonies:
currently, behavior in Atom 1.7.4 looks like this:

bad-drag

after an entry is dropped on a directory, that directory can no longer become `.selected` on `dragEnter` events.

this PR resets the `dragEventCounts` value for that directory so that subsequent `dragEnter` events give it the `.selected` class.

good-drag

i'm having a hard time replicating the broken behavior in specs, which may have something to do with how drag events are being built

### DIFF
--- a/lib/tree-view.coffee
+++ b/lib/tree-view.coffee
@@ -870,6 +870,7 @@ class TreeView extends View
     return unless entry instanceof DirectoryView
 
     entry.classList.remove('selected')
+    @dragEventCounts.set(entry, 0)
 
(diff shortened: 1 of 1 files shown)


Ben3eeE:
Noticed this bug when testing https://github.com/atom/tree-view/pull/793. Tested this PR and it fixes the problem for me.

50Wliu:
@Ben3eeE was this fixed by #1179?

Ben3eeE:
@50wliu i think so. I don't see this bug any longer.

PR #859, opened 2016-06-03
Store filesystem stats on tree-view entries

Alhadis:
Developers may need to access more granular data when traversing the tree-view's contents... a file's modification time, its filesize, permissions, and so forth. All of this data is already available to the tree-view as it's traversing the filesystem, but only a narrow subset of it is exposed to the `File` and `Directory` instances it creates. This PR amends the constructor to include the "pure" filesystem stats on an instanced property, `.stats`.

Implementation
---------------------
Rather than passing a direct reference to the value returned by `lstatSyncNoException`, I've flattened its instanced properties into a new object, cutting out any filesystem methods. This was done out of a (perhaps excessive) concern for performance - I didn't want to keep anything from being garbage collected, and I'm unsure how Node goes about maintaining filesystem connections under the hood.

Performance benefits
-----------------------------
Aside from convenience, a boost to performance is gained by circumventing the need for excess filesystem calls. For instance, I used the event emission callback pending in #833 to run a simplistic benchmark:
<table><tr><th>Without <code>.stats</code></th><th>With <code>.stats</code></th></tr/><tr valign="top"><td><pre><code>start = window.performance.now()

statCache = {}
grabStats = (from) ->
 {path} = from
 stat = fs.lstatSyncNoException(path)
 symlink = stat.isSymbolicLink?()
 if symlink
  stat = fs.statSyncNoException(path)
 statCache[path] = stat

grabStats dir
for path, file of dir.entries
 grabStats file

end = window.performance.now()
"Time: #{end - start}"
</code></pre></td><td><pre><code>start = window.performance.now()

statCache = {}
grabStats = (from) ->
 statCache[path] = from.stats

grabStats dir
for path, file of dir.entries
 grabStats file

end = window.performance.now()
"Time: #{end - start}"</code></pre></td></tr><tr align="center"><td><b>Time:</b> 3.5</td><td><b>Time:</b> 0.37</td></tr></table>
The times shown above are averages (trailing digits clipped to a sane length), tested against a folder holding 632 files ([most of them blank](https://github.com/DanBrooker/file-icons/tree/v2/examples)). This was run on fairly modern hardware (MacBook Pro / Mavericks 10.9.5 / Processor: 2.3 GHz Intel Core i7 / Memory: 16 GBs 1600 MHz DDR3). While the reduction shown above isn't exactly thrilling, scalability should always be considered.


Alhadis:
@50Wliu Spare a little advice? I noticed Travis failed the build, but that's the same odd error I got at sporadic intervals when running the package specs. It only appeared after reloading the package the first time... whenever I ran it after that, everything passed.

Have you ever encountered anything like that before when running specs...?

50Wliu:
I haven't seen that spec fail before, but it also seems like a recent addition to the suite.

Alhadis:
S'alright. =) I'll see if I can work it out tomorrow.

Alhadis:
Okay, I *think* I managed to fix the failing specs. Based on [the notes](https://github.com/atom/tree-view/blob/master/spec/tree-view-spec.coffee#L568-L571) @BinaryMuse left in d928555, I deduced the failure might've been connected to the use of `.click()` to open the directory in [one of the tests](https://github.com/atom/tree-view/pull/859/commits/8bde2883e19345002caf893eae9406051f7074b7#diff-ddd2fe70381e781d323bc21251794b7aR3359)). I used the `.expand()[0]` method instead, but because of the error's unpredictable nature, I've no way of knowing if this actually fixed it for good.

lib/directory.coffee
```diff
@@ -177,6 +177,9 @@ class Directory
       stat = fs.lstatSyncNoException(fullPath)
       symlink = stat.isSymbolicLink?()
       stat = fs.statSyncNoException(fullPath) if symlink
+      statFlat = _.pick stat, _.keys(stat)...
+      for i in ["atime", "birthtime", "ctime", "mtime"]
```
thomasjo: Minor this, but this is not an index, so I think the `i` should be changed to `key`, `property`, `type`, or some other colloquialism that is more appropriate than `i`.

Alhadis: Changed.

(FWIW, I usually mean `iteration` when I write `i`).

Alhadis:
@BinaryMuse Would you be able to explain why the build might be failing? It's stopping at a test you added:
```
TreeView
  when the package first activates and there is a file open (regression)
    when the file is pending
      it marks the pending file as permanent
        Expected null to be <TextEditor 176>. (spec/tree-view-spec.coffee:594:67)
```

This was working before, and I added a null-coalescing operator to prevent breakage for undefined properties (which is entirely unrelated). The build's suddenly failing again, so I'm not sure what might be causing this.

Alhadis:
[Another example of a package that could benefit from this](https://atom.io/packages/tree-view-finder).

Alhadis:
@50Wliu AppVeyor builds appear to be failing all over the place over the last couple days...

```
Error processing argument at index 9, conversion failure from 
TypeError: Error processing argument at index 9, conversion failure from 
    at TypeError (native)
    at Object.module.exports.showMessageBox (C:\projects\tree-view\Atom\resources\atom.asar\browser\api\lib\dialog.js:155:20)
    at EventEmitter.<anonymous> (C:\projects\tree-view\Atom\resources\app.asar\src\browser\atom-window.js:221:27)
    at emitOne (events.js:82:20)
    at EventEmitter.emit (events.js:169:7)
Specs Failed
```

**Other failures**
- [#889]: https://ci.appveyor.com/project/Atom/tree-view/build/32/job/rfewc9wey1tgbxr9
- [#833]: https://ci.appveyor.com/project/Atom/tree-view/build/35/job/heeo372jynfy3d2s
- [#825]: https://ci.appveyor.com/project/Atom/tree-view/build/33/job/rkkjm0db5t1d4i7o

That last one also has a tonne of build errors, none of which appear to be related. :S This is clearly an issue with AppVeyor, but I'm not sure where to begin reporting it... so I'm pinging you (please let me know if there's a better place to bring this up if it happens again in future).

50Wliu:
See #791.  tree-view specs currently do not run on Windows yet, so AppVeyor can be safely ignored.

In addition that other failing spec you mentioned on June 6 appears to be flaky even on master.

Alhadis:
Ah right, thanks! That's good to know. =)

> In addition that other failing spec you mentioned on June 6 appears to be flaky even on master.

So we're dealing with a Heisenbug? Great... :S :\

Alhadis:
@as-cii Nagging you because I feel you're the one most likely to respond first.

as-cii:
Thanks, @Alhadis!

Alhadis:
NO, THANK YOU.

Issue #248, opened 2014-10-25
right click on a file can open the directory right-click menu instead


octosquid:
When you right click onto the blank space at the left, or slightly above or below the text, the directory right-click menu opens instead:
![s1](https://cloud.githubusercontent.com/assets/4958615/4778967/2f6fd558-5c0f-11e4-86d9-b833c047ebe4.png)

Normal behaviour is seen when rightclicking onto text:
![s2](https://cloud.githubusercontent.com/assets/4958615/4778968/3522e648-5c0f-11e4-82c2-39ba983a930c.png)

Issue on atom core: atom/atom#3960

PR #861, opened 2016-06-06
Move title attribute to parent element

esdoppio:
Helps fixing #248 by exposing file/dir name to the "clickable area". This still makes sense semantically; if `title` isn't present on an element, the `title` of that element's parent  is relevant for it.

Now a tooltip emerges when hovering over anywhere in the "clickable area" of an entry rather than just the "name area". Might be a desirable side-effect.

### DIFF
--- a/lib/directory-view.coffee
+++ b/lib/directory-view.coffee
@@ -13,6 +13,7 @@ class DirectoryView extends HTMLElement
 
     @header = document.createElement('div')
     @header.classList.add('header', 'list-item')
+    @header.title = @directory.name
 
     @directoryName = document.createElement('span')
     @directoryName.classList.add('name', 'icon')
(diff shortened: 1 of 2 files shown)


Alhadis:
This will break anything that uses the `@directoryName.title` element to get the title attribute.

esdoppio:
Thought about it; there shouldn't be many things relying on this, considering the possible uses of the output of the tree-view—mostly in the context menu I think.

And mostly file/dir names are not needed in the context menu; if they're needed, it probably means trouble, as in https://github.com/atom/markdown-preview/issues/356

@Alhadis Could you give me an exp? I will change this if it really breaks anything.

Alhadis:
Off the top of my head, no. However, it wouldn't be unusual for somebody to be applying icon-related styling to some files in the tree view using a `[title]` selector. A similar approach (but using a different attribute) is described in the [File-Icons](https://github.com/DanBrooker/file-icons#customisation) package.

So while this isn't going to break much, it isn't really adding much either.

PR #889, opened 2016-07-15
Change 'Show in Explorer' to open location in default file manager

victusmortuus:
I've changed 'Show in Explorer' handler to show selected item in default file manager instead of 'explorer.exe'. It is more friendly way to handle this. It works like ShellExecute in Windows. Tested on Explorer and Directory Opus.

### DIFF
--- a/lib/tree-view.coffee
+++ b/lib/tree-view.coffee
@@ -487,30 +487,12 @@ class TreeView extends View
     entry = @selectedEntry()
     return unless entry
 
-    isFile = entry instanceof FileView
-    {command, args, label} = @fileManagerCommandForPath(entry.getPath(), isFile)
-
-    handleError = (errorMessage) ->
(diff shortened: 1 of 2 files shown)


lib/tree-view.coffee
```diff
@@ -487,30 +487,12 @@ class TreeView extends View
...
-    showProcess = new BufferedProcess({command, args, stderr, exit})
-    showProcess.onWillThrowError ({error, handle}) ->
-      handle()
-      handleError(error?.message)
+    path = entry.getPath()
+    if path
+      # open in default file manager (Directory Opus, Explorer, Finder, etc.)
+      shell.showItemInFolder(path)
+    else
+      console.warn('Path not found!')
```
victusmortuus: If tree-view show only files on disk this should never happen. The only situation is if in-memory file buffer will be visible in tree view.

Issue #862, opened 2016-06-06
Flickering test?


esdoppio:
https://github.com/atom/tree-view/blob/master/spec/tree-view-spec.coffee#L593-L594

These two assertions fail once in a while against the same production code.

nathansobo:
Closed via #900.

PR #900, opened 2016-07-27
Hopefully fix intermittent test failures

nathansobo:
In the modified test, the click events create a promise, which we were previously not waiting to be resolved before moving on to the next test. I hypothesize that in some cases, the promise would end up resolving in the middle of the *next* test, hence polluting it. It's hard to prove a negative, but I'm going to try merging this and seeing if our situation improves over the next few builds.

/cc @kuychaco @BinaryMuse


Ben3eeE:
https://github.com/atom/tree-view/issues/862?

nathansobo:
Thanks for the heads up

PR #905, opened 2016-08-04
add new open-selected-entry-without-activate command

t9md:
# Add new command:

 Name: `tree-view:open-selected-entry-without-activate`
 Scope: `.tree-view`
 defaultKeymap: `o`

# What's difference from normal `open-selected-entry`?

- `tree-view:open-selected-entry-without-activate` don't change focus to newly opened item, focus remains on treeView so that I can continue to open next item by `o` for quick previewing.

- Item is opened with `pending` mode.

# Motivation

I want to quickly preview multiple file, so don't want to loose focus of treeView after opening item.
I can do this by single clicking item with mouse, but I want to do this with **keyboard navigation**.
With this new feature and keybinding, user can do `jojojo` to quickly preview three files sequentially.

open-selected-item-without-activate

# Name of command

Which is better? not sure
`tree-view:open-selected-entry-without-activate` vs `tree-view:preview`

### DIFF
--- a/keymaps/tree-view.cson
+++ b/keymaps/tree-view.cson
@@ -68,6 +68,7 @@
   'alt-left': 'tree-view:recursive-collapse-directory'
   'h': 'tree-view:collapse-directory'
   'enter': 'tree-view:open-selected-entry'
+  'o': 'tree-view:open-selected-entry-without-activate'
   'ctrl-C': 'tree-view:copy-full-path'
   'm': 'tree-view:move'
   'f2': 'tree-view:move'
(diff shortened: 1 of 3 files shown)


calebmeyer:
I would vote for `tree-view:preview` or `tree-view:open-selected-entry-keep-focus`

t9md:
I noticed there was similar issue in the past https://github.com/atom/tree-view/pull/413.
In spite of [this](https://github.com/atom/tree-view/pull/413#issuecomment-175812901) comment.
It hasn't implemented yet.
The 'right' key activate newly opened pane original request of #413 is to set false for activatePane arg.
So this issue is duplicate request but with updated version.

Please take a look, this is long lived and common feature request which is not resolved yet.

PR #909, opened 2016-08-05
Declare calling context for icon-service providers

Alhadis:
Following on from #825 after @as-cii [ordered me](https://github.com/atom/tree-view/pull/825#issuecomment-237788504) to remove an auxiliary addition in order to merge a bug-fix.

This PR (along with atom/fuzzy-finder#249 and atom/tabs#359) offers a solution to a shortcoming in the icon-service API. Currently, package authors have no way of determining where/how an icon is being displayed. This limitation makes granular adjustments to icon-classes impossible without resorting to hideous CSS hackery.

**Obvious example of why this might be an issue:**
A package might have a setting to control whether custom icons are shown in tabs or not. @as-cii suggested having a per-package setting called `disableCustomIcons` that would suppress service calls, but I don't believe that's solving the full issue.

**Less obvious example:**
A package might need to include (or remove) certain CSS classes depending on where an icon is being shown. Real-world example: the File Icons package has a *"Colour only on changes"* setting that adjusts tree-view entry colours based on whether the file's been modified:

Figure 1

This serves only to mark modifications in the project's scope, so the colour should be confined to the tree-view only. Obviously, this would require a package to understand the context of an API's call.

### DIFF
--- a/lib/file-view.coffee
+++ b/lib/file-view.coffee
@@ -19,7 +19,7 @@ class FileView extends HTMLElement
     @fileName.dataset.name = @file.name
     @fileName.dataset.path = @file.path
 
-    iconClass = FileIcons.getService().iconClassForPath(@file.path)
+    iconClass = FileIcons.getService().iconClassForPath(@file.path, "tree-view")
     if iconClass
       unless Array.isArray iconClass
         iconClass = iconClass.toString().split(/\s+/g)
--- a/spec/tree-view-spec.coffee
+++ b/spec/tree-view-spec.coffee
@@ -3349,7 +3349,8 @@ describe 'Icon class handling', ->
     jasmine.attachToDOM(workspaceElement)
 
     FileIcons.setService
-      iconClassForPath: (path) ->
+      iconClassForPath: (path, context) ->
+        expect(context).toBe "tree-view"
         [name, id] = path.match(/file-(\d+)\.txt$/)
         switch id
           when "1" then 'first second'


Alhadis:
@as-cii I look forward to hearing your thoughts on this.

as-cii:
Overall I think this might be okay, and I guess how to implement this feature is ultimately a decision that should be taken in the `file-icons` package. Still I'd ❤️ for Atom core packages to depend on stable and solid APIs that don't change very often, so it'd be nice to hear your thoughts about an hypothetical alternative implementation where the calling package is responsible for telling whether or not the icon can have colors or not. For instance:

```js
// in tree-view
iconsService.iconForPath(path, {supportsColors: true})

// in tabs
iconsService.iconForPath(path, {supportsColors: false})
```

Seems like the less abstract dependency should depend on the more abstract one, and by leaking the caller name we're introducing a small dependency in the opposite direction.

However, if you feel strongly about using a string identifier for the calling package, I think I am okay with that as well.

Alhadis:
No, let's keep it short and simple, please... a simple string identifier is honest, transparent and minimal, but pretty much includes everything a package author will need to know to tune the result.

> and by leaking the caller name we're introducing a small dependency in the opposite direction.

Normally I'd never propose a solution like this. Dependencies should always be unidirectional whenever possible. This is different, though, because CSS and styling are involved. The potential criteria a design might require are limited only by the creator's imagination. Maybe some developer wants the icons scaled a different size in the tab-bar for a theme, and slightly smaller in the tree-view? Who knows what their design looks like in Photoshop, and whether they'll want that controlled conditionally? Point is, we can't assume designers will only ever want the obvious.

Through all the years I've worked as a front-end web developer, I can assure you 90% of unmaintainable code wouldn't exist if the original author had taken a pragmatic approach instead of what they perceived to be the "right" one at the time. I'll spare you my rambling, but I'll just say this: people who believe it's wrong to style an element using its ID are wrong and should be headbutted. =)

as-cii:
> Maybe some developer wants the icons scaled a different size in the tab-bar for a theme, and slightly smaller in the tree-view? Who knows what their design looks like in Photoshop, and whether they'll want that controlled conditionally? Point is, we can't assume designers will only ever want the obvious.

I think that's a fair point, I guess what I am proposing (as for `enableCustomIcons` settings) is to allow customization in the package itself rather than in file-icons, keeping that service interface high-level. 

Anyways, let's go ahead and use a string, I tend to agree it might be the most pragmatic choice for now.

Alhadis:
Words can't express what I'm feeling right now, so I'll go with this:



**GRAZIE!**

PR #913, opened 2016-08-07
Use colors to indicate git status

pbiggar:
Currently, the coloring of status in tree-view makes it very difficult to understand the git status of files. For example, changed files (to use the git meaning of the term) are shown the same as updated files, while added files are shown differently.

For example, for a repo in this state:
screenshot 2016-08-07 12 39 27

This is what tree-view tells me:
screenshot 2016-08-07 12 41 14

This is because tree-view doesn't differentiate between whether a file is in the git index or not. It's best shown in the current code:

```coffee
     if repo.isPathIgnored(@path)
      newStatus = 'ignored'
     else
      status = repo.getCachedPathStatus(@path)
      if repo.isStatusModified(status)
        newStatus = 'modified'
      else if repo.isStatusNew(status)
        newStatus = 'added'
```

Basically, the status can be status-ignored, status-modified, or status-added.

This isn't particularly useful if you use git, as I'm sure you all do. I find myself going to the command line frequently to figure out the actual status of my files. In particular, I find it very hard to make partial commits (dissecting my changes into multiple logical commits) as I constantly have to switch into the command line and back.

This is particularly annoying given that tree-view is one of the areas pointed out as "git integration": http://blog.atom.io/2014/03/13/git-integration.html. I thought perhaps that status-modified and status-added had different meanings, but it's clear (in the blog post and the code) that they are supposed to represent git.

I initially tried to make this a package, but none of the APIs were exposed and it involved a lot of fighting (races and CSS specificity mostly), and this felt like this was a bug in tree-view as opposed to a nice extra, hence making this change here.

OK, done with the justification, onto the changes.

For the same repo as above, here's what this looks like:
screenshot 2016-08-07 12 51 30

As you can see, while you can't tell anything in the current version of tree-view, with these changes you can really see the git status quite well.

Instead of taking a distillation of the the git status as it currently stands, these changes expose the actual git statuses, taking into account that some changes are in the index and some are not. It takes the git status code, analyses it to break the changes into 5 major statuses, which mirror git config's names for them (see color.status.<slot> in https://git-scm.com/docs/git-config for discussion):

- status-added
- status-updated
- status-changed
- status-unmerged
- status-untracked

And then some extras (which also come from the git status code):
- status-conflicted
- status-unmodified
- status-ignored

I also added styling (well, colors) for the statuses. I use the colors from `@ui-variables` - is that portable across themes? Styling currently seem to be handled by themes, but I added defaults so that we can ship this and let the themes catch up (we can remove the colors later if needs be). 

Note that files can be in two statuses at the same time. For example if you `git add` a file, then edit it afterwards, then the file will be both status-updated and status-changed. In the git CLI, this is displayed as a green M followed by a red M. To try and represent this here, I used a gradient. It looks quite pretty 😄. 

For directories, I cut it down to status-ignored, status-changed, status-conflicted, status-updated, status-unmodified. I'd appreciate feedback on this (well on anything, but esp on this :)).

I've fixed existing tests (except one, I'll be working on that soon). I also have more tests to write. Would love suggestions on what things people worry about here.

I'm not sure about status-conflicted. AFAICT, when a file is conflicted, it will have other statuses to indicate how the conflict happened. I choose to use status-conflicted instead of trying to represent those statuses, which I think makes the most sense. This bubbles up to directories: a directory is conflicted if any files in it are. However, I use the same color for conflicted (red) as for untracked, so maybe I should add italics or something. Suggestions welcome.

Ideally, some of this code should be moved to git-utils. git-utils doesn't expose much right now. That might be a nice followup, but it feels like a bad idea to try and make the change in git-utils first, without a downstream consumer, and it seems easier and more productive to get this into tree-view first. That's why I didn't refactor the git status code too much. Happy to revisit this if you folks don't like. (I will say that after many years of contributing code to multi-repo projects, I feel there's a real risk of being asked to make the change upstream first, and then being told that there's no consumers yet, so make it downstream. I would really like to avoid this 😄) 

There is some eventing code that will now receive an array instead of a status. It's undocumented so I presume that not much uses it (I tried to use it in my package and it seemed flaky, but hard to tell). It doesn't break any of the other git tools I use, but if there's a better way to address this, I'm all ears.

For some reason, status-unmodified is never reported. I'm looking into that now, but since I don't style it, it makes no functional difference for now.

~~Currently, the selected file (in one-dark-ui at least) doesn't show the status color. I think we should change that, but I'm not sure how. Punting for now.~~ NVM, this seems to work fine :)

OK, I think that's everything. Would love as much feedback as possible before I invest a lot of time in testing it :)

### DIFF
--- a/lib/directory-view.coffee
+++ b/lib/directory-view.coffee
@@ -57,8 +57,13 @@ class DirectoryView extends HTMLElement
     @expand() if @directory.expansionState.isExpanded
 
   updateStatus: ->
-    @classList.remove('status-ignored', 'status-modified', 'status-added')
-    @classList.add("status-#{@directory.status}") if @directory.status?
+    @classList.remove(
+      'status-ignored', 'status-changed', 'status-conflicted',
+      'status-updated', 'status-unmodified')
+
+    if @directory.statuses?
+      @classList.add.apply(@classList, @directory.statuses.map (s) -> "status-#{s}")
+
 
   subscribeToDirectory: ->
     @subscriptions.add @directory.onDidAddEntries (addedEntries) =>
--- a/lib/directory.coffee
+++ b/lib/directory.coffee
@@ -27,7 +27,7 @@ class Directory
     @expansionState ?= {}
     @expansionState.isExpanded ?= false
     @expansionState.entries ?= {}
-    @status = null
+    @statuses = []
     @entries = {}
 
     @submodule = repoForPath(@path)?.isSubmodule(@path)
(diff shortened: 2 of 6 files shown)


pbiggar:
Hey, just pinging on this. I have some more improvements to tree-view, but want to make sure this project actually takes PRs before I do.

50Wliu:
Yes it does.  We have a large backlog of PRs so there's large fluctuations when it comes to how fast each PR is reviewed.

pbiggar:
@50Wliu Ah yes, indeed. Can I help? I think I've got the whole codebase down, so happy to help review or triage or whatever.

pbiggar:
@50Wliu Do you know what the backlog is? I was hoping to contribute a lot to atom but this isn't a great start 😞

lee-dohm:
@pbiggar Yes, we accept PRs. We currently have a huge backlog of them to deal with because two of our main team members (and we only have eight total) were out on parental leave at the same time earlier this year. We're still trying to catch up.

This is on the list of PRs to review (see the `needs-review` tag) and we're working on a new system to get them through the pipe quicker.

I'll see if someone on the team can pick this up. It is a fairly complex change though and would have far-reaching implications for all of Atom, not just tree-view. It isn't something that would get accepted without a lot of discussion.

pbiggar:
@lee-dohm Understood about the backlog. Thanks for getting back to me. Happy to help clear the backlog by doing reviews if useful. I know how much of a burden they can be, and I think contributing to OSS means doing reviews too 😄 

Regarding the complexity of the change, agreed. It was hard to tell whether this would be a welcome change and I understand if it isn't (I would argue that it should be, but I haven't seen the product roadmap, if there is one 😃).

lee-dohm:
I talked this over with the dev team and this is more complexity than we want to take on for the default git status support visualization. Keep in mind that there are aspects to that visualization spread throughout Atom, including [tree-view](https://github.com/atom/tree-view), [tabs](https://github.com/atom/tabs), the gutter, [atom-ui](https://github.com/atom/atom-ui), the [styleguide](https://github.com/atom/styleguide) and probably some other places I'm forgetting.

What we can offer is the dev team would be willing to working with you on a [Behavior Service](https://github.com/atom/atom/issues/9820) similar to the [file icons service](https://github.com/search?utf8=%E2%9C%93&q=user%3Aatom+is%3Apr+file+icon+service) that was added to allow packages to customize the icons used to represent individual files. The default behavior would be the same, but people could install a package that would hook in to the git-status behavior service and provide more detailed git status like you're suggesting.

How does that sound?

pbiggar:
Hey Lee, that's a great solution. I had a lot of problems implementing this as a plugin, so I'm very happy with the outcome where it will be possible. After all, even if I believe this behaviour should be default, the best way to achieve that is to make plugins that so many people love that it's obvious the plugin should be included in the default atom 😄  (This also gives me time to complete the functionality, which I think was enough to ship, but maybe not enough to be so obviously better that it definitely should be shipped).

I read the behaviour service proposal, and that seems like an excellent implementation. Very emacs-y. How do we get started?

And thanks again, this is great.

lee-dohm:
> I read the behaviour service proposal, and that seems like an excellent implementation. Very emacs-y.

Thanks!

> How do we get started?

If you could open an Issue with a description of the API that you think would work best, I'll ensure that the dev team assigns someone to review it. Once we're all comfortable with the plan, we can get the PRs going. I haven't taken too deep a look at the PR here, but it sounds like it would be something fairly similar to the file-icons service ... essentially handing back a CSS class or classes to decorate the element based on the git status of whatever the element represents.

When you open the Issue, please @mention me on it and we'll go from there :grinning:

Thanks for contributing!

PR #916, opened 2016-08-08
Add "Sort By Case" option

TrevorBurnham:
Re: https://github.com/atom/tree-view/issues/235, this PR adds an option to put uppercase file/dir names above lowercase ones, consistent with the GitHub web UI:

tree-view-sort

Right now it only affects the relative sort order between characters in the `A-Z` and `a-z` ranges. It's probably desirable to extend that to a wider set of uppercase/lowercase characters (e.g. so that `ÃLPHA` comes before `omega`), but that would require some [very lengthy regexes](http://apps.timwhitlock.info/js/regex#). For the initial PR, I wanted to keep things simple.

### DIFF
--- a/lib/directory.coffee
+++ b/lib/directory.coffee
@@ -171,7 +171,12 @@ class Directory
       names = fs.readdirSync(@path)
     catch error
       names = []
-    names.sort(new Intl.Collator(undefined, {numeric: true, sensitivity: "base"}).compare)
+
+    localizedCompare = new Intl.Collator(undefined, {numeric: true, sensitivity: "base"}).compare
+    if atom.config.get('tree-view.sortByCase')
+      @sortByCase(names, localizedCompare)
+    else
+      names.sort(localizedCompare)
 
     files = []
     directories = []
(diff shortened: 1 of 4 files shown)


50Wliu:
> consistent with the GitHub web UI

Huh.  Never even noticed that.

Some lint errors that you might want to check out, but otherwise this looks pretty nice.

TrevorBurnham:
@50Wliu Ah, sorry for the lints! I've fixed those and rebased my branch to keep things tidy.

dcalhoun:
@50Wliu if I work to get an up-to-date version of his PR with a passing build, is this something that you'd merge? Are there other thing blocking this?

lee-dohm:
Thanks for the ping @dcalhoun.

In the intervening time, we've decided that we aren't going to accept PRs for more customized sort orders in the Tree View. Thanks very much for your interest and your excitement about Atom.

dcalhoun:
@lee-dohm Alright. Seems odd, given that it diverges from the GitHub.com behavior, but thank you for the swift response regardless.

PR #926, opened 2016-08-23
Guard against undefined path

thomasjo:
Passing in `undefined` to `path` module functions results is no longer supported as of Node v6. This PR is part of the effort to upgrade Atom to run on the latest release of Electron (atom/atom#12300).


maxbrunsfeld:
❤️  Thanks @thomasjo!

Issue #927, opened 2016-08-25
Uncaught TypeError: Cannot read property 'getTime' of undefined


adam31:
[Enter steps to reproduce below:]

1. clone atom from https://github.com/atom/atom.git
2. build atom with script/build
3. start atom
4. From menu select Atom->About Atom
5. Observe "Uncaught TypeError" dialog.

**Atom Version**: 1.11.0-dev-bcc0aba
**Electron Version**: 0.37.8
**System**: Mac OS X 10.11
**Thrown From**: [tree-view](https://github.com/atom/tree-view) package, v0.209.0

### Stack Trace

Uncaught TypeError: Cannot read property 'getTime' of undefined

```
At /Applications/Atom.app/Contents/Resources/app.asar/node_modules/tree-view/lib/tree-view.js:599

TypeError: Cannot read property 'getTime' of undefined
    at TreeView.<anonymous> (/Applications/Atom.app/Contents/Resources/app.asar/node_modules/tree-view/lib/tree-view.js:599:36)
    at TreeView.module.exports.TreeView.updateRoots (/Applications/Atom.app/Contents/Resources/app.asar/node_modules/tree-view/lib/tree-view.js:619:10)
    at /Applications/Atom.app/Contents/Resources/app.asar/node_modules/tree-view/lib/tree-view.js:369:24
    at Function.module.exports.Emitter.simpleDispatch (/Applications/Atom.app/Contents/Resources/app.asar/node_modules/event-kit/lib/emitter.js:25:14)
    at Emitter.module.exports.Emitter.emit (/Applications/Atom.app/Contents/Resources/app.asar/node_modules/event-kit/lib/emitter.js:129:28)
    at Project.module.exports.Project.addPath (/Applications/Atom.app/Contents/Resources/app.asar/src/project.js:270:29)
    at AtomEnvironment.module.exports.AtomEnvironment.openLocations (/Applications/Atom.app/Contents/Resources/app.asar/src/atom-environment.js:1147:26)
    at EventEmitter.outerCallback (/Applications/Atom.app/Contents/Resources/app.asar/src/application-delegate.js:258:18)
    at emitThree (events.js:115:20)
    at EventEmitter.emit (events.js:188:7)
```

### Commands

```
```

### Config

```json
{}
```

### Installed Packages

```coffee
# User
atom-dark-syntax, v0.27.0 (inactive)
atom-dark-ui, v0.52.0 (inactive)
atom-light-syntax, v0.28.0 (inactive)
[...]

PR #930, opened 2016-08-27
Avoid breaking when calling lstat on virtual paths

Alhadis:
This should fix atom/tree-view#927.

To bottom-line it: if you open Atom's `about` view in a blank workspace window, it adds the `atom://about/` opener URI to the list of project paths. I didn't even know it was possible to add a project that wasn't a real filesystem path, so I didn't factor that in when working on #859.

Anyway, if a spec is needed for this, please tell me exactly where it should go (in which spec-file, suite, etc).

### DIFF
--- a/lib/tree-view.coffee
+++ b/lib/tree-view.coffee
@@ -276,6 +276,7 @@ class TreeView extends View
 
     @roots = for projectPath in atom.project.getPaths()
       stats = fs.lstatSyncNoException(projectPath)
+      continue unless stats
       stats = _.pick stats, _.keys(stats)...
       for key in ["atime", "birthtime", "ctime", "mtime"]
         stats[key] = stats[key].getTime()
--- a/lib/tree-view.coffee
+++ b/lib/tree-view.coffee
@@ -275,8 +275,7 @@ class TreeView extends View
     @loadIgnoredPatterns()
 
     @roots = for projectPath in atom.project.getPaths()
-      stats = fs.lstatSyncNoException(projectPath)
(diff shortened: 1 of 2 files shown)


50Wliu:
I can confirm that the exception no longer appears.  I would just put the spec in `file-stats-spec.coffee`: `it 'does not throw when accessing virtual filepaths'` or something like that.

Alhadis:
If you're curious why this is taking so long, it's because I'm trying to write a spec that won't give you a false positive. =)

So far, zero luck, and this is incredibly frustrating. 👍

Alhadis:
Okay, done. Issue was that `atom.workspace.open("atom://about")` wasn't triggering the path-change event that causes the exception to occur.

I had to explicitly call `atom.project.setPaths(["atom://about"])` instead, which had the fortunate side-effect of failing instantly (as opposed to asynchronously, which might've been adding to my frustration with Jasmine).

lib/tree-view.coffee
```diff
@@ -276,6 +276,7 @@ class TreeView extends View
 
     @roots = for projectPath in atom.project.getPaths()
       stats = fs.lstatSyncNoException(projectPath)
+      continue unless stats
```
50Wliu: Nitpick: this can be `continue unless stats = fs....`, right?

Alhadis: This was easier to comment out to verify the specs were trustworthy. :p But okay.

50Wliu:
Thanks!

PR #938, opened 2016-09-07
Trigger events when moving, creating or deleting files/directories

Alhadis:
There's currently no reliable way to detect when a user's renamed or moved a file in the tree-view. Filesystem events such as `did-destroy` are by no means fit for detecting a file's deletion; such events are also triggered on `File` objects when dragging `FileView` objects between folders.

Furthermore, it's difficult to determine a file or directory's previous path when responding to a path change. Short of individually monitoring every file and directory's status, package authors have no easy way to respond to user-directed changes in the filesystem.

This PR addresses this matter by introducing a number of events that're emitted at certain points in the `tree-view`'s lifecycle:

* `directory-created`
	- `path`: Resolved/normalised path of the directory
* `entry-copied`
	- `oldPath`: Path to source file/directory
	- `newPath`: Path of newly-created file/directory
* `entry-deleted`
	- `path`: Last-known path of a deleted resource
* `entry-moved`: Self-explanatory. Invoked by both cut-and-paste as well as drag-and-drop/rename.
	- `oldPath`
	- `newPath`
* `file-created`
	- `path`: Resolved/normalised path of the file

These events are emitted on the main `tree-view` instance, which is accessible through both the DOM as well as via `atom.packages.activePackages["tree-view"].mainModule.treeView`.

### DIFF
--- a/lib/copy-dialog.coffee
+++ b/lib/copy-dialog.coffee
@@ -32,8 +32,10 @@ class CopyDialog extends Dialog
     try
       if fs.isDirectorySync(@initialPath)
         fs.copySync(@initialPath, newPath)
+        @trigger 'entry-copied', [@initialPath, newPath]
       else
-        fs.copy @initialPath, newPath, ->
+        fs.copy @initialPath, newPath, =>
+          @trigger 'entry-copied', [@initialPath, newPath]
           atom.workspace.open newPath,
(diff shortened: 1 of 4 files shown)


Alhadis:
Bloody hell, I do write some crap, don't I.

Issue #430, opened 2015-04-19
Feature: Colorization of staged files


alanzanattadev:
Hi,

That would be great to have staged files (after a git add) colored in another color (blue ? cyan ? red ?) and to be able to right click and select a "Git Add" option to stage the file. Because it's painful having to switch to a terminal just to add files and to check status. I checked sources of tree view to have an idea on how it is implemented, and I saw that Atom API doesn't have anything to check if a file is staged or not.. I could do a pull request to add the feature in tree-view with nodegit but I think you prefer to follow the Atom API.

Please ! :)

Thank you !

pbiggar:
I agree. In tree-view, a modified file/directory will show up identically regardless of whether it is modified in the working directory, or in the index. That is, the file will appear orange in both cases.

This is quite different from the expectation I have from git. In the git cli, modified files show as a red Ms, and modified-but-in-the-index show as green Ms.

I believe a modified-but-in-the-index file should be green in tree-view, instead.

I propose the following change:

- files which have a red M in the git cli should be orange in tree-view
- files which have a green M in the git cli should be green in tree-view

I believe this should be straightforward to do for just modified files (see references below), and I volunteer to do it. Would this change be accepted? I would probably follow up by filling out the rest of the possible statuses, once the first change was merged.

Reference:

- the statuses are [here in libgit2] (https://github.com/libgit2/libgit2/blob/HEAD/include/git2/status.h#L32-50)
- the status are [here in git-utils](https://github.com/atom/git-utils/blob/master/src/git.coffee)
- the statuses are converted into CSS classes in [file.coffee](https://github.com/atom/tree-view/blob/master/lib/file.coffee#L54) and in [file-view.coffee](https://github.com/atom/tree-view/blob/master/lib/file-view.coffee#L29)

Proposed implementation:

- split status-modified into more statuses: probably status-staged-modified and status-modified
- status-modified-staged would be green, not orange.
- if it has both status-staged-modified and status-modified (eg MM in the git cli), we would use a different indicator, maybe a gradient between orange and green.

PR #955, opened 2016-10-09
Bolding file name if it’s not staged.

victor-torres:
As mentioned in these issues across Atom repositories:

- https://github.com/atom/tree-view/issues/430
- https://github.com/atom/atom/issues/8529

I found it difficult to choose a new color schema to identify unstaged files so I’ve just **bolded** the filename to identify files not staged yet. 

I maintain a plugin called [atom-hg](https://github.com/victor-torres/atom-hg) and my `isStatusStaged` method is already exposed. There’s a method with the same name in [git-utils](https://github.com/atom/git-utils/blob/v3.0.0/src/git.coffee)  but it doesn’t seem to be exposed yet so I manually added the same verification as a fallback (`(status & 15) > 0`). 

It would help a lot because we’ll finally have a clue whether a file is staged or not.

### DIFF
--- a/lib/directory-view.coffee
+++ b/lib/directory-view.coffee
@@ -58,8 +58,9 @@ class DirectoryView extends HTMLElement
     @expand() if @directory.expansionState.isExpanded
 
   updateStatus: ->
-    @classList.remove('status-ignored', 'status-modified', 'status-added')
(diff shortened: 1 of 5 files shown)


victor-torres:
Could someone take a look here?

a7madgamal:
this is awesome. any updates?

PR #958, opened 2016-10-12
Fix scrolling performance issues in tree-view

easyaspi314:
Part of fix for atom/atom#12949.

Forces Chromium to use the compositor to render scroll containers. This greatly decreases CPU usage when scrolling.

### DIFF
--- a/styles/tree-view.less
+++ b/styles/tree-view.less
@@ -46,6 +46,7 @@
   flex: 1;
   width: 100%;
   overflow: auto;
+  will-change: transform;
 }
 
 .tree-view {


simurai:
Tested this change and scrolling sometimes seems smoother, but all the time. Also it seems that even though the `.tree-view-scroller` uses a compositor layer, it still needs painting. See how the paint count goes up when scrolling:

perf2

I think the problem is that the content inside a scrolling area needs to be clipped and therefore repainted. Not sure if it's possible to use `overflow: auto` without having to repaint? The editor uses compositor layers without repainting, but has to take care of moving the lines/tiles itself.

PR #967, opened 2016-10-28
Remove unnecessary onWillDeactivate method from file-icons service

maxbrunsfeld:
There is no need for this method, since Atom's service API already provides a way for consuming packages to be notified when a service is removed: the `Disposable` returned from the consume callback.

I'm leaning towards *not* bumping the semver version of `file-icons`, since there shouldn't be any problem with a provider implementing an unused `onWillDeactivate` method. Thoughts?

/cc @Alhadis @DanBrooker

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,4 +19,3 @@ The Tree View displays icons next to files. These icons are customizable by inst
 The `atom.file-icons` service must provide the following methods:
 
 * `iconClassForPath(path)` - Returns a CSS class name to add to the file view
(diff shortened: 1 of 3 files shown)


Alhadis:
Burn it. Not only is it redundant, but it's also dysfunctional. Calling it the way you might expect ends up wiping the icons before they're displayed, resulting in one confusing mess.

maxbrunsfeld:
Temporarily ignoring the appveyor failure since it is already happening on master.

Alhadis:
Uncropped version of Appveyor's logo:

It all makes sense

PR #970, opened 2016-11-02
don't attempt to expand the selected entry if it is a file

zoo1:
fix for #921


Ben3eeE:
Can we change this to only expand if the selected entry is a directory or check if `selectedEntry` exists? This function will throw an error if nothing is selected even after this change I think.

Issue #1005, opened 2016-12-23
Deleting collapsed subdirectory breaks tree-view


marcoms:
### Prerequisites

* [x] Reproduced the problem in Safe Mode: http://flight-manual.atom.io/hacking-atom/sections/debugging/#using-safe-mode
* Followed all applicable steps in the debugging guide: http://flight-manual.atom.io/hacking-atom/sections/debugging/
* Checked the FAQs on the message board for common solutions: https://discuss.atom.io/c/faq
* [x] Checked that your issue isn't already filed: https://github.com/issues?utf8=✓&q=is%3Aissue+user%3Aatom
* [x] Checked that there is not already an Atom package that provides the described functionality: https://atom.io/packages

### Description

If there is single subdirectory which causes the tree-view to collapse it into a single entry like `folder-a/folder-b`, if the user attempts to delete it by right clicking and deleting it, the subdirectory is deleted, but the entry is still listed as `folder-a/folder-b`, when only `folder-a` exists. Attempting to delete it again using the tree-view gives an error ("The following file couldn't be moved to trash").

### Steps to Reproduce

1. Launch Atom
1. Ensure `tree-view.squashDirectoryNames` is `true`
1. Create a directory with a single subdirectory
2. Attempt to delete the collapsed subdirectory through the tree-view UI

**Expected behavior:** The directory should no longer be collapsed and display only the first directory

**Actual behavior:** The subdirectory is still listed although it is deleted

**Reproduces how often:** 100%

### Versions

```
Atom    : 1.13.0-beta8
Electron: 1.3.13
Chrome  : 52.0.2743.82
Node    : 6.5.0
```

tree-view version 0.211.1

### Additional Information

Collapsing and expanding the parent of the directory refreshes the state and displays properly.

Related #702

miguelcg:
Pull request #1008 opened for this issue, feel free to review it! :)

50Wliu:
Fixed by #1008.

Issue #799, opened 2016-04-08
"The following file couldn't be moved to trash (is gvfs-trash installed?)" when attempting to delete empty nested folder with "Collapse direc


SlimeQ:
![screenshot from 2016-04-08 17 21 42](https://cloud.githubusercontent.com/assets/3934698/14398008/d9b86574-fdae-11e5-8cdf-f2ba6e8d7bd2.png)

with "Collapse directories" enabled, deleting an empty nested directory fails

---
## ubuntu 14.04, atom 1.6.2

**update:** the first time attempting to delete `/1/2/3`, directory 3 is deleted but the tree is not updated. if the delete is attempted again the message above is shown. if the parent directory is collapsed and expanded again, it now shows `/1/2`. this process can be repeated until all directories are gone.

this seems somewhat related to #702

mnzt:
Same error issue on arch linux:

```
Atom    : 1.7.2
Electron: 0.36.12
Chrome  : 47.0.2526.110
Node    : 5.1.1

 Kernel Release: 4.5.1-1-ARCH
OS: Arch Linux x86_64
```

azat-io:
Same error 😞

piecedigital:
So, like... is this a recent issue? This wasn't happening a few days ago. It worked right out the box, unlike the Windows version.

50Wliu:
Well, the notification was added in a recent release.  Before failing to delete a file would fail silently without an error message.
@mnzt, @azat-io, @piecedigital: are you all trying to delete a nested directory with "collapse directories" enabled?

piecedigital:
@50Wliu I'm trying to delete several files in one directory, not even the directory itself.

mnzt:
@50Wliu same story as @piecedigital. Even doing the following results in the error.
```
cd ~/
atom some.file
```
followed  by trying to delete `some.file`

piecedigital:
Okay... I fixed my issue simply by changing the owner to myself.

Not sure how it changed in the first place...

mnzt:
I still cannot delete _any_ file, even if owned by the user who launched atom

50Wliu:
Ok.  As this issue is talking specifically about deleting nested directories while "collapse empty directories" is enabled, can you please create a different issue?  Thanks!

adnanzulkarnain:
Same error

Thojad:
Same error on Elementary OS Freya...

gauthier-delacroix:
[...]

PR #1008, opened 2016-12-29
🐛 Deleting squashed directory

miguelcg:
### Requirements

### Description of the Change

Calling `@updateRoots()` after the deletion of a squashed directory refresh the view and solves the bug in #1005.

### Alternate Designs

-

### Benefits

It solves a bug. The expected behavior after deleting a directory is to make it disappear from the tree view. That's possible with this pull request. The current behavior is that the directory is deleted in the file system but the tree view keeps showing the un-existing directory.

### Possible Drawbacks

-

### Applicable Issues

#1005 #799


50Wliu:
Thanks!

Issue #1003, opened 2016-12-22
files closable within tree


Albertc:
Is it possible to add closable icon "x" in the file name? without using "tree-view-open-files" package, like Textmate:

![image](https://cloud.githubusercontent.com/assets/124529/21423574/08e9b9ce-c83e-11e6-8e61-0e72f89bff96.png)

### Steps to Reproduce

1. Launch Atom
1. Open a file
1. Click `X` next to file in Tree View

**Expected:** Open file tab is closed via normal closing behavior

miguelcg:
Pull request is on the way for this enhancement. I just need to figure out how to close a file (Destroy the `TextEditor` I guess) when the `X` icon is clicked... Any ideas about that @lee-dohm ?

lee-dohm:
To be honest, I'm not sure this is something that we're going to want to accept yet. Most of the team is still out for the holidays right now. I've got it on my list of things to discuss with the team when we get back next week.

If you're still curious, I would look at the implementation of the `core:close` command to see how it works.

miguelcg:
Thanks for letting me know @lee-dohm. I have this feature currently working (I'm polishing it) so when it's ready I'll create a pull request and it's up to you and the team to accept it or not. At least people will be able to see the code and add the feature to their own fork if they interested.
What do you think?

lee-dohm:
Sounds good to me. Thanks! :+1:

miguelcg:
No problem @lee-dohm, PR opened as agreed! #1010

lee-dohm:
Thanks very much for the suggestion @Albertc and the work @miguelcg :grinning: I've discussed this with the team and this isn't a feature that we want to add to Tree View. It could probably be implemented as a package that adds the button to Tree View or a fork of Tree View itself.

Thanks again for contributing to Atom :tada:

PR #1010, opened 2017-01-02
Close files within the tree view

miguelcg:
### Requirements

### Description of the Change
Adds a new feature. When a file is opened an icon (icon-x) is shown on the right of the file name in the tree view. If this icon is clicked the file gets closed.

### Alternate Designs

-

### Benefits

New feature.

### Possible Drawbacks

-

### Applicable Issues

#1003

### DIFF
--- a/lib/file-view.coffee
+++ b/lib/file-view.coffee
@@ -25,8 +25,24 @@ class FileView extends HTMLElement
         iconClass = iconClass.toString().split(/\s+/g)
       @fileName.classList.add(iconClass...)
 
+    @fileCloser = document.createElement('span')
+    @fileCloser.classList.add('icon-x')
+
+    @subscriptions.add @file.onDidOpen => @open()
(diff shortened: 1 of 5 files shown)


50Wliu:
A possible UI improvement would be to only show the `x` icon on hover or when the file is focused in the tree view.  I can see the benefits of always having it visible though.

miguelcg:
Not sure about that @50Wliu. I think that having the `x` icon visible in all the opened files is more user-friendly. It allows users to see which files are opened at a glance.

lee-dohm:
Thanks for the work you invested in this @miguelcg :grinning: We don't want to add this feature but it could probably be refactored into a package if you'd like to do that 👍 

Thanks again :bow:

miguelcg:
No worries @lee-dohm 😃 I'll think about the package idea 😉

Issue #856, opened 2016-06-01
Unable to delete files on Windows without a Recycle Bin


manuelbieh:
The issue is probably related to https://github.com/atom/tree-view/issues/666 – i still cannot delete files or folders from within the tree-view when the files are located on a mapped virtual drive.

Steps to reproduce:
- Create a new registry entry (and restart Windows afterwards):

```
REGEDIT4

[HKEY_LOCAL_MACHINE\SYSTEM\CurrentControlSet\Control\Session Manager\DOS Devices] 
"W:"="\\??\\D:\\htdocs"
```

_edit:_
`subst W: D:\htdocs` – causes the same problem
- Create a file on your freshly mapped DOS device `W:\example\example.txt`
- Open folder `W:\` in atom
- Try to delete `W:\example\example.txt` using the tree-view -> fails

Opening `D:\htdocs` instead and then trying to delete `example.txt` works fine. File and folder rights are identical. Creating files or folders or renaming them works completely fine, only deletion is not possible (see my attached screenshots).

Comparison:
![comparison](http://i.imgur.com/E6vDk31.png)

Broken
![broken](http://i.imgur.com/kzzPTLe.png)

Deletion when opening `D:\htdocs` instead works fine!

Ben3eeE:
I have the same issue with network mapped drives on Windows. Always assumed it was a duplicate of https://github.com/atom/tree-view/issues/830 so never created an issue for it.

manuelbieh:
Seems to be pretty similar, indeed.

manuelbieh:
I think I found the reason: virtual drives do not have an explicit recycle bin folder where Atom could move the files into:

```
Manuel@Manuel-406 /cygdrive
$ ls /cygdrive/c/'$Recycle.Bin'
S-1-5-18  S-1-5-21-2028212831-3322830245-2971025389-1001  S-1-5-21-4197599123-4253082845-4033610845-500

Manuel@Manuel-406 /cygdrive
$ ls /cygdrive/w/'$Recycle.Bin'
ls: Zugriff auf '/cygdrive/w/$Recycle.Bin' nicht möglich: No such file or directory
```

That's probably the reason why it fails: the folder Atom is trying to move the file into just does not exist.

manuelbieh:
this workaround by saeed sagharichi works for me at the moment!

[...]

PR #1011, opened 2017-01-02
Cover trash failures for macOS and Windows with a message

lee-dohm:
### Description of the Change

Deleting files from the Tree View can fail if the volume, drive or share doesn't support the platform's implementation of a "trash". Because this occurs much more frequently on Linux, we previously covered this with an enhanced error message in https://github.com/atom/tree-view/pull/695. This problem can still occur on Windows or macOS though and people on those platforms deserve a better error message to give them some idea of why Atom is unable to delete their files for them.

### Alternate Designs

Considered returning more information from Electron's [`shell.moveItemToTrash` function](http://electron.atom.io/docs/all/#shellmoveitemtotrashfullpath) so that it could be detected if it was truly a trash problem or some other failure to delete the file. Adding support for better error conditions in a cross-platform way is beyond the scope of my time and abilities though, so this is what I opted for.

### Benefits

* People will get a less cryptic error message

### Possible Drawbacks

* The error message still may not point them in the right direction

### Applicable Issues

* #856 
* https://discuss.atom.io/t/can-not-delete-files-from-tree-view/37639

### DIFF
--- a/lib/tree-view.coffee
+++ b/lib/tree-view.coffee
@@ -588,7 +588,7 @@ class TreeView extends View
     switch process.platform
       when 'linux' then 'Is `gvfs-trash` installed?'
       when 'darwin' then 'Is Trash enabled on the volume where the files are stored?'
-      when 'win32' then 'Is RecycleBin support enabled on the drive where the files are stored?'
+      when 'win32' then 'Is there a Recycle Bin on the drive where the files are stored?'
 
   # Public: Copy the path of the selected entry element.
   #         Save the path in localStorage, so that copying from 2 different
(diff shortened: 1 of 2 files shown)


lib/tree-view.coffee
```diff
@@ -573,12 +573,23 @@ class TreeView extends View
...
+  formatTrashFailureMessage: (failedDeletions) ->
+    fileText = if failedDeletions.length > 1 then 'files' else 'file'
+
+    "The following #{fileText} couldn't be moved to the trash. #{@formatTrashEnabledMessage()}"
+
+  formatTrashEnabledMessage: ->
+    switch process.platform
+      when 'linux' then 'Is `gvfs-trash` installed?'
+      when 'darwin' then 'Is Trash enabled on the volume where the files are stored?'
+      when 'win32' then 'Is RecycleBin support enabled on the drive where the files are stored?'
```
50Wliu: Recycle Bin should be two words.  An alternative wording could also be `Is there a Recycle Bin on the drive where the files are stored?`

lee-dohm: Done. Thanks!

PR #1022, opened 2017-01-10
🐛 Prevent files to be expanded

miguelcg:
### Description of the Change

Solves #921. Prevents files to be expanded (and therefore causing an error).
This PR is a copy of #970 but with a test scenario.

### Benefits

Solves #921

### Applicable Issues

#921

### DIFF
--- a/lib/tree-view.coffee
+++ b/lib/tree-view.coffee
@@ -392,6 +392,7 @@ class TreeView extends View
 
   expandDirectory: (isRecursive=false) ->
     selectedEntry = @selectedEntry()
+    return unless selectedEntry instanceof DirectoryView
     if isRecursive is false and selectedEntry.isExpanded
       @moveDown() if selectedEntry.directory.getEntries().length > 0
     else
--- a/spec/tree-view-spec.coffee
+++ b/spec/tree-view-spec.coffee
@@ -1098,12 +1098,12 @@ describe "TreeView", ->
 
       describe "when trying to expand a file", ->
         it "won't raise an error", ->
-          sampleTxt.click()
-          treeView.roots[0].collapse()
+          file = root1.find('.file:contains(tree-view.js)')[0]
+          treeView.selectEntry(file)
 
-          expect(treeView.roots[0]).not.toHaveClass 'expanded'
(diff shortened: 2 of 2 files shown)


50Wliu:
It doesn't look like the spec properly covers this change.  Does it pass if you comment out that line?

miguelcg:
You were right @50Wliu. If you commented the line it didn't work but for other reason. Now the test actually tests the right thing.
Let me know if there's anything pending to get this merged.

miguelcg:
@50Wliu Is there anything pending from my side?

50Wliu:
Will take a look soon.

miguelcg:
Thanks @50Wliu!

50Wliu:
I still think the test can be improved.  Here's what I have:
```coffee
describe "when trying to expand a file", ->
        it "doesn't expand the file and doesn't error", ->
          file = root1.find('.file:contains(tree-view.js)')[0]
          treeView.selectEntry(file)
          expect(file).not.toHaveClass 'expanded'

          expect(-> treeView.expandDirectory(true)).not.toThrow()
          expect(file).not.toHaveClass 'expanded'

          expect(-> treeView.expandDirectory(false)).not.toThrow()
          expect(file).not.toHaveClass 'expanded'
```
Now there's an explicit check to make sure that it doesn't throw.

miguelcg:
Nice one @50Wliu! Can you commit the test? 😃

50Wliu:
Also, now the implementation differs from recursive-collapse-directory, which searches for the closest directory and collapses that.  What do you think about doing the same thing for recursive-expand-directory?

miguelcg:
@50Wliu Actually I think that it should be the other way around. `recursive-collapse-directory` should recursively collapse the directories and not only collapse the closest one I reckon.
What do you think?

miguelcg:
Finally AppVeyor check passing! Nice work @as-cii!

50Wliu:
Sorry, what I meant was that recursive-collapse-directory searches for the closest directory (not sure if this means the parent directory or not) if you execute it on a file, and then recursively collapses that directory.

PR #1028, opened 2017-01-31
Remove jQuery from `spec/tree-view-spec.coffee`

as-cii:
The benefits of this pull request are twofold:

1. **Tests will now pass on Windows.** We have finally discovered the cause of the hard crashes we were observing on Windows when running tests. We believe it can be attributed to jQuery leaking `pathwatcher` objects when wrapping tree view elements inside a `$()` function.
2. **It paves the way for the `atom-space-pen-views` removal from production code.** In facts, the work we did on this pull request should have happened anyways as part of the space-pen removal we are planning to do. By doing it first and in a separate pull request, however, we can ensure the old behavior is preserved both on Windows and on macOS.

/cc: @atom/maintainers

### DIFF
--- a/lib/dialog.coffee
+++ b/lib/dialog.coffee
@@ -29,7 +29,7 @@ class Dialog extends View
       @miniEditor.getModel().setSelectedBufferRange(range)
 
   attach: ->
-    @panel = atom.workspace.addModalPanel(item: this.element)
+    @panel = atom.workspace.addModalPanel(item: this)
     @miniEditor.focus()
     @miniEditor.getModel().scrollToCursorPosition()
 
--- a/spec/tree-view-spec.coffee
+++ b/spec/tree-view-spec.coffee
@@ -2555,11 +2555,7 @@ describe "TreeView", ->
         expect(zetaEntries).toEqual(["zeta.txt"])
 
       it "squashes two dir names when the first only contains a single dir", ->
-        if path.sep is '\\'
-          # First escape the backslashes for Coffeescript, then escape them for jQuery
-          betaDir = findDirectoryContainingText(treeView.roots[0], "alpha\\beta")
-        else
-          betaDir = findDirectoryContainingText(treeView.roots[0], "alpha#{path.sep}beta")
+        betaDir = findDirectoryContainingText(treeView.roots[0], "alpha#{path.sep}beta")
         betaDir.expand()
         betaEntries = [].slice.call(betaDir.children[1].children).map (element) ->
           element.innerText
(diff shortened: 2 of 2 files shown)


as-cii:
It turns out that removing jQuery helped locally but was unfortunately only a red herring as it still fails pretty consistently on AppVeyor: my main hypothesis right now is that there is a race condition happening in path watcher and that removing jQuery only made it less evident on my virtual machine. The work on this pull request is still valuable though, as it will be used in a subsequent pull request that will get rid of atom-space-pen-views from tests as well as from production code.

Therefore, I am closing this for now. I am currently investigating another approach for this in https://github.com/atom/node-pathwatcher/pull/118.

nathansobo:
Just a thought... would it be worth exploring `nsfw` for this? Another thing I've thought about is a centralized path watching service as part of core so we can minimize the overhead of watching the same paths multiple times in different packages.

smashwilson:
> Another thing I've thought about is a centralized path watching service as part of core so we can minimize the overhead of watching the same paths multiple times in different packages.

+:100: Cross-platform path watching seems like a really common need that would benefit greatly from standardization as a core API. I know that [`File`](https://atom.io/docs/api/v1.13.1/File) and [`Directory`](https://atom.io/docs/api/v1.13.1/Directory) already have some watching capabilities. Maybe we could expand that and back it with nsfw?

If we do go with nsfw, we'll need to address the thread-per-watcher architecture. That could get out of hand pretty quickly once every package is able to start its own watchers.

nathansobo:
> I know that `File` and `Directory` already have some watching capabilities

I don't necessarily *love* those early object-oriented file system APIs. Something oriented around paths or URIs would feel simpler to me, though I'm open to being convinced otherwise.

smashwilson:
> I don't necessarily love those early object-oriented file system APIs.

Fair, those are definitely more Coffeescript than JS. Maybe introduce a new watcher API (`atom.watch`, or a versioned package service?), back File and Directory with those, then deprecate them... ?

as-cii:
I definitely like the idea of having a core facility that provides a path watching service. On the other hand, however, I don't feel comfortable exposing an abstraction that's based on nsfw until we sort out all of its issues. Moreover, getting this facility right so that it can make part of the core API (i.e. ensuring that the behavior doesn't change) might take a couple of days and it feels a little like a detour from our main goal of reducing startup time (considering that the task of removing atom-space-pen-views is somewhat a detour itself).

Thoughts? I am open to discuss this further in case you feel we should invest time in such abstraction. In the meantime I have created a pull request on `pathwatcher` that prevents the aforementioned race condition and opened a pull request on this repository that fixes tests on Windows (this time for real).

nathansobo:
Yeah, I wasn't necessarily suggesting doing it immediately. Should have made that clear. Just wanted to raise the issue.

iolsen:
> I definitely like the idea of having a core facility that provides a path watching service. On the other hand, however, I don't feel comfortable exposing an abstraction that's based on nsfw until we sort out all of its issues. Moreover, getting this facility right so that it can be part of the core API (i.e. ensuring that the behavior doesn't change) might take a couple of days and it feels a little like a detour from our main goal of reducing startup time (considering that the task of removing atom-space-pen-views is somewhat a detour itself).

I agree with all of this except the "couple days." 😈 I suspect it'll take longer than that. I think things are working well enough now that this isn't an urgent priority, but we should keep it in mind. @smashwilson has been working on IDing the remaining crash(es) in nsfw we're seeing very rarely on Windows.

PR #1029, opened 2017-02-01
Fix tests on Windows

as-cii:
Supersedes #1028. Depends on https://github.com/atom/node-pathwatcher/pull/118.

This pull request bumps path watcher in order to solve a race condition that was causing hard crashes on Windows, thus causing the specs to fail.

### DIFF
--- a/package.json
+++ b/package.json
@@ -14,7 +14,7 @@
     "event-kit": "^1.0.0",
     "fs-plus": "^2.3.0",
     "minimatch": "~0.3.0",
-    "pathwatcher": "^6.2",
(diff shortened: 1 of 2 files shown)


as-cii:
/cc: @atom/maintainers

It would be awesome to get extra 👀 also on https://github.com/atom/node-pathwatcher/pull/118 as it involves multithreading and I'd like to make sure we're not introducing new race conditions or deadlocks.

as-cii:
The Travis build seems stuck for some reason, but AppVeyor seems to work smoothly now! I'll 🚢 this and start working on removing custom elements and jQuery.

PR #1032, opened 2017-02-06
Remove atom-space-pen-views and custom elements

as-cii:
This pull request replaces atom-space-pen-views with manual DOM manipulation. It also changes `DirectoryView` and `FileView` so that they don't register themselves as custom elements, but are simple JavaScript objects managing a DOM element. Feature-wise nothing should change, but this will help remove jQuery from Atom.

/cc: @ungb for 👀

### DIFF
--- a/lib/tree-view.coffee
+++ b/lib/tree-view.coffee
@@ -219,15 +219,13 @@ class TreeView
       @show()
 
   entryClicked: (e) ->
-    entry = e.target.closest('.entry')
-    isRecursive = e.altKey or false
-    @selectEntry(entry)
-    if entry.classList.contains('directory')
-      entry.toggleExpansion(isRecursive)
-    else if entry.classList.contains('file')
-      @fileViewEntryClicked(e)
-
-    false
+    if entry = e.target.closest('.entry')
+      isRecursive = e.altKey or false
+      @selectEntry(entry)
+      if entry.classList.contains('directory')
+        entry.toggleExpansion(isRecursive)
+      else if entry.classList.contains('file')
+        @fileViewEntryClicked(e)
 
   fileViewEntryClicked: (e) ->
     filePath = e.target.closest('.entry').getPath()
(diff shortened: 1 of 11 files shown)


as-cii:
@ungb: some things in this pull request that might be worth testing 💭 

1. Drag-and-dropping files.
2. Drag-and-dropping directories.
3. Drag-and-dropping project roots.
4. Keyboard navigation.
5. Mouse navigation.
6. Interactions with third-party packages.
7. Tree-view commands in the Command Palette.

lib/tree-view.coffee
```diff
@@ -205,19 +219,19 @@ class TreeView extends View
       @show()
 
   entryClicked: (e) ->
-    entry = e.currentTarget
+    entry = e.target.closest('.entry')
     isRecursive = e.altKey or false
     @selectEntry(entry)
-    if entry instanceof DirectoryView
+    if entry.classList.contains('directory')
```
Alhadis: This raises an exception if the user clicks on the tree-view's background, rather than a specific entry:

~~~
tree-view.coffee [sm]:225
Uncaught TypeError: Cannot read property 'classList' of null
	module.exports.TreeView.entryClicked  @ tree-view.coffee [sm]:225
	(anonymous function)                  @ tree-view.coffee [sm]:100
~~~

as-cii: Thank you so much for the heads-up! Fixed in 31c472d. 👍

Alhadis: Haha, my pleasure! I'm running latest changes through Mocha over and again, so it's easy to spot when something breaks. =)

Ben3eeE:
`tree-view:show-in-file-manager` and `tree-view:show-current-file-in-file-manager` opens the `Documents` folder instead of the folder of the selected/current file. I am on Windows 10 Atom `1.14.0-beta4 ia32`

Alhadis:
[`seti-ui`](https://github.com/jesseweed/seti-ui/blob/master/lib/headers.coffee#L35) threw an error after I enabled it, and it was one of the errors I fixed to get `file-icons` future-proof:

Figure 1

Might be worth shimming `on` and having Grim issue a deprecation notice for users.

as-cii:
Thanks again for the amazing feedback and for providing a fix on the seti-ui package, @Alhadis. Your help is really appreciated! ✨

I am on the fence about shimming because it seems like it could be non-trivial to provide a compatible interface. In particular, there are a lot of jQuery methods that might be called from third-party packages and re-implementing them by using DOM primitives could take long, other than being potentially inaccurate. 

An alternative might be to use a `Proxy` object and forward all the unknown methods to jQuery but, again, I am not entirely sure that would be the best path forward. Maybe we should let this bake for a while on beta until everyone has the chance to transition away from jQuery APIs? Most of the time the fix will be trivial for package authors and we could also write a blog post informing everyone about the change.

Alhadis:
Oh, I was only talking about shimming the tree-view's `on` method, since it has the greatest change of being tampered with by package authors (\*cough\* …guilty as charged). In which case, it's just a simple matter of `Object.defineProperty(treeView.scroller, "on", {value(event, fn){ return this.addEventListener(event, fn); }});`, or something similar.

To be honest, breakage will probably get packages fixed quicker than deprecation notices. That is, if the speed at which `seti-ui` removed their shadow DOM selectors is anything to go by... (6 days ago)

as-cii:
> Oh, I was only talking about shimming the tree-view's on method, since it has the greatest change of being tampered with by package authors (*cough* …guilty as charged). In which case, it's just a simple matter of Object.defineProperty(treeView.scroller, "on", {value(event, fn){ return this.addEventListener(event, fn); }});, or something similar.

Yeah, my main concern with that is that we would also need to wrap the DOM event inside a jQuery event object and give users a way of providing a selector as the second parameter of the `on` call. We would also need to do something similar on the tree-view object itself, as well as on other internals of the tree-view package.

> To be honest, breakage will probably get packages fixed quicker than deprecation notices. That is, if the speed at which seti-ui removed their shadow DOM selectors is anything to go by... (6 days ago)

Agreed. I also think that the number of packages affected by this is actually low, so it might be worth seeing what happens on beta and then decide whether a shim is worth it or not. 👍

Alhadis:
Oh right, I completely forgot about the `[0]` part as well... uggggh, yeah, you're right, screw it. Not worth the complexity, and authors should've been using documented APIs in the first place. 😀

lib/tree-view.coffee
```diff
@@ -401,26 +432,26 @@ class TreeView extends View
...
 
-    if directory = $(selectedEntry).closest('.expanded.directory')[0]
+    if directory = selectedEntry.closest('.expanded.directory')
       directory.collapse(isRecursive)
       @selectEntry(directory)
 
   openSelectedEntry: (options={}, expandDirectory=false) ->
     selectedEntry = @selectedEntry()
-    if selectedEntry instanceof DirectoryView
+    if selectedEntry.classList.contains('directory')
```
Alhadis: This will break when nothing is selected, because `selectedEntry` is null.

Happened after I hit <kbd>Enter</kbd> with focus on the tree-view by mistake.

as-cii: Fixed in a7e7be1. @Alhadis: thanks so much for finding these pesky regressions! ⚡️

ungb:
Looks good to me, but I'm also seeing the same issue @Alhadis is with seti-ui, I wonder if this will cause other breakages for other packages.  We can maybe put it out on beta and use bugsnag to capture more exceptions.

PR #1044, opened 2017-03-10
Update editors when containing directory is renamed or removed

### LINKED ISSUE #948: If renaming a folder while file in that folder open, saving the file will save it in old location
I have a file open in the editor, e.g. `some/path/file.js`.
If I rename the folder via tree view to be `some/other-path`, while my file is open, when I then proceed to edit the file again and save it, it will save it as `some/path/file.js`, not realising that the new path for the file is now `some/other-path/file.js`.

This can be rather frustrating, as the new file (in the old location) is created silently, and you're left wondering why your changes aren't appearing/working, until you realise you are editing the wrong file.

Happens on OS X with Atom 1.10.0-dev-98fc8db

kuychaco:
Fixes #948.
Fixes atom/tabs#12.
Fixes atom/tabs#413.
Fixes atom/atom#9875.

Prior to this PR, renaming or removing a folder would not result in open text editors being updated appropriately. 

**Changes introduced in this PR include:**
* folder renames result in updating the paths for editors corresponding to files within the renamed folder (for both drag & drop and move dialog renames)
* folder deletions result in the closing of editors corresponding to files within the deleted folder

**Alternative approaches were discussed and ultimately dismissed for the reasons outlined:**
* Use a directory-watching library such as [nsfw](https://github.com/Axosoft/nsfw/) to catch rename and delete events
  * Benefits: We would additionally be able to support the case where folders are renamed or deleted outside of Atom. 
  * Possible drawbacks: Watching large directories on Linux is too costly so we would only be able to support this on Mac and Windows. Thus, we would have to implement a solution such as the one in this PR to fix these issues on Linux.
* Centralize file system operations in Atom core and create new APIs such as `atom.project.rename` or `atom.fs.rename`
  * Benefits: Other could rely on these APIs and trust that Atom will take care of updating editors appropriately 
  * Possible drawbacks: This would take careful thought and API design as we would be introducing APIs that we would have to continue to support.

### DIFF
--- a/lib/helpers.coffee
+++ b/lib/helpers.coffee
@@ -22,3 +22,10 @@ module.exports =
       fullExtension = extension + fullExtension
       filePath = path.basename(filePath, extension)
     fullExtension
+
+  updateEditorsForPath: (oldPath, newPath) ->
+    editors = atom.workspace.getTextEditors()
+    for editor in editors
+      filePath = editor.getPath()
+      if filePath.startsWith(oldPath)
+        editor.getBuffer().setPath(filePath.replace(oldPath, newPath))
--- a/lib/helpers.coffee
+++ b/lib/helpers.coffee
@@ -27,5 +27,5 @@ module.exports =
     editors = atom.workspace.getTextEditors()
     for editor in editors
       filePath = editor.getPath()
-      if filePath.startsWith(oldPath)
+      if filePath?.startsWith(oldPath)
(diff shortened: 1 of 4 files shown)


Alhadis:
@kuychaco Any chance you could implement this using #966 instead? =) Part of the same issue.

kuychaco:
@Alhadis cool, thanks for the ping :). Will pick this up next week and take a look then 👍

PR #1047, opened 2017-03-16
Only check panes in center

matthewwithanm:
atom/atom#13977 adds multiple pane locations and updates the
`atom.workspace.getPanes()` API to return the panes from all of them.
This fixes the resultant test failures by only checking the panes in
the center.


maxbrunsfeld:
🚂  🚂

PR #1049, opened 2017-03-25
Update editors when containing directory is renamed or removed (redux)

### LINKED ISSUE #948: If renaming a folder while file in that folder open, saving the file will save it in old location

BinaryMuse:
This is another pass at https://github.com/atom/tree-view/pull/1044, but basing the work on @Alhadis' changes at https://github.com/atom/tree-view/pull/966 (which fit in perfectly with this feature). See the former PR for more details on the feature.

Katrina and I are moving on to other projects, but since we didn't get to this in the past week, I wanted to take care of it now.

@Alhadis, I pulled in the changes from your PR and made some changes on top. Namely:

* `CopyDialog` and `MoveDialog` now take callbacks instead of relying on the emitter directly
* I've standardized the key names from the new event objects so that if there are two paths they are always `initialPath` and `newPath`, and if there is only one it's always `path`
* I modified the tests to use the new public methods directly instead of reaching into the emtiter internals

Let me know if that causes any conern for you, and thank you so much for your work on this!

Closes https://github.com/atom/tree-view/pull/1044
Closes https://github.com/atom/tree-view/pull/966

Fixes #948
Fixes atom/tabs#12
Fixes atom/tabs#413
Fixes atom/atom#9875

/cc @kuychaco

### DIFF
--- a/lib/tree-view.coffee
+++ b/lib/tree-view.coffee
@@ -2,7 +2,7 @@ path = require 'path'
 {shell} = require 'electron'
 
 _ = require 'underscore-plus'
-{BufferedProcess, CompositeDisposable} = require 'atom'
+{BufferedProcess, CompositeDisposable, Emitter} = require 'atom'
 {repoForPath, getStyleObject, getFullExtension} = require "./helpers"
 fs = require 'fs-plus'
 
@@ -41,6 +41,7 @@ class TreeView
     @element.appendChild(@resizeHandle)
 
     @disposables = new CompositeDisposable
+    @emitter = new Emitter
     @focusAfterAttach = false
     @roots = []
     @scrollLeftAfterAttach = -1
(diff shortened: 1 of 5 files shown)


BinaryMuse:
@kuychaco This is ready for a review, and subsequently ready for a merge pending @Alhadis' 👀

Alhadis:
> `CopyDialog` and `MoveDialog` now take callbacks instead of relying on the emitter directly

How does another package respond to these callbacks? It'd seem that an entirely different `Dialog` instance would need to be created each time.

BinaryMuse:
> How does another package respond to these callbacks?

They wouldn't; this is just an implementation detail. Other packages would consume the top-level API.

Alhadis:
Oh, sorry... I misinterpreted completely.

👍  In which case, looks good to me!

lib/tree-view.coffee
```diff
@@ -73,6 +74,14 @@ class TreeView
     @element.style.width = "#{state.width}px" if state.width > 0
     @attach() if state.attached
 
+    @disposables.add @onEntryMoved ({initialPath, newPath}) ->
+      updateEditorsForPath(initialPath, newPath)
+
+    @disposables.add @onEntryDeleted ({path}) ->
+      for editor in atom.workspace.getTextEditors()
+        if editor?.getPath()?.startsWith(path)
+          editor.destroy()
```
steelbrain: Maybe not destroy them if they have unsaved changes?

BinaryMuse: Hm, this code path didn't change (it was only moved). I'll double check again but I was pretty sure that it left the editor open with unsaved changes.

BinaryMuse: Yeah, you're right - it definitely does get destroyed. Sounds like a bug to me! I'll check for another issue and if one doesn't exist, create one. Thanks.

Issue #68, opened 2014-03-04
Allow Tree-View when no file/folder


monkeymonk:
Because I usually drag a project folder inside the Tree-View... but for now that open a new window Editor. That's mean I have an editor with my project folder and the tree-view and a editor with nothing...

metalim:
No idea what should it show, when there's no file, but when there IS a file, then - folder containing that file.

monkeymonk:
In `Sublime Text`, we have a simple empty pane.
![Uploading Capture d’écran 2015-01-14 à 21.59.00.png . . .]()

xenio:
I agree, tree view should be open also with no files, it is a simple behaviour of finding the start of the code always on the same position. If I start Atom without any files the code start on the left side of the screen and it is very annoying on large screen.

ricardograca:
Well, I just completely wiped my Atom installation and user configuration and reinstalled Atom from scratch (two times) because I thought there was some bug in my configuration preventing the tree view from opening. When even the "Toggle Tree View" command does nothing, it sure looks like a bug.

kfix:
I have this at the end of my `init.coffee` to deal with this/my quirky behavior:

```
#bare opens of atom.app should bring up the file tree
if not atom.project.getPath()
  atom.project.setPath(atom.config.get('core.projectHome'))
  atom.commands.dispatch(atom.workspaceView.element, 'tree-view:show')
```

mnquintana:
From https://github.com/atom/atom/issues/6179#issuecomment-92413675, it might be worth showing a notification if you try opening the tree view with no project folder added. That way it won't seem broken if nothing happens.

ricardograca:
Or just open an empty tree view since that's probably the expected behavior.

thedaniel:
> Or just open an empty tree view since that's probably the expected behavior.

Yes, and perhaps the tree view could have an empty state like the panes do that display a message.

mnquintana:
> Yes, and perhaps the tree view could have an empty state like the panes do that display a message.
:+1: That sounds good

RacerDelux:
Going to poke this along, I just ran into this issue as well, waisted a couple hours of work time before I figured out what was wrong. Really something that needs to be added.

dcrockwell:
Getting caught by this problem myself on Windows 7. Even when opening from Explorer, Atom cannot toggle the treeview with `Ctrl+\`, however it can be toggled via the drop-down menu.

Swivelgames:
[...]

PR #1056, opened 2017-03-31
Make the tree view a dock item

nathansobo:
⛑ This can be merged once 1.17 reaches stable. Tests are red because we're building against a version of Atom that doesn't have the dock APIs ⛑ 

screen shot 2017-03-31 at 5 00 31 pm

Hey, @simurai. I don't think we're really ready to ship this in the next release cycle because it needs some design ❤️.

In the screenshot above I deliberately sized the tree view so the tab stretched across 100% of its width. But when the tree view is wider, the narrower tab looks awkward:

screen shot 2017-03-31 at 5 34 57 pm

I also think the tab being white when the tree view isn't focused looks weird. Do you think you could come up with some mock ups, a PR on the tabs package, or just do some styling overrides on this branch for a treatment of tabs in docks that feels more natural? I don't think having a tab for the tree view is fundamentally bad. At least it vertically aligns the content of the tree view with the panes, which I think is nice. But I do think the dock tabs need to look somewhat different and more subtle than the pane tabs.


nathansobo:
Another option could be to do not render the tabs at the top for the left and right panels, but instead render them at the left and the right as icons. This might work better because these spaces are fairly narrow. I guess it depends on how many tabs we try to fit in. At the moment, in the default configuration, we only have the tree view so it doesn't matter much. The real problem is making the default configuration not look awkward while still making room for people to add more items.

simurai:
Just merged https://github.com/atom/one-light-ui/pull/96. So if you build `atom/atom` you should see it like this (plus the docks border):

screen shot 2017-04-01 at 11 22 42 am

Single tabs are now pretty seamless and look more like titles. Just once you start to add more items, they feel like tabs:

screen shot 2017-04-01 at 11 28 37 am

nathansobo:
Looks amazing. I'm even more excited about this feature now.

/cc @maxbrunsfeld @matthewwithanm

nathansobo:
BTW I think we should make this title-bar free mode the default.

Alhadis:
Where's the option to disable this?

matthewwithanm:
Looks amazing @simurai!! 😍

nathansobo:
@Alhadis I'm considering leaving an option to render the tree view in a panel instead of a dock. It's a bit more code complexity but could be a nice escape hatch for people who only want a tree view at the left and don't plan on using the docks at all.

Alhadis:
Well that would be me, but I'm kinda hoping there'll be something that will leave the tree-view looking like it hasn't been touched at all...

simurai:
You could add

```less
atom-dock.left .tab-bar {
  display: none;
}
```

then it feels like the current panel.

simurai:
Ohh.. wait, even better. I forgot about the "Always Show Tab Bar" option:

no-tabs

matthewwithanm:
> @Alhadis I'm considering leaving an option to render the tree view in a panel instead of a dock.

Is this despite the setting @simurai mentioned?

matthewwithanm:
> @Alhadis I'm considering leaving an option to render the tree view in a panel instead of a dock.

@nathansobo Is this despite the setting @simurai mentioned?

matthewwithanm:
imo it should be docks-only and any tweaks we wanna allow should be done within that framework. for example, if the existing "Always Show Tab Bar" option isn't enough, we could add one that's dock-specific. branching just makes it more complicated and i'm not sure what the gain is.

i'm a fan of the workspace items concept not just because devs don't have to duplicate logic (toggle side, etc) or because it gives users more control (since they can drag to new locations), but also because the centralization gives Atom more control. for example, Atom could add a mode to docks where they only show one item—the active one—and no tabs. or an option to always add new items in a split. the bottom line is that these things become consistent and predictable instead of having each one governed by a unique set of rules.

maxbrunsfeld:
omg @simurai that looks awesome.

nathansobo:
I think the setting covers it. Worst case there's a user style sheet if someone is upset about the toggle affordance.

simurai:
> @nathansobo Worst case there's a user style sheet if someone is upset about the toggle affordance.

The One themes already have a setting for that:

dock-buttons

For those that like a more minimal UI and are ok toggling with the keyboard only (or status-bar/tool-bar buttons).

simurai:
Should find-and-replace also be moved to a dock (bottom)? I think yes. Then the rule of thumb would be:

- **Larger** things that you want to show/hide "temporarily" -> **docks** (tree-view, find-and-replace, etc).
- **Smaller** things that you want to show "permanently" -> **panels** (status-bar, custom title-bar, tool-bar).

It could overlap, but then you can just keep a dock open all the time, or keep toggling a panel (like tool-bar).

maxbrunsfeld:
We can't merge this until Atom 1.17.0 is released, because it depends on the new Docks feature. But we want this to ship with 1.17.0-beta0 so we've cut a prerelease off of this branch and we're using that release in Atom master.

Alhadis:
@maxbrunsfeld How do you recommend I handle merge conflicts with #1009?

smashwilson:
Looks like the selectors in `menus/tree-view.cson` were out of date; the commit I pushed brings them back into alignment.

50Wliu:
I'm getting `IntersectionObserver.observe(target): target element is not a descendant of root.` warnings that point to Tree View.

nathansobo:
@50Wliu Apparently this warning has been removed in newer versions of Chrome. At this point I'm not sure it's worth worrying about it.

50Wliu:
This fixes #68

50Wliu:
Pending pane items aren't working:
pending-pane-items-broken

fabianfiorotto:
What is going to happen with the config `showOnRightSide`? I see that is deleted from the config form but you are still using it in the code

nathansobo:
@fabianfiorotto We still honor that config setting on package activation because we don't want to flip the tree view back to the left for users that have chosen to put it on the right when we upgrade. Going forward, however we want to use the drag-and-drop behavior afforded by docks rather than custom configuration.

Alhadis:
Looks like we can merge this now. =)

nathansobo:
❤️

PR #1057, opened 2017-04-01
:art: refactor getFullExtension function

bobintornado:
Make getFullExtension shorter, simpler


bobintornado:
The tests shouldn't pass on windows. But it did. Not sure why.

PR #1059, opened 2017-04-01
:art: refactor getFullExtension function

bobintornado:
Make getFullExtension shorter, simpler, easier to read and reasoning.

### DIFF
--- a/lib/helpers.coffee
+++ b/lib/helpers.coffee
@@ -17,11 +17,9 @@ module.exports =
     styleObject
 
   getFullExtension: (filePath) ->
-    fullExtension = ''
-    while extension = path.extname(filePath)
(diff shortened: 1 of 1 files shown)


thomasjo:
Unless this helper function is indirectly covered by [tree-view-spec.coffee#L1585](https://github.com/atom/tree-view/blob/e7b94519f68fd724ab1f9613f3e6d72302ccd313/spec/tree-view-spec.coffee#L1585), we're going to require some tests that ensure the original functionality remains unchanged after merging this PR.

/cc @atom/maintainers

50Wliu:
Yes, that's the indirect test (I was the one that wrote this helper).  It shouldn't be too hard to write a test directly covering this behavior.

PR #1060, opened 2017-04-01
:arrow_up: upgrade event-kit, minimatch and temp

bobintornado:
Upgrade event-kit, minimatch and temp

### DIFF
--- a/package.json
+++ b/package.json
@@ -10,11 +10,11 @@
   },
   "private": true,
   "dependencies": {
-    "event-kit": "^1.0.0",
+    "event-kit": "^2.2.0",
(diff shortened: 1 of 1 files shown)


50Wliu:
Why?

bobintornado:
Just based on feedback from https://david-dm.org/atom/tree-view
And thought later versions could bring in better stuff.

50Wliu:
If you don't have a concrete explanation for how this PR improves tree-view or fixes existing bugs, I'm sorry to say that it most likely will not be accepted.

50Wliu:
Closing as the only dependency left on this list is minimatch, which we can look into separately.

PR #1089, opened 2017-05-04
Make list inherit its container's width

### LINKED ISSUE #1075: Clicking on empty space behavior is unexpected in 0.217.0-5
### Prerequisites

* [x] Put an X between the brackets on this line if you have done all of the following:
    * Reproduced the problem in Safe Mode: http://flight-manual.atom.io/hacking-atom/sections/debugging/#using-safe-mode
    * Followed all applicable steps in the debugging guide: http://flight-manual.atom.io/hacking-atom/sections/debugging/
    * Checked the FAQs on the message board for common solutions: https://discuss.atom.io/c/faq
    * Checked that your issue isn't already filed: https://github.com/issues?utf8=✓&q=is%3Aissue+user%3Aatom
    * Checked that there is not already an Atom package that provides the described functionality: https://atom.io/packages

### Description

Clicking on empty space about tree view items has two different behaviors:

- If it's the active item, it's toggled (for dirs)
- If it's text width is equal to or less than the name of longest time, it's selected and toggled as expected
- If it's text width is greater than the longest name and isn't the active item, clicking on empty space in front of it has **zero** effect

### Steps to Reproduce

1. Open any repository
2. Click on the empty space in front of the name of a dir/file that's not active and is the longest

**Expected behavior:** [What you expect to happen]

The item is selected or toggled regardless

**Actual behavior:** [What actually happens]

Mentioned above

**Reproduces how often:** [What percentage of the time does it reproduce?]

Everytime

### Versions

```
Atom    : 1.17.0-beta3
Electron: 1.3.14
Chrome  : 52.0.2743.82
Node    : 6.5.0
```

Tree-View v0.217.0-5

### Additional Information

I'm on macOS if it matters, also here's a gif

kapture 2017-04-20 at 22 54 09

maxbrunsfeld:
Fixes #1075

/cc @nathansobo - What do you think of my new implementation of `getPreferredWidth`? We talked about doing it this way, but I'm not sure if there's a better way to compute an element's min-content width value.

### DIFF
--- a/lib/tree-view.coffee
+++ b/lib/tree-view.coffee
@@ -112,7 +112,10 @@ class TreeView
   isPermanentDockItem: -> true
 
   getPreferredWidth: ->
-    @list.offsetWidth
+    @list.style.width = 'min-content'
+    result = @list.offsetWidth
+    @list.style.width = ''
+    result
 
   onDirectoryCreated: (callback) ->
     @emitter.on('directory-created', callback)
--- a/styles/tree-view.less
+++ b/styles/tree-view.less
@@ -23,7 +23,6 @@
      * auto-created layer.
      */
     isolation: isolate;
-    width: min-content;
     padding-left: @component-icon-padding;
     padding-right: @component-padding;
     background-color: inherit;


nathansobo:
This looks great.

PR #1115, opened 2017-05-22
fix unselect multiple entiries on right click

UziTech:
### Description of the Change

fixes unselecting multiple entries when right clicking on a selected entry.

screenshot

### DIFF
--- a/spec/tree-view-spec.js
+++ b/spec/tree-view-spec.js
@@ -48,9 +48,19 @@ describe('TreeView', () => {
     it('should leave multiple entries selected on right click', () => {
       const treeView = new TreeView({})
       const entries = treeView.roots[0].entries
-      treeView.selectEntry(entries.children[0])
-      treeView.selectMultipleEntries(entries.children[1])
-      treeView.showMultiSelectMenu()
+
+      treeView.onMouseDown({
+        stopPropagation() {},
+        target: entries.children[0],
+        button: 0,
+      })
+
+      treeView.onMouseDown({
+        stopPropagation() {},
+        target: entries.children[1],
+        button: 0,
+        metaKey: true,
+      })
 
       let child = entries.children[0];
       while (child.children.length > 0) {
@@ -60,7 +70,7 @@ describe('TreeView', () => {
       treeView.onMouseDown({
         stopPropagation() {},
         target: child,
-        button: 2
+        button: 2,
       })
 
       expect(treeView.getSelectedEntries().length).toBe(2);
--- a/spec/tree-view-spec.js
+++ b/spec/tree-view-spec.js
@@ -53,7 +53,9 @@ describe('TreeView', () => {
       treeView.showMultiSelectMenu()
 
       let child = entries.children[0];
-      while (child.children.length > 0 && (child = child.firstChild));
+      while (child.children.length > 0) {
+        child = child.firstChild;
+      }
 
       treeView.onMouseDown({
         stopPropagation() {},
@@ -62,6 +64,7 @@ describe('TreeView', () => {
       })
 
       expect(treeView.getSelectedEntries().length).toBe(2);
+      expect(treeView.multiSelectEnabled()).toBe(true);
     })
   });
 })
--- a/spec/tree-view-spec.js
+++ b/spec/tree-view-spec.js
@@ -43,4 +43,25 @@ describe('TreeView', () => {
       )
     })
   })
+
+  describe('clicking', () => {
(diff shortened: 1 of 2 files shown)


50Wliu:
Nice.  Can you please add a spec for this?

UziTech:
When ever I try to run the tests on my machine I get

`Error: Cannot find module '../build/Release/pathwatcher.node'`

50Wliu:
Did you `apm install` instead of `npm install`?

UziTech:
That worked thank you

UziTech:
Any word on this? @50Wliu is there a reason the `needs-testing` flag is still on this?

50Wliu:
Sorry for the delay.  `needs-testing` can serve different purposes - I use it to denote that a feature needs to be tested (as in, by interacting with the change in Atom), not that it needs specs.

Finishing up my last few language-php PRs and working on a language-html PR, then I'll probably get to this.

UziTech:
sounds good. thanks

50Wliu:
There's some styling issues going on when multi-selecting: the Tree View gets pushed slightly to the left.  Looks like a lot of the styles only target `.full-menu` and not `.multi-select`.

|Full Menu|Multi Select|
|-----------|-------------|
|full-menu|multi-select|

UziTech:
it seems like just using `.list-tree` would work better than `.full-menu, .multi-select`. That way if any other state class was added in the future we wouldn't run into the same issue.

UziTech:
never mind `.list-tree` seems to be on any expanded folder. A new class should probably be added.

simurai:
Ouch.. :weary: Didn't realize that `.full-menu` is more like a "state". Added a new `.tree-view-root` class in https://github.com/atom/tree-view/pull/1130.

spec/tree-view-spec.js
```diff
@@ -43,4 +43,25 @@ describe('TreeView', () => {
...
+      let child = entries.children[0];
+      while (child.children.length > 0 && (child = child.firstChild));
+
+      treeView.onMouseDown({
+        stopPropagation() {},
+        target: child,
+        button: 2
+      })
+
+      expect(treeView.getSelectedEntries().length).toBe(2);
```
50Wliu: How about another assertion to make sure that the `multi-select` class is present?

spec/tree-view-spec.js
```diff
@@ -43,4 +43,25 @@ describe('TreeView', () => {
...
+  describe('clicking', () => {
+    it('should leave multiple entries selected on right click', () => {
+      const treeView = new TreeView({})
+      const entries = treeView.roots[0].entries
+      treeView.selectEntry(entries.children[0])
+      treeView.selectMultipleEntries(entries.children[1])
+      treeView.showMultiSelectMenu()
+
+      let child = entries.children[0];
+      while (child.children.length > 0 && (child = child.firstChild));
```
50Wliu: Can this while loop be expanded?

50Wliu:
Going to block this on #1130 to ensure that we don't forget about that.

mrveress:
Hello everyone! Sorry for the question: when this functionality be in the new release? I waiting this almost 3 month and press context button on the keyboard (my context key almost destroyed because of this 😄 )

UziTech:
@mrveress kind of hacky but I am using this in atom now by doing the following steps:

1. Fork this branch
2. Change the `name` property in `package.json` to "tree-view-temp" (or anything that is not the name of another package)
3. Disable the core "tree-view" package in atom
4. Install the forked package with the github url

Just make sure to uninstall that package and enable "tree-view" when it ships with this merged.

simurai:
FYI: PR #1130 is now merged.

50Wliu:
@UziTech please rebase and then I'll give this a final test run.

UziTech:
All test passed 👍 💯

spec/tree-view-spec.js
```diff
@@ -43,4 +43,28 @@ describe('TreeView', () => {
...
     })
   })
+
+  describe('clicking', () => {
+    it('should leave multiple entries selected on right click', () => {
+      const treeView = new TreeView({})
+      const entries = treeView.roots[0].entries
+      treeView.selectEntry(entries.children[0])
+      treeView.selectMultipleEntries(entries.children[1])
+      treeView.showMultiSelectMenu()
```
50Wliu: To avoid explicitly stating this in the spec, is it possible to do a `treeView.onMouseDown` event on `entries.children[1]`?

50Wliu:
:+1:

PR #1130, opened 2017-06-16
Fix multi-select shifting

simurai:
### Description of the Change

This fixes the tree-view from shifting to the left when selecting multiple files.

|Full Menu|Multi Select|
|-----------|-------------|
|full-menu|multi-select|

### Alternate Designs

We could use a child selector like `.tree-view > .list-tree` or both "states" `.full-menu, .multi-select`, but having just a single class seems simpler and communicates its intend better.

### Benefits

No shifting.

### Possible Drawbacks

None

### Applicable Issues

Raised in https://github.com/atom/tree-view/pull/1115#issuecomment-308879413
Possibly regressed in https://github.com/atom/tree-view/pull/1111

### DIFF
--- a/lib/tree-view.coffee
+++ b/lib/tree-view.coffee
@@ -31,7 +31,7 @@ class TreeView
     @element.tabIndex = -1
 
     @list = document.createElement('ol')
-    @list.classList.add('full-menu', 'list-tree', 'has-collapsable-children', 'focusable-panel')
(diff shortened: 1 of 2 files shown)


50Wliu:
Thanks!

50Wliu:
@simurai is this mergeable?

simurai:
> is this mergeable?

Yes, I think so. I'll merge it. 🚢

Issue #903, opened 2016-08-01
Empty label for drive paths (e.g. Z:\) under Windows


magv:
When you add a project folder like `Z:\` under Windows (e.g. a local or a network mapped disk), tree view gives it an empty label. I suspect using path `/` on Unix will result in similar behavior.

This problem originates in `updateRoots` in [lib/tree-view.coffee](/atom/tree-view/blob/ced5f95b75abf1b04b705e498728a6ca1c73a74c/lib/tree-view.coffee#L279), where directory's name is defined as `path.basename(projectPath)`. Since `path.basename("C:\\")` is an empty string, this value propagates into `DirectoryView` you get an empty label as the result.

A change like this should fix the situation:
```diff
--- tree-view.coffee.orig	2016-08-01 13:46:00 +0300
+++ tree-view.coffee	2016-08-01 13:46:02 +0300
@@ -276,7 +276,7 @@
 
     @roots = for projectPath in atom.project.getPaths()
       directory = new Directory({
-        name: path.basename(projectPath)
+        name: path.basename(projectPath) or projectPath
         fullPath: projectPath
         symlink: false
         isRoot: true
```

... but any other solution would be welcome as well, because anything is better than an empty label.

liuderchi:
> I suspect using path `/` on Unix will result in similar behavior too.

Yes. I did some other tests on Ubuntu, Win7:

| `projectPath`               | `path.basename(projectPath)` | os     | note                |
|-----------------------------|------------------------------|--------|---------------------|
| `/`                         | *EMPTY*                      | Ubuntu | root dir            |
| `D:`                        | *EMPTY*                      | win7   | local disk          |
| `\\10.90.7.16\share`        | *EMPTY*                      | win7   | shared folder       |
| `\\10.90.7.16\share\common` | `common`                     | win7   | inner shared folder |

found that *inner shared folder* works as expected

magv:
Are you sure about `\\10.90.7.16\share`? It seems to work for me.

liuderchi:
@magv that was my mistake. Fixed.

riemers:
I came from the issue #700 , so how can i simply get those changes above into the tree-view package? Since the original are coffee files i presume i need to edit these first then do some building process that makes the .js files then alter the package in atom.  I don't know if above is sufficient to solve, otherwise i can just make a PR for that and someone can merge it? I don't know what is holding this change back if it looks in my eyes that simple?

magv:
[...]

Issue #700, opened 2016-01-14
Uncaught TypeError: Path must be a string. Received null


AlinV12:
[Enter steps to reproduce below:]

1. ...
2. ...

**Atom Version**: 1.4.0
**System**: Ubuntu 14.04.3
**Thrown From**: [tree-view](https://github.com/atom/tree-view) package, v0.198.0

### Stack Trace

Uncaught TypeError: Path must be a string. Received null

```
At path.js:8

TypeError: Path must be a string. Received null
    at assertPath (path.js:8:11)
    at Object.posix.join (path.js:477:5)
    at MoveDialog.module.exports.MoveDialog.onConfirm (/home/alin/usr/share/atom/resources/app.asar/node_modules/tree-view/lib/move-dialog.js:38:24)
    at space-pen-div.atom.commands.add.core:confirm (/home/alin/usr/share/atom/resources/app.asar/node_modules/tree-view/lib/dialog.js:48:26)
    at CommandRegistry.module.exports.CommandRegistry.handleCommandEvent (/home/alin/usr/share/atom/resources/app.asar/src/command-registry.js:260:29)
    at /home/alin/usr/share/atom/resources/app.asar/src/command-registry.js:3:61
    at KeymapManager.module.exports.KeymapManager.dispatchCommandEvent (/home/alin/usr/share/atom/resources/app.asar/node_modules/atom-keymap/lib/keymap-manager.js:536:16)
    at KeymapManager.module.exports.KeymapManager.handleKeyboardEvent (/home/alin/usr/share/atom/resources/app.asar/node_modules/atom-keymap/lib/keymap-manager.js:359:22)
    at WindowEventHandler.module.exports.WindowEventHandler.handleDocumentKeydown (/home/alin/usr/share/atom/resources/app.asar/src/window-event-handler.js:97:36)
    at HTMLDocument.<anonymous> (/home/alin/usr/share/atom/resources/app.asar/src/window-event-handler.js:3:61)
```

### Commands

```
     -0:51.4.0 core:move-right (atom-text-editor.editor.is-focused)
     -0:51.2.0 core:move-down (atom-text-editor.editor.is-focused)
  3x -0:51 core:move-left (atom-text-editor.editor.is-focused)
  2x -0:49.9.0 core:backspace (atom-text-editor.editor.is-focused)
     -0:48.9.0 core:move-up (atom-text-editor.editor.is-focused)
  4x -0:48.7.0 core:move-right (atom-text-editor.editor.is-focused)
[...]

PR #1140, opened 2017-06-29
Fix empty label for drive paths

riemers:
### Description of the Change

Fix empty label for drive paths as stated in the issue. https://github.com/atom/tree-view/issues/903#issuecomment-311930641

### Alternate Designs

Issue been open for a long time, no other solution at this moment.

### Benefits

Not having a . as label in the project root.

### Possible Drawbacks

None that i am aware off.

### Applicable Issues

https://github.com/atom/tree-view/issues/903 partial https://github.com/atom/tree-view/issues/700

### DIFF
--- a/lib/tree-view.coffee
+++ b/lib/tree-view.coffee
@@ -282,7 +282,7 @@ class TreeView
         stats[key] = stats[key].getTime()
 
       directory = new Directory({
-        name: path.basename(projectPath)
+        name: path.basename(projectPath) or projectPath
         fullPath: projectPath
         symlink: false
         isRoot: true


50Wliu:
Can you add a spec for this please?

riemers:
To be honest, i don't know what you mean. I'am just a guy that made a pull request from all the comments concerning the issues from this PR that should resolve the issue from comments with verified results. A link to what needs to be done, if i can, i will surely update!

50Wliu:
Sure thing.  By "specs" I mean tests to make sure that this behavior doesn't regress again.  Those tests are contained in https://github.com/atom/tree-view/blob/master/spec/tree-view-package-spec.coffee.  For a brief rundown on how specs work, see http://flight-manual.atom.io/hacking-atom/sections/writing-specs/.

riemers:
Ah, those kinds of specs. I understand them, and i do see the need for it. Problem is that i was just trying to help by creating a PR. I don't know all the ins & outs in that language too. I do know that it would be something like:

describe: Foldername should always have proper root slash
it: expect(path.basename(projectPath)).toEqual("/")

But that is just from reading the howto's. I don't have a build environment either. (not lazy, just don't know enough) and to top it all off, there is even a chance this won't fix my issue of the network drive. Feel free to close the PR if needed (as i can imagine, not everybody wants to write specs for other people's PR's)

50Wliu:
Closing because this doesn't fix the problem; see https://github.com/atom/tree-view/issues/903#issuecomment-328292794.

PR #1144, opened 2017-07-05
Repo status color on project root directory

captin411:
This makes it easier to identify which projects have uncommitted
changes when working with more than one project.

### Requirements

### Description of the Change

I occasionally work with software projects that have more than one project folder added to them.

In these cases, it is difficult to identify which repositories I have uncommitted changed on.

This change bubbles up the '.status-modified' and '.status-added' CSS classes to the '.project-root' elements so that the repo modification status can be clearly seen when all folders are collapsed.

### Alternate Designs

The code changes in the `directory.coffee` file could alternatively be put into the atom core.  It would require a change to the logic of 'repo.getDirectoryStatus()'  method.

The reason I avoided changin Atom core are twofold:
- first is my concern with unexpected consequences for other plugin developers who use the repo.getDirectoryStatus (is the root folder coming back with a status of 0 a feature or a bug?)
- second, an update here allows people to get this new feature by updating the plugin instead of all of Atom

In short, there is a bug(?) in repo.getDirectoryStatus('/my/root') which will calculate the relative path as @relativize("/my/root") + "/" resulting in the string "/".  Then all of the git statuses (formatted like "subdir/file.txt") are evaluated and if statusPath.indexOf(relative_git_root) == 0 then it will consider the status of the parent folder to be of a particular value.

The problem is that, in the case of the root folder, it will never match character 0 since none of the status paths ever have a leading "/".

### Benefits

People who want to see the repo status at the root level will be able to do so - especially useful for those who use multi-project often.

### Possible Drawbacks

It is aesthetic and people are opinionated -- so some people might not like it.

### Applicable Issues

n/a

### DIFF
--- a/lib/directory-view.coffee
+++ b/lib/directory-view.coffee
@@ -56,8 +56,9 @@ class DirectoryView
       @header.classList.add('project-root-header')
     else
       @element.draggable = true
-      @subscriptions.add @directory.onDidStatusChange => @updateStatus()
-      @updateStatus()
+
+    @subscriptions.add @directory.onDidStatusChange => @updateStatus()
+    @updateStatus()
 
     @expand() if @directory.expansionState.isExpanded
 
--- a/lib/directory.coffee
+++ b/lib/directory.coffee
@@ -93,14 +93,14 @@ class Directory
     else
       status = 0
       if @isRoot
-          # repo.getDirectoryStatus will always fail for the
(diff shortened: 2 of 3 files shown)


captin411:
Here is an example of the final result:

screen shot 2017-07-04 at 15 19 25

daviwil:
Looks great, thanks a lot @captin411!

Issue #1066, opened 2017-04-11
Need "Sort Files Before Folder" option


arfordweb:
### Prerequisites

* [x] Put an X between the brackets on this line if you have done all of the following:
    * Reproduced the problem in Safe Mode: http://flight-manual.atom.io/hacking-atom/sections/debugging/#using-safe-mode
    * Followed all applicable steps in the debugging guide: http://flight-manual.atom.io/hacking-atom/sections/debugging/
    * Checked the FAQs on the message board for common solutions: https://discuss.atom.io/c/faq
    * Checked that your issue isn't already filed: https://github.com/issues?utf8=✓&q=is%3Aissue+user%3Aatom
    * Checked that there is not already an Atom package that provides the described functionality: https://atom.io/packages

I put an X above because this is a feature modification request, not a bug.  I'm fairly certain I understand the existing functionality and what would need to change to get the desired functionality.

### Description
I am currently on version 0.214.1 of tree-view on Mac.

Right now, there are 2 options available for sorting folders and files.

1. Sort Folders Before Files CHECKED
  - Keeps folders and files grouped with each other instead of intermingled
![screen shot 2017-04-11 at 1 30 28 am](https://cloud.githubusercontent.com/assets/1002408/24894186/94fd9318-1e56-11e7-9356-93db2744719f.png)
2. Sort Folders Before Files UNCHECKED
  - Folders and files are in alphabetical order and not grouped, so folders and files are mixed in together.
![screen shot 2017-04-11 at 1 30 16 am](https://cloud.githubusercontent.com/assets/1002408/24894193/a071a234-1e56-11e7-92a5-a4171da617f6.png)

You select between these 2 options with a checkbox, making it an either/or choice.  But a nice third option would be to sort files before folders.  This could be achieved with a dropdown containing 3 options instead of 2.  Or maybe a second checkbox that is disabled when the first is selected.

#### My Use Case ####
I use a folder hierarchy to keep up with a hierarchy of components and sub-components in a UI I'm building.  In doing so, my components contain a few standard files and potentially lots of sub-folders.  When these sub-folders are expanded, it's visually difficult to tell which set of files belong to which of the expanded folders above.

I'd love for the files to be sorted before the folders since the files are associated with the folder that contains them, and would make better sense located directly under the containing filename.

lee-dohm:
Thanks for the submission.

[...]

PR #1152, opened 2017-07-24
Sort Files Before Folder

rye761:
### Requirements

### Description of the Change

These changes make it possible to sort all files before all folders in the tree view. Previously, it was only possible to sort folders before files or mix folders with files. Migration code has been added so that users who have set the old sortFoldersBeforeFiles option will see their preferences reflected in the sortFolders option after updating.

### Alternate Designs

A dual checkbox design was considered that wouldn't have changed the sortFoldersBeforeFiles option, but a dropdown was used instead for simplicity.

### Benefits

Easier for some users to find their files in a sea of folders.

### Possible Drawbacks

None.

### Applicable Issues

 This additional option was requested in issue #1066.

### DIFF
--- a/lib/tree-view-package.js
+++ b/lib/tree-view-package.js
@@ -7,6 +7,16 @@ const TreeView = require('./tree-view')
 module.exports =
 class TreeViewPackage {
   activate () {
+    const oldSort = atom.config.get('tree-view.sortFoldersBeforeFiles')
+    if (oldSort !== undefined) {
+      console.log('Migrating settings....')
+      atom.config.unset('tree-view.sortFoldersBeforeFiles')
+      if (oldSort) {
+        atom.config.set('tree-view.sortFolders', 'Before Files')
+      } else {
+        atom.config.set('tree-view.sortFolders', 'Alongside Files')
+      }
+    }
     this.disposables = new CompositeDisposable()
     this.disposables.add(atom.commands.add('atom-workspace', {
       'tree-view:show': () => this.getTreeViewInstance().show(),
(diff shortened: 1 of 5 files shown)


package.json
```diff
@@ -48,10 +48,15 @@
...
-      "description": "When listing directory items, list subdirectories before listing files."
+    "sortFolders": {
+      "type": "string",
+      "default": "Before Files",
+      "description": "When listing directory items, where subdirectories will be listed.",
+      "enum": [
+        "Before Files",
+        "After Files",
+        "Alongside Files"
+      ]
```
thomasjo: This definition should be changed to something along the lines of
``` json
"enum": [
  { "value": "before", "description": "Before Files" },
  { "value": "after", "description": "After Files" },
  { "value": "alongside", "description": "Alongside Files" }
]
```
This will obviously also mean changing the current string comparisons throughout this PR.

lib/tree-view-package.js
```diff
@@ -7,6 +7,16 @@ const TreeView = require('./tree-view')
...
+    const oldSort = atom.config.get('tree-view.sortFoldersBeforeFiles')
+    if (oldSort !== undefined) {
+      console.log('Migrating settings....')
+      atom.config.unset('tree-view.sortFoldersBeforeFiles')
+      if (oldSort) {
+        atom.config.set('tree-view.sortFolders', 'Before Files')
+      } else {
+        atom.config.set('tree-view.sortFolders', 'Alongside Files')
+      }
+    }
```
thomasjo: This migration code should be extracted as a separate function (or method), and invoked here.

lee-dohm:
Thanks very much for your contribution and the effort you put into this.

Upon further reflection, we've decided that we aren't going to add any more sorting options for the Tree View. My apologies for not catching this and the associated issue sooner.

Issue #996, opened 2016-11-30
Introduce option for disabling moving files and/or folders 


jbergstroem:
I've more than once moved a folder into another by mistake. Having an option that disables similar options in the tree view would be awesome!

50Wliu:
Thanks for taking the time to contribute!

We noticed that this is a duplicate of #588. In line with the comments on #623, we're going to focus on improving the UX of drag-and-drop as the solution, rather than a setting to disable this functionality.

Because we treat our issues list as [the Atom team's backlog](https://en.wikipedia.org/wiki/Scrum_(software_development)#Product_backlog), we close duplicates to focus our work and not have to touch the same chunk of code for the same reason multiple times. This is also why we may mark something as duplicate that isn't an exact duplicate but is closely related.

For information on how to use GitHub's search feature to find out if something is a duplicate before filing, see [the How Can I Contribute? section of the Atom CONTRIBUTING guide](https://github.com/atom/atom/blob/master/CONTRIBUTING.md#how-can-i-contribute).

benyanke:
>  we're going to focus on improving the UX of drag-and-drop as the solution, rather than a setting to disable this functionality.

Why not both? One is a short term fix which has little impact on those who don't want it, and the other is a longer-term feature add which can peacefully coexist with the shortterm fix.

benyanke:
Since the Atom team seems firm about refusing to add this functionality....

I'm new to atom packages, but if it gets an install base, I'd happy to keep this very simple fork up to date, with a hardcoded disable of drag and drop. In any case, it's working on the current version for current use.

I'll keep an eye on the download counts in the atom package store to see if I decide to go with ongoing maintenance (which could perhaps even be scripted, since it's a search/replace for true/false).

https://github.com/benyanke/tree-view-no-drag
https://atom.io/packages/tree-view-no-drag

benyanke:
Note, to those who followed this thread:

> Given that we're not going to take on significant work on the tree view any time soon, adding this checkbox, while less than ideal, feels like the right thing now.
> I can't re-open this PR because the source repo is gone, but we'd accept a good PR adding this. I'll reopen #588.

[source](https://github.com/atom/tree-view/pull/623#issuecomment-330957065)

Issue #919, opened 2016-08-10
Add drag and drop threshold setting (or use system default)


strindhaug:
The drag and drop theshold is ridiculous, it appears to be essentially 0px (or 1-2px; it's hard to tell on my high PPI monitor), meaning I cannot click in the tree view without risking accidentally dragging and dropping files or folders. (There is no undo for this either with makes this even more problematic). 

On my system (Ubuntu) in most native apps such as the filemanager the drag theshold is 50px it's stored in dconf under `org.gnome.settings-daemon.peripherals.mouse.drag-threshold` as an integer, presumably this setting is available to all apps. But (assumning this setting *is* available to applications) Atom doesn't use this.

There must be a way to change this threshold to something bigger (preferably without any upper limit, so that we can effectively disable it since having a setting to disable drag and drop was wontfixed: https://github.com/atom/tree-view/pull/623 ).

lee-dohm:
In order to change this, we would have to either fiddle with some low-level Chromium bits or create a complete architecture for how to do drag-and-drop in Atom. Both are a bit beyond what we have the resources available to do at this point. We may revisit this in the future when we have more resources available. If we do, we'll reopen this Issue.

strindhaug:
@lee-dohm so, you're saying this really is a Chromium issue? Actually it seems it may be, just noticed that the tabs in Chrome also starts dragging without a threshold. Maybe I should report this upstream to the chromium project...

pwnall:
Are you guys building your own Chromium? If so, you could patch the values in `third_party/WebKit/Source/core/input/MouseEventManager.cpp`.

FWIW, 50px strike me as excessive. The default values for the OS preferences I've looked at are 3px-8px. Also, AFAIK these are CSS pixels, not actual pixels on a HiDPI screen.

Last, Chromium has [a bug for this issue](http://crbug.com/653490). Unfortunately, based on the current bug popularity vs implementation complexity, I don't expect we'll be able to address it any time soon.

lee-dohm:
:wave: @pwnall The Atom team isn't building our own Chromium, no. We get it as part of [Electron](https://github.com/electron/electron) and use it as-is (though theoretically Electron could do as you suggest if it would make sense for _all_ Electron applications. The Electron team tries to avoid adding more divergences from base Chromium, especially in tricky parts of the code, because then that creates maintenance overhead as we go down the line.

Thanks very much for the advice and the update! I'll pass it along to the Electron team :grinning:

PR #1154, opened 2017-08-01
Toggle drag and drop with env variable

strindhaug:
Minimal fix for #566 only using environment variable.
As suggested by @warren-bank in https://github.com/atom/tree-view/issues/566#issuecomment-319284053

### Description of the Change

Does the minimum needed change to make drag and drop disableable by optionally setting the environment variable `atom_tree_view_disable_drag_drop`.

- defaults to: true
- to disable globally, append to ~/.bashrc:
 `export atom_tree_view_disable_drag_drop='1'`

- to disable for a single run:
  `(export atom_tree_view_disable_drag_drop='1'; atom)` 

- to re-enable for a single run: 
 `(export atom_tree_view_disable_drag_drop=''; atom)`

### Alternate Designs

There have been other suggested fixes such as https://github.com/atom/tree-view/pull/623 and https://github.com/atom/tree-view/pull/748 that were closed as _wontfix_ because it added configuration.

### Benefits

- Makes users who do not want drag and drop, able to disable this without having to maintain their own fork.
- This fix does not add any visible settings; thus no clutter.
- It only changes two hardcoded `true`'s into `(not process.env.atom_tree_view_disable_drag_drop)` which also defaults to `true`; thus not adding measurably to cost of maintenance.

### Possible Drawbacks

None unless the user sets the `atom_tree_view_disable_drag_drop` variable.

### Applicable Issues
Some related issues: https://github.com/atom/tree-view/issues/566 https://github.com/atom/tree-view/issues/996 https://github.com/atom/tree-view/issues/919

### DIFF
--- a/lib/directory-view.coffee
+++ b/lib/directory-view.coffee
@@ -55,7 +55,7 @@ class DirectoryView
       @element.classList.add('project-root')
       @header.classList.add('project-root-header')
     else
-      @element.draggable = true
+      @element.draggable = (not process.env.atom_tree_view_disable_drag_drop)
       @subscriptions.add @directory.onDidStatusChange => @updateStatus()
       @updateStatus()
 
--- a/lib/file-view.coffee
+++ b/lib/file-view.coffee
@@ -9,7 +9,7 @@ class FileView
 
     @element = document.createElement('li')
     @element.setAttribute('is', 'tree-view-file')
-    @element.draggable = true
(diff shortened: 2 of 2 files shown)


lee-dohm:
As I mentioned in [my response in the issue](https://github.com/atom/tree-view/issues/566#issuecomment-319347602), this isn't something that the Atom team is going to accept. Thank you for your effort and your interest in Atom.

Issue #1155, opened 2017-08-01
Flaky test - 'it adds a file, closes the dialog, selects the file in the tree-view ...'


maxbrunsfeld:
This failed on Circle on Atom's 1.19-releases branch:

https://circleci.com/gh/atom/atom/4476

```
Package tests failed for tree-view:
.....................................................................................................................................................................................................................................................F.................................................................--....................................................

TreeView
  file modification
    tree-view:add-file
      when a file is selected
        when the path without a trailing '/' is changed and confirmed
          when no file exists at that location
            it adds a file, closes the dialog, selects the file in the tree-view, and emits an event
              Expected 'test-dirnew-test-file.txttest-file.txt' to be 'new-test-file.txt'.
                at jasmine.Spec.<anonymous> (/Users/distiller/atom/node_modules/tree-view/spec/tree-view-package-spec.coffee:1816:81)


Finished in 104.717 seconds
363 tests, 1892 assertions, 1 failure, 2 skipped
```

jasonrudolph:
We got another similar failure on Circle on Atom's 1.19-releases branch:

https://circleci.com/gh/atom/atom/4479

```
Package tests failed for tree-view:
......................................................................................................................................................................................................................................................F................................................................--....................................................

TreeView
  file modification
    tree-view:add-file
      when a file is selected
        when the path without a trailing '/' is changed and confirmed
          when no file exists at that location
            it adds file in any project path
              TypeError: Cannot read property 'textContent' of null
                at jasmine.Spec.<anonymous> (/Users/distiller/atom/node_modules/tree-view/spec/tree-view-package-spec.coffee:1841:72)

Finished in 105.645 seconds
363 tests, 1891 assertions, 1 failure, 2 skipped
```

This failure isn't _exactly_ the same as the one in the issue body, but both failures are occurring in the same nested context, so there's a decent chance that they're related.

PR #1156, opened 2017-08-01
Fix flaky tests re: adding new file

jasonrudolph:
Fixes https://github.com/atom/tree-view/issues/1155.

## TODO

- [x] Fix intermittent failure in "it adds a file, closes the dialog, selects the file in the tree-view, and emits an event" https://github.com/atom/tree-view/issues/1155#issue-247146435 

  Resolved in 55e1b25bc5315d60143859606523fd55cbef5f96. Please see the commit message for details.

- [x] Fix intermittent failure in "it adds file in any project path" https://github.com/atom/tree-view/issues/1155#issuecomment-319471558

  Resolved in 6bb70056fab19c0adae4aa627ec2dec6438cbaed. Please see the commit message for details.

### DIFF
--- a/spec/tree-view-package-spec.coffee
+++ b/spec/tree-view-package-spec.coffee
@@ -1814,6 +1814,7 @@ describe "TreeView", ->
 
               waitsFor "tree view selection to be updated", ->
                 treeView.element.querySelector('.file.selected')
+
               runs ->
                 expect(treeView.element.querySelector('.selected').textContent).toBe path.basename(newPath)
                 expect(callback).toHaveBeenCalledWith({path: newPath})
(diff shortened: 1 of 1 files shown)


maxbrunsfeld:
This seems reasonable to me. So there's some async step between adding the file to the tree view and selecting it? Were you able to reproduce the failure?

jasonrudolph:
@maxbrunsfeld: Thanks for the review. :bow:

> Were you able to reproduce the failure?

Yes. 😅

> So there's some async step between adding the file to the tree view and selecting it?

I think that's right. When a new file is created, we add it to the tree view, but we don't immediately select it in the tree view. As far as I can tell, the file doesn't get selected in the tree view until the new file's pane item is opened and Atom emits the "did-change-active-pane-item" event. [[code](https://github.com/atom/tree-view/blob/6bb70056fab19c0adae4aa627ec2dec6438cbaed/lib/tree-view.coffee#L181-L182)]

Issue #179, opened 2014-06-26
Rename root folder


Bengt:
tree-view can currently not rename the opened root folder. When I click the root folder in the tree view and press `F2` or select `right-click` - `Rename` nothing happens.

alexchandel:
This is still an issue in 1.36.1, and it shouldn't be.  Renaming the root folder is easily something Atom could do.

aldevel:
Yep, seemingly this very basic feature is still missing for a while now. Probably it's feasible by removing it obviously not from Atom and somehow making Atom to forget about it what I'm just trying to figure out somehow but it's definitely weird not having this option by default for a very long time now.

PR #1161, opened 2017-08-13
Project root does not need a rename operation.

yubaoquan:
### Requirements

### Description of the Change

Remove the `rename` operation in project root's context menu.

### Alternate Designs

Maintain that operation, and when user right click on an entry, first find out whether the entry is project's root, if it is, disable the menuItem, make it gray.
This alternate design will increase the complexity of code. 

### Benefits

The context menu of tree-view become less confusing to users. User can only see what they can do. And I guess when calling the function of rename, there is no need to judge whether the entry is a root.

### Possible Drawbacks

If the config in the selector `'.tree-view .full-menu .project-root > .header'` is not only used for root entry, the other places depends on this config item will be affected.

### DIFF
--- a/menus/tree-view.cson
+++ b/menus/tree-view.cson
@@ -54,7 +54,6 @@
     {'label': 'New Folder', 'command': 'tree-view:add-folder'}
     {'type': 'separator'}
 
-    {'label': 'Rename', 'command': 'tree-view:move'}
     {'label': 'Duplicate', 'command': 'tree-view:duplicate'}
     {'label': 'Delete', 'command': 'tree-view:remove'}
     {'label': 'Copy', 'command': 'tree-view:copy'}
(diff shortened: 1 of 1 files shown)


50Wliu:
There's some other commands that don't seem to be available on project folders: Duplicate, Delete, and Cut.

Ben3eeE:
Fixes https://github.com/atom/tree-view/issues/179?

yubaoquan:
Yes, removed useless menu items from root

sheddup:
Bump

ungb:
This doesn't seem to work as desired after testing this change:
| Before PR: | Testing this PR: |
| :------------- | :------------- |
|image | image
|

yubaoquan:
Maybe there is more complicated rules of showing items in context menu. And I don't know how to test a core atom package. It's my bad to think the change too easy. Sorry

Issue #433, opened 2015-04-28
Provide a service API


basarat:
Particularly interested in getting told about file/folder renames. 

Would make it easier for me to support folder / file move style refactoring for atom-typescript.

Would you guys be open to consider such a PR.

thomasjo:
The tree view uses `atom.project.onDidChangePaths` to handle renames etc, so not really sure if it makes sense to expose a service just to wrap existing functionality. Are there any other particular scenarios the service would/could cover outside of renames, or is there perhaps some reason for not using the existing event?

basarat:
> atom.project.onDidChangePaths

Exactly what I needed. I just didn't know about it :rose:

basarat:
Reopening as `atom.project.onDidChangePaths` is not designed to be triggered on tree view file renames (refs https://github.com/atom/tree-view/issues/437#issuecomment-97294630).

Might investigate adding something along similar lines though.

thomasjo:
The way this is done in the tree view right now, is that is recursively adds "change" listeners starting at the projects roots (where `Project::onDidChangePaths` comes into play). It uses it's own `Directory` and `File` implementations though, not really sure why they aren't extending the ones exposed by Atom.

I'm starting to see the usefulness of either a service API, or at least some emitted event.

basarat:
> at least some emitted event.

Basically needs to be emitted here : https://github.com/atom/tree-view/blob/master/lib/move-dialog.coffee#L35

basarat:
Opinion needed before I PR:

1. a new global `atom.treeview` OR
1. tag this on the `atom.project` OR 
1. something else

If its going to be on `atom.project` (i.e. 2.) then the design would become:

* `onConfirm` [from move dialog](https://github.com/atom/tree-view/blob/master/lib/move-dialog.coffee#L20) will trigger `atom.project.emitter.emit('did-rename-path')` (calling it by this name  because the command is `'tree-view:rename'`).
* Add a new function to atom/project [here](https://github.com/atom/atom/blob/4160a8c239f8a7ed54d2b6395c81a225748a8e98/src/project.coffee#L114-L115) called `onDidRenamePath`.

mnquintana:
I don't think a new global for Tree View makes sense since Tree View is a package and can be disabled.

basarat:
@mnquintana cool. What about `2`? or would you suggest something else.

zephraph:
[...]

PR #1162, opened 2017-08-13
Provide treeView service API

UziTech:
### Description of the Change

Provided a service that lets other packages consume the treeView instance.

At the moment any package that interacts with the tree-view package to get selected files or other information are doing so by getting the active package's `mainModule` or checking classes and data attributes on elements.

There is a standard way to get this information through the [Services API](http://flight-manual.atom.io/behind-atom/sections/interacting-with-other-packages-via-services/)

### Alternate Designs

We may want to change what actually get provided if there are unstable features in the tree-view module.

### Benefits

Other packages can interact with this package in a standard way.

### Possible Drawbacks

none

### Applicable Issues

#433

### DIFF
--- a/lib/tree-view-package.js
+++ b/lib/tree-view-package.js
@@ -54,7 +54,7 @@ class TreeViewPackage {
 
   provideTreeView () {
     return {
-      getTreeViewInstance: () => this.getTreeViewInstance()
+      getSelectedPaths: () => this.getTreeViewInstance().selectedPaths()
     };
   }
 
--- a/lib/tree-view-package.js
+++ b/lib/tree-view-package.js
@@ -52,6 +52,12 @@ class TreeViewPackage {
     )
   }
 
+  provideTreeView () {
+    return {
+      getTreeViewInstance: () => this.getTreeViewInstance()
+    };
+  }
+
   getTreeViewInstance (state = {}) {
     if (this.treeView == null) {
       this.treeView = new TreeView(state)
--- a/lib/tree-view-package.js
+++ b/lib/tree-view-package.js
@@ -54,7 +54,8 @@ class TreeViewPackage {
 
   provideTreeView () {
     return {
-      getSelectedPaths: () => this.getTreeViewInstance().selectedPaths()
+      getSelectedPaths: () => this.getTreeViewInstance().selectedPaths(),
+      getEntry: (path) => this.getTreeViewInstance().entryForPath(path)
     };
   }
 
--- a/lib/tree-view-package.js
+++ b/lib/tree-view-package.js
@@ -54,8 +54,8 @@ class TreeViewPackage {
 
   provideTreeView () {
     return {
-      getSelectedPaths: () => this.getTreeViewInstance().selectedPaths(),
-      getEntry: (path) => this.getTreeViewInstance().entryForPath(path)
+      selectedPaths: () => this.getTreeViewInstance().selectedPaths(),
+      entryForPath: (path) => this.getTreeViewInstance().entryForPath(path)
     };
   }
 
--- a/lib/tree-view-package.js
+++ b/lib/tree-view-package.js
@@ -55,7 +55,7 @@ class TreeViewPackage {
   provideTreeView () {
     return {
       selectedPaths: () => this.getTreeViewInstance().selectedPaths(),
-      entryForPath: (path) => this.getTreeViewInstance().entryForPath(path)
+      entryForPath: (entryPath) => this.getTreeViewInstance().entryForPath(entryPath)
     };
   }
 
--- a/package.json
+++ b/package.json
@@ -30,6 +30,14 @@
       }
     }
   },
+  "providedServices": {
+    "tree-view": {
+      "description": "Get treeView Instance",
+      "versions": {
+        "1.0.0": "provideTreeView"
+      }
+    }
+  },
   "configSchema": {
     "squashDirectoryNames": {
       "type": "boolean",
--- a/package.json
+++ b/package.json
@@ -32,7 +32,7 @@
   },
   "providedServices": {
     "tree-view": {
-      "description": "Get treeView Instance",
+      "description": "A tree-like view of directories and files",
       "versions": {
         "1.0.0": "provideTreeView"
       }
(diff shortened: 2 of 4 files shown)


UziTech:
Here is an example of why this would be useful:

https://github.com/subesokun/atom-tree-view-git-status/blob/master/lib/main.coffee#L154

The `tree-view-git-status` package uses 

```javascript
atom.packages.getActivePackage('tree-view').mainModule
```

to get the treeView instance. This poses two problems:

  1. It will break if the module changes the way it stores the treeView instance. (Ass it seems to have done in v1.18.0)
  2. It limits `tree-view-git-status` to only work with the `tree-view` package instead of also working with any fork of the `tree-view` package.

thomasjo:
Thanks for the contribution! I do however have concerns. Your proposition exposes the entire tree-view instance, which does not feel like an appropriate solution. The only justification I can think of is that it's the easiest approach with the least amount of effort. Keep in mind that most (if not all) of the tree-view object consists of private methods.

I propose that we expose an actual API, instead of simply handing developers an entire object, which essentially has no lifetime guarantees etc. For instance, if someone relies on this object, and the tree-view package is disabled, what do you think/expect will happen?

To begin with, which aspects of the tree-view do you think we have to expose? Let's begin with the bare minimum of features that are actually needed by external packages rights now — YAGNI™ and so on.

UziTech:
I agree, exposing the entire tree-view instance is probably not the right way to start out.

There are already some methods in tree-view class that are marked [public](https://github.com/atom/tree-view/blob/master/lib/tree-view.coffee#L843). I'm assuming those are supposed to be public to the rest of the package. Those seem unlikely to change in any major way.

The most common thing other packages seem to do is get the selected paths. If we wanted to start really small we should start with that.

thomasjo:
> The most common thing other packages seem to do is get the selected paths. If we wanted to start really small we should start with that.

OK, that sounds like a reasonable approach :1st_place_medal: 
Expose that functionality via a wrapper object. Ideally something that can withstand the tree-view being disabled, but doing something simple like we do in status-bar is probably acceptable as well. We can always improve that part later without breaking any dependent package.

UziTech:
I added `getEntry` as well for packages like `tree-view-git-status` that want to edit the elements in tree view.

`getEntry` sounds kind of weird to me. Should I change it to `getElement` or something like that?

lib/tree-view-package.js
```diff
@@ -52,6 +52,13 @@ class TreeViewPackage {
     )
   }
 
+  provideTreeView () {
+    return {
+      getSelectedPaths: () => this.getTreeViewInstance().selectedPaths(),
+      getEntry: (path) => this.getTreeViewInstance().entryForPath(path)
```
thomasjo: I am one of those people who dislike the `get` and `set` prefixes, since both words are extremely difficult to summarize/describe — look up both words in any English dictionary, and you will see what I mean (see e.g. [get](http://dictionary.cambridge.org/dictionary/english/get), and [set](http://dictionary.cambridge.org/dictionary/english/set)).

**TL; DR** What about changing these `selectedPaths` and `entryForPath`, just as they are named in the underlying API.

UziTech: I agree using the same name as the treeView instance might be helpful

UziTech:
@thomasjo anything else holding this up?

spec/tree-view-package-spec.coffee
```diff
@@ -3847,6 +3847,23 @@ describe "TreeView", ->
     files = Array.from(element.querySelectorAll('.entries .file'))
     files.find((file) -> file.fileName.textContent is text)
 
+describe "Service provider", ->
+  [provideTreeView, treeView] = []
+  beforeEach ->
+    waitForPackageActivation()
+
+    runs ->
+      provideTreeView = atom.packages.getActivePackage('tree-view').mainModule.provideTreeView()
```
thomasjo: Change the name of this to `treeViewService`, and move it below `treeView` (also re-order the variable declarations above).

spec/tree-view-package-spec.coffee
```diff
@@ -3847,6 +3847,23 @@ describe "TreeView", ->
...
+      treeView = atom.workspace.getLeftDock().getActivePaneItem()
+
+  it "should provide selected paths", ->
+    expect(provideTreeView.selectedPaths()).toEqual([atom.project.getPaths()[0]])
+
+  it "should provide entry for a path", ->
```
thomasjo: The description of both of these tests should be a bit more meaningful I think, but sadly I am unable to churn out anything right now to help you. Maybe one of the other maintainers can chime in with some ideas on how to rewrite them.

/cc @atom/feedback

package.json
```diff
@@ -30,6 +30,14 @@
       }
     }
   },
+  "providedServices": {
+    "tree-view": {
+      "description": "Get treeView Instance",
```
thomasjo: Rewrite this to something more descriptive and helpful to developers; perhaps something like:
"A tree-like view of directories and files"

/cc @atom/feedback

spec/tree-view-package-spec.coffee
UziTech: I'm not sure what this should test other than that the method exists on the service. The internal methods should have their own tests to make sure they are actually working properly.

thomasjo:
This LGTM! Thanks for not giving up, @UziTech :stuck_out_tongue_winking_eye: 

---
@atom/maintainers I'll give everyone a few hours to chime in on this, then I'll merge and ship this.

nathansobo:
@thomasjo You cool documenting this in the README before merging?

thomasjo:
@nathansobo Sure thing! Good call 🥇

thomasjo:
@UziTech You mind providing Atom maintainers push access to your fork? That wait I can push my commits directly to your fork before merging.

UziTech:
@thomasjo "Allow edits from maintainers." is checked is there anything else I need to do?

thomasjo:
Travis is being slow, so I'm going to merge this since the build was green prior to my documentation (and argument name tweak).

@UziTech Thanks for this! 🙇

Alhadis:
Well now, looks like we're another step closer to `tree-view` services providing DOM access. ;) Perhaps [this](https://github.com/atom/tree-view/pull/1146) should be the next API addition to review?

PR #1169, opened 2017-08-23
Switch internal directory representation to use maps

### LINKED ISSUE #1079: Directory named __proto__ causes file tree to throw an error
### Prerequisites

* [x] Put an X between the brackets on this line if you have done all of the following:
    * Reproduced the problem in Safe Mode: http://flight-manual.atom.io/hacking-atom/sections/debugging/#using-safe-mode
    * Followed all applicable steps in the debugging guide: http://flight-manual.atom.io/hacking-atom/sections/debugging/
    * Checked the FAQs on the message board for common solutions: https://discuss.atom.io/c/faq
    * Checked that your issue isn't already filed: https://github.com/issues?utf8=✓&q=is%3Aissue+user%3Aatom
    * Checked that there is not already an Atom package that provides the described functionality: https://atom.io/packages

### Description

If you open a project containing a directory named `__proto__`, the file tree throws several errors:

```
TypeError: entry.destroy is not a function
    at Directory.module.exports.Directory.reload (C:\Users\Daniel\AppData\Local\atom\app-1.16.0\resources\app.asar\node_modules\tree-view\lib\directory.js:354:15)
    at Directory.module.exports.Directory.expand (C:\Users\Daniel\AppData\Local\atom\app-1.16.0\resources\app.asar\node_modules\tree-view\lib\directory.js:383:12)
    at DirectoryView.module.exports.DirectoryView.expand (C:\Users\Daniel\AppData\Local\atom\app-1.16.0\resources\app.asar\node_modules\tree-view\lib\directory-view.js:185:24)
    at new DirectoryView (C:\Users\Daniel\AppData\Local\atom\app-1.16.0\resources\app.asar\node_modules\tree-view\lib\directory-view.js:77:14)
    at TreeView.<anonymous> (C:\Users\Daniel\AppData\Local\atom\app-1.16.0\resources\app.asar\node_modules\tree-view\lib\tree-view.js:604:18)
    at TreeView.module.exports.TreeView.updateRoots (C:\Users\Daniel\AppData\Local\atom\app-1.16.0\resources\app.asar\node_modules\tree-view\lib\tree-view.js:609:10)
    at new TreeView (C:\Users\Daniel\AppData\Local\atom\app-1.16.0\resources\app.asar\node_modules\tree-view\lib\tree-view.js:82:12)
    at Object.createView (C:\Users\Daniel\AppData\Local\atom\app-1.16.0\resources\app.asar\node_modules\tree-view\lib\main.js:118:25)
    at Object.activate (C:\Users\Daniel\AppData\Local\atom\app-1.16.0\resources\app.asar\node_modules\tree-view\lib\main.js:24:14)
    at Package.module.exports.Package.activateNow (C:\Users\Daniel\AppData\Local\atom\app-1.16.0\resources\app.asar\src\package.js:218:19)
    at C:\Users\Daniel\AppData\Local\atom\app-1.16.0\resources\app.asar\src\package.js:190:32
    at Package.module.exports.Package.measure (C:\Users\Daniel\AppData\Local\atom\app-1.16.0\resources\app.asar\src\package.js:96:15)
    at C:\Users\Daniel\AppData\Local\atom\app-1.16.0\resources\app.asar\src\package.js:183:26
    at Package.module.exports.Package.activate (C:\Users\Daniel\AppData\Local\atom\app-1.16.0\resources\app.asar\src\package.js:180:34)
    at PackageManager.module.exports.PackageManager.activatePackage (C:\Users\Daniel\AppData\Local\atom\app-1.16.0\resources\app.asar\src\package-manager.js:550:34)
    at C:\Users\Daniel\AppData\Local\atom\app-1.16.0\resources\app.asar\src\package-manager.js:531:29
    at Config.module.exports.Config.transactAsync (C:\Users\Daniel\AppData\Local\atom\app-1.16.0\resources\app.asar\src\config.js:337:18)
    at PackageManager.module.exports.PackageManager.activatePackages (C:\Users\Daniel\AppData\Local\atom\app-1.16.0\resources\app.asar\src\package-manager.js:526:19)
    at PackageManager.module.exports.PackageManager.activate (C:\Users\Daniel\AppData\Local\atom\app-1.16.0\resources\app.asar\src\package-manager.js:508:46)
    at C:\Users\Daniel\AppData\Local\atom\app-1.16.0\resources\app.asar\src\atom-environment.js:826:28
```

```
Uncaught TypeError: packagePath.indexOf is not a function
module.exports.NotificationIssue.getPackageName	@	C:\Users\Daniel\AppData\Local\atom\app-1.16.0\resources\app.asar\node_modules\notifications\lib\not…:221
module.exports.NotificationIssue.getRepoUrl	@	C:\Users\Daniel\AppData\Local\atom\app-1.16.0\resources\app.asar\node_modules\notifications\lib\not…:169
NotificationElement.renderFatalError	@	C:\Users\Daniel\AppData\Local\atom\app-1.16.0\resources\app.asar\node_modules\notifications\lib\not…:169
NotificationElement.render	@	C:\Users\Daniel\AppData\Local\atom\app-1.16.0\resources\app.asar\node_modules\notifications\lib\not…:161
NotificationElement.initialize	@	C:\Users\Daniel\AppData\Local\atom\app-1.16.0\resources\app.asar\node_modules\notifications\lib\not…:52
(anonymous function)	@	C:\Users\Daniel\AppData\Local\atom\app-1.16.0\resources\app.asar\node_modules\notifications\lib\mai…:89
module.exports.ViewRegistry.createView	@	C:\Users\Daniel\AppData\Local\atom\app-1.16.0\resources\app.asar\src\view-registry.js:117
module.exports.ViewRegistry.getView	@	C:\Users\Daniel\AppData\Local\atom\app-1.16.0\resources\app.asar\src\view-registry.js:84
addNotificationView	@	C:\Users\Daniel\AppData\Local\atom\app-1.16.0\resources\app.asar\node_modules\notifications\lib\mai…:126
(anonymous function)	@	C:\Users\Daniel\AppData\Local\atom\app-1.16.0\resources\app.asar\node_modules\notifications\lib\mai…:27
module.exports.Emitter.simpleDispatch	@	C:\Users\Daniel\AppData\Local\atom\app-1.16.0\resources\app.asar\node_modules\event-kit\lib\emitter…:25
module.exports.Emitter.emit	@	C:\Users\Daniel\AppData\Local\atom\app-1.16.0\resources\app.asar\node_modules\event-kit\lib\emitter…:141
module.exports.NotificationManager.addNotification	@	C:\Users\Daniel\AppData\Local\atom\app-1.16.0\resources\app.asar\src\notification-manager.js:54
module.exports.NotificationManager.addFatalError	@	C:\Users\Daniel\AppData\Local\atom\app-1.16.0\resources\app.asar\src\notification-manager.js:45
(anonymous function)	@	C:\Users\Daniel\AppData\Local\atom\app-1.16.0\resources\app.asar\node_modules\notifications\lib\mai…:53
module.exports.Emitter.simpleDispatch	@	C:\Users\Daniel\AppData\Local\atom\app-1.16.0\resources\app.asar\node_modules\event-kit\lib\emitter…:25
module.exports.Emitter.emit	@	C:\Users\Daniel\AppData\Local\atom\app-1.16.0\resources\app.asar\node_modules\event-kit\lib\emitter…:141
(anonymous function)	@	C:\Users\Daniel\AppData\Local\atom\app-1.16.0\resources\app.asar\src\atom-environment.js:897
```

### Steps to Reproduce
1. Create a new empty directory
2. Open the directory as a project in Atom
3. Right-click the project and select "New Folder". Call the folder `__proto__`

**Expected behavior:** 
Nothing bad happens

**Actual behavior:**
Errors are thrown. Even after closing and reopening Atom, errors continue to be thrown and the file tree is completely unusable until you delete the directory.

**Reproduces how often:** 100%

### Versions
Windows 10.0.16176
```
Atom    : 1.16.0
Electron: 1.3.13
Chrome  : 52.0.2743.82
Node    : 6.5.0
```

### Additional Information

Shout out to @zertosh for discovering that the string `__proto__` broke stuff on Facebook as well, and suggesting that someone sees whether it breaks anything in Atom/Nuclide 😛

50Wliu:
### Requirements

### Description of the Change

This PR switches the internal representations of directory entries to use `Map`s instead of `Object`s.  While they are very similar, there is one big disadvantage to using Objects - Objects have internal and inherited properties/keys.  Since setting object keys sets the key directly, this means that certain directory names can conflict and override the existing Object properties.  A notable example of this is `__proto__`.  Naming a directory `__proto__` causes Tree View to override the base Object prototype, which leads to Very Bad Things(tm).  Maps do not have this limitation.

### Alternate Designs

Switching to `Map` was the first thing I considered, and it worked.

### Benefits

The name of a directory should no longer render Atom unusable.

### Possible Drawbacks

This changes the output of `onDidRemoveEntries` from an Object to a Map.  I'm unaware if that's being used anywhere outside of Tree View.  In addition, the serialized expansion state has also changed from an Object to a Map, meaning that extra conversion code will need to be added for a smooth transition.

### Applicable Issues

Fixes #1079

/cc @Alhadis to make sure that this doesn't affect file-icons (in my testing it didn't)
/cc @Daniel15

TODO:
* [x] Specs
* [x] Serialization conversion
* [x] Make sure this doesn't break any packages

### DIFF
--- a/lib/directory-view.coffee
+++ b/lib/directory-view.coffee
@@ -108,10 +108,12 @@ class DirectoryView
       view = new FileView(entry)
 
     subscription = @directory.onDidRemoveEntries (removedEntries) ->
-      for removedName, removedEntry of removedEntries when entry is removedEntry
+      removedEntries.forEach (removedEntry, removedName) ->
+        return unless entry is removedEntry
         view.element.remove()
         subscription.dispose()
-        break
+        # TODO: When this file is converted to JS, convert this forEach loop
+        # to a for...of loop and add a break here for performance
     @subscriptions.add(subscription)
 
     view
--- a/lib/directory-view.coffee
+++ b/lib/directory-view.coffee
@@ -108,12 +108,9 @@ class DirectoryView
       view = new FileView(entry)
 
     subscription = @directory.onDidRemoveEntries (removedEntries) ->
-      removedEntries.forEach (removedEntry, removedName) ->
-        return unless entry is removedEntry
+      if removedEntries.has(entry)
         view.element.remove()
         subscription.dispose()
-        # TODO: When this file is converted to JS, convert this forEach loop
-        # to a for...of loop and add a break here for performance
     @subscriptions.add(subscription)
 
     view
--- a/lib/directory.coffee
+++ b/lib/directory.coffee
@@ -263,7 +263,8 @@ class Directory
       if @expansionState.entries.has(name)
         @expansionState.entries.delete(name)
 
-    @emitter.emit('did-remove-entries', removedEntries) if entriesRemoved
+    # Convert removedEntries to a Set containing only the entries for O(1) lookup
+    @emitter.emit('did-remove-entries', new Set(removedEntries.values())) if entriesRemoved
 
     if newEntries.length > 0
       @entries.set(entry.name, entry) for entry in newEntries
--- a/lib/directory.coffee
+++ b/lib/directory.coffee
@@ -31,10 +31,14 @@ class Directory
     # of time has passed since @expansionState.entries
     # has been converted to a Map
     unless @expansionState.entries instanceof Map
-      temp = new Map()
-      for name, entry of @expansionState.entries
-        temp.set(name, entry)
-      @expansionState.entries = temp
+      convertEntriesToMap = (entries) ->
+        temp = new Map()
+        for name, entry of entries
+          entry.entries = convertEntriesToMap(entry.entries) if entry.entries?
+          temp.set(name, entry)
+        return temp
+
+      @expansionState.entries = convertEntriesToMap(@expansionState.entries)
 
     @expansionState.entries ?= new Map()
     @status = null
--- a/lib/directory.coffee
+++ b/lib/directory.coffee
@@ -26,6 +26,16 @@ class Directory
     @isRoot ?= false
     @expansionState ?= {}
     @expansionState.isExpanded ?= false
+
+    # TODO: This can be removed after a sufficient amount
+    # of time has passed since @expansionState.entries
+    # has been converted to a Map
+    unless @expansionState.entries instanceof Map
+      temp = new Map()
+      for name, entry of @expansionState.entries
+        temp.set(name, entry)
+      @expansionState.entries = temp
+
     @expansionState.entries ?= new Map()
     @status = null
     @entries = new Map()
--- a/lib/directory.coffee
+++ b/lib/directory.coffee
@@ -26,9 +26,9 @@ class Directory
     @isRoot ?= false
     @expansionState ?= {}
     @expansionState.isExpanded ?= false
-    @expansionState.entries ?= {}
+    @expansionState.entries ?= new Map()
     @status = null
-    @entries = {}
+    @entries = new Map()
 
     @submodule = repoForPath(@path)?.isSubmodule(@path)
 
@@ -154,9 +154,9 @@ class Directory
       @watchSubscription.close()
       @watchSubscription = null
 
-    for key, entry of @entries
+    @entries.forEach (entry, key) =>
       entry.destroy()
-      delete @entries[key]
+      @entries.delete(key)
 
   # Public: Watch this directory for changes.
   watch: ->
@@ -188,15 +188,15 @@ class Directory
         statFlat[key] = statFlat[key]?.getTime()
 
       if stat.isDirectory?()
-        if @entries.hasOwnProperty(name)
+        if @entries.has(name)
           # push a placeholder since this entry already exists but this helps
           # track the insertion index for the created views
           directories.push(name)
         else
-          expansionState = @expansionState.entries[name]
+          expansionState = @expansionState.entries.get(name)
           directories.push(new Directory({name, fullPath, symlink, expansionState, @ignoredPatterns, @useSyncFS, stats: statFlat}))
       else if stat.isFile?()
-        if @entries.hasOwnProperty(name)
+        if @entries.has(name)
           # push a placeholder since this entry already exists but this helps
           # track the insertion index for the created views
           files.push(name)
@@ -225,12 +225,12 @@ class Directory
   # Public: Perform a synchronous reload of the directory.
   reload: ->
     newEntries = []
-    removedEntries = _.clone(@entries)
+    removedEntries = new Map(@entries)
     index = 0
 
     for entry in @getEntries()
-      if @entries.hasOwnProperty(entry)
-        delete removedEntries[entry]
+      if @entries.has(entry)
+        removedEntries.delete(entry)
         index++
         continue
 
@@ -239,20 +239,20 @@ class Directory
       newEntries.push(entry)
 
     entriesRemoved = false
-    for name, entry of removedEntries
+    removedEntries.forEach (entry, name) =>
       entriesRemoved = true
       entry.destroy()
 
-      if @entries.hasOwnProperty(name)
-        delete @entries[name]
+      if @entries.has(name)
+        @entries.delete(name)
 
-      if @expansionState.entries.hasOwnProperty(name)
-        delete @expansionState.entries[name]
+      if @expansionState.entries.has(name)
+        @expansionState.entries.delete(name)
 
     @emitter.emit('did-remove-entries', removedEntries) if entriesRemoved
 
     if newEntries.length > 0
-      @entries[entry.name] = entry for entry in newEntries
+      @entries.set(entry.name, entry) for entry in newEntries
       @emitter.emit('did-add-entries', newEntries)
 
   # Public: Collapse this directory and stop watching it.
@@ -273,9 +273,10 @@ class Directory
   serializeExpansionState: ->
     expansionState = {}
     expansionState.isExpanded = @expansionState.isExpanded
-    expansionState.entries = {}
-    for name, entry of @entries when entry.expansionState?
-      expansionState.entries[name] = entry.serializeExpansionState()
+    expansionState.entries = new Map()
+    @entries.forEach (entry, name) ->
+      return unless entry.expansionState?
+      expansionState.entries.set(name, entry.serializeExpansionState())
     expansionState
 
   squashDirectoryNames: (fullPath) ->
--- a/spec/file-stats-spec.coffee
+++ b/spec/file-stats-spec.coffee
@@ -31,13 +31,13 @@ describe "FileStats", ->
       temp.cleanup()
 
     it "passes stats to File instances", ->
-      stats = treeView.roots[0].directory.entries["file1.txt"].stats
+      stats = treeView.roots[0].directory.entries.get("file1.txt").stats
       expect(stats).toBeDefined()
       expect(stats.mtime).toBeDefined()
       expect(stats.size).toEqual(file1Data.length)
 
     it "passes stats to Directory instances", ->
-      stats = treeView.roots[0].directory.entries["subdir"].stats
+      stats = treeView.roots[0].directory.entries.get("subdir").stats
       expect(stats).toBeDefined()
       expect(stats.mtime).toBeDefined()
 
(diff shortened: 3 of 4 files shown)


Alhadis:
The same error probably occurs with files that're named `__proto__`, too. Did you check?

50Wliu:
Really?  I'm pretty sure I tested out a __proto__ file and nothing broke.

lib/directory-view.coffee
```diff
@@ -108,10 +108,12 @@ class DirectoryView
...
 
     subscription = @directory.onDidRemoveEntries (removedEntries) ->
-      for removedName, removedEntry of removedEntries when entry is removedEntry
+      removedEntries.forEach (removedEntry, removedName) ->
+        return unless entry is removedEntry
         view.element.remove()
         subscription.dispose()
-        break
+        # TODO: When this file is converted to JS, convert this forEach loop
+        # to a for...of loop and add a break here for performance
```
thomasjo: Another option is to avoid the "problem" by using `Array.prototype.some` instead;

``` coffee
removedEntries.some (removedEntry, removedName) ->
  return true unless entry is removedEntry
  view.element.remove()
  subscription.dispose()
``` 

This code will behave essentially exactly like the old version — it will iterate until the desired item is encountered, then short-circuit the iteration. You could of course also achieve this by using `Array.prototype.every`, but the version above more closely mimics the old version I think (at least in terms of lines of code, and overall legibility).

thomasjo: Nevermind... just realized you're working with `Map` and not `Array` — d'oh!

Alhadis:
Well, I created the `__proto__` file from the tree-view's context-menu, and this happened as soon as I entered the filename:

Figure 1

After reloading the window, I encountered a different error:

Figure 2

50Wliu:
@Alhadis are you trying with my PR?  The `Directory` class handles files to some extent as well.

Alhadis:
Oh! Whoops, my bad. Yeah, your branch handles files as well. Never mind. 👼

50Wliu:
Just reminding myself which part of the serialization I need to work on:
```
Uncaught (in promise) TypeError: this.expansionState.entries.get is not a function
    at Directory.module.exports.Directory.getEntries (file:///C:/Users/user/Documents/GitHub/tree-view/lib/directory.coffee:196:52)
    at Directory.module.exports.Directory.reload (file:///C:/Users/user/Documents/GitHub/tree-view/lib/directory.coffee:231:5)
    at Directory.module.exports.Directory.expand (file:///C:/Users/user/Documents/GitHub/tree-view/lib/directory.coffee:269:6)
    at DirectoryView.module.exports.DirectoryView.expand (file:///C:/Users/user/Documents/GitHub/tree-view/lib/directory-view.coffee:133:18)
    at new DirectoryView (file:///C:/Users/user/Documents/GitHub/tree-view/lib/directory-view.coffee:62:6)
    at TreeView.<anonymous> (file:///C:/Users/user/Documents/GitHub/tree-view/lib/tree-view.coffee:296:14)
    at TreeView.module.exports.TreeView.updateRoots (file:///C:/Users/user/Documents/GitHub/tree-view/lib/tree-view.coffee:277:14)
    at new TreeView (file:///C:/Users/user/Documents/GitHub/tree-view/lib/tree-view.coffee:57:6)
    at TreeViewPackage.getTreeViewInstance (C:\Users\user\Documents\GitHub\tree-view\lib\tree-view-package.js:57:23)
    at Object.deserialize (C:\Users\user\AppData\Local\Atom x64\app-dev\resources\app\src\package.js:562:58)
    at DeserializerManager.deserialize (C:\Users\user\AppData\Local\Atom x64\app-dev\resources\app\src\deserializer-manager.js:95:35)
    at C:\Users\user\AppData\Local\Atom x64\app-dev\resources\app\src\pane.js:48:36
    at Array.map (native)
    at Function.module.exports.Pane.deserialize (C:\Users\user\AppData\Local\Atom x64\app-dev\resources\app\src\pane.js:47:27)
    at DeserializerManager.deserialize (C:\Users\user\AppData\Local\Atom x64\app-dev\resources\app\src\deserializer-manager.js:95:35)
    at PaneContainer.deserialize (C:\Users\user\AppData\Local\Atom x64\app-dev\resources\app\src\pane-container.js:58:44)
    at Dock.deserialize (C:\Users\user\AppData\Local\Atom x64\app-dev\resources\app\src\dock.js:378:30)
    at Workspace.deserialize (C:\Users\user\AppData\Local\Atom x64\app-dev\resources\app\src\workspace.js:416:40)
    at C:\Users\user\AppData\Local\Atom x64\app-dev\resources\app\src\atom-environment.js:1263:35
```

lib/directory-view.coffee
50Wliu: Yep, it's annoying because Coffeescript's `for...of` is JS's `for...in`, and of course its `for...in` is just a fancy for loop and not `for...of`.  So there's no way to loop Map entries.  Oh well.

lib/directory.coffee
```diff
@@ -26,6 +26,16 @@ class Directory
     @isRoot ?= false
     @expansionState ?= {}
     @expansionState.isExpanded ?= false
+
+    # TODO: This can be removed after a sufficient amount
+    # of time has passed since @expansionState.entries
```
Alhadis: Waiting until after #1146 should be enough, too. 😜

50Wliu:
`grep -r "onDidRemoveEntries" .` found three packages with that text:
* https://atom.io/packages/tree-view-no-drag
  * Fork that automatically updates based on a script, shouldn't be a problem
* https://atom.io/packages/tree-view-remote
  * Fork with 0 downloads, last commit Nov. 24, 2014
* https://atom.io/packages/greyatom-ide-tree
  * Fork maintained by Greyatom IDE

So it looks like only Tree View forks are using `onDidRemoveEntries` and therefore it should be safe to proceed with this implementation.

Alhadis:
@50Wliu Just an FYI, there were indeed regressions with `file-icons`; but I've got it all under control. Specifically, directory icons were missing, and icons weren't responding to async changes in file/icon mappings. No big deal.

lib/directory-view.coffee
as-cii: Since you already checked that using `Map`s does not break consumers of this API, I wonder if we could change the parameters to make this code path more efficient. In particular, what about exposing a `Set` of entries that were removed so that we can retrieve it in `O(1)` here? Like:

```js
if (removedEntries.has(entry)) {
  // cleanup
}
```

An alternative might be to rely on the name and avoid changing the API altogether:

```js
if (removedEntries.has(entry.name)) {
  // cleanup
}
```

Although I am slightly concerned about the order of events when a rename occurs if we decide to go choose the second option.

What do you think? 💭

as-cii:
One more question, related to @Alhadis' comment:

> Just an FYI, there were indeed regressions with file-icons; specifically, directory icons were missing, and icons weren't responding to async changes in file/icon mappings. Not to worry though, I've got it all under control.

Are we worried at all about breaking other packages by shipping this pull-request?

Alhadis:
> Are we worried at all about breaking other packages by shipping this pull-request?

@as-cii I don't imagine there'd be a risk of runtime breakage. They were POJOs in the first place, which means any package using them would probably only be looping through their properties with a `for..in` loop.

However, this might lead to subtle regressions in packages which expect to find filenames in the form of enumerated key/value pairs (as `file-icons` had been).

50Wliu:
@as-cii could you expand on your concerns about the event order with the second option?

lib/directory-view.coffee
as-cii: > could you expand on your concerns about the event order with the second option?

Yeah, sorry if that comment was a little obscure. My primary concern is that in the same tick of the loop the following could happen:

1. An entry with name `x` is added.
2. A previous entry with name `x` is deleted. 

I am not sure this would be possible in the current model, but if we receive change events asynchronously I worry that the add and the delete event could get grouped together and cause the wrong entry (or both, in the example above) to be deleted.

If that's a circumstance that we are sure we will never encounter, then I think we could go ahead and implement 2).

50Wliu:
@as-cii I have implemented the first option, just to be safe.

Issue #656, opened 2015-11-19
Copying and pasting directories inside themselves has undesirable results


Mr-Wallet:
Version 1.2.2, Windows 8.
- Select a folder `A` with file `x`.
- Perform a copy, then perform 3 pastes.
- The following structure results:

```
A
| A
| | x
|
| A0
| | A
| | | x
| | x
|
| A1
| | A
| | | x
| |
| | A0
| | | A
| | | | x
| | | x
| | x
| x
```

The paste operation is looking at the source path _at the moment of pasting,_ which is unintuitive. I would expect it to store a list of all file paths _at the moment of copying._

I'm not familiar with the exact effects of #631 but in 1.2.2, the effects of selecting the pasted item can be disastrous:
- Select a folder `A` with file `x`.
- Perform a copy, then perform a paste.
- Select the folder `A/A`.
- Perform a paste.
- The following structure results:

```
A
| A
| | A
| | | A
| | | | A
| | | | | A
| | | | | | A
...
| | x
| x
```

Atom locks up and continues making empty `A` directories until the application is forcibly closed.
When reproducing this issue, it's very easy to end up with file paths so long that Windows is unable to delete them. I found [rimraf](http://superuser.com/questions/78434/how-to-delete-directories-with-path-names-too-long-for-normal-delete) very helpful in cleaning up afterward.

glangford:
Under #631 - this sequence causes an ENAMETOOLONG to be thrown (rather than generating tons of directories):

* Select a folder A with file x.
* Perform a copy, then perform a paste. [paste triggers reselection]
* Perform a paste.

Ben3eeE:
Duplicate https://github.com/atom/tree-view/issues/509?

Mr-Wallet:
The 2nd half looks like a duplicate, but the first half is not attested in 509, even though I believe they probably have a common cause.

openks:
yes,I got the same result,it take me for half an hour to remove the useless folders, before i remove all the folders when i open the atom it continues make new folders... it's not easy to remove these folders beacse its too long for windows file system. something must be wrong with it,

Ben3eeE:
@ungb can you reproduce this and check https://github.com/atom/tree-view/issues/509 also?

ungb:
I can repro this.  

We should do what the OS does and not allow this behavior.  Neither mac or windows File Explorer allows this.

[...]

Issue #562, opened 2015-08-30
Tree view doesn't show broken symbolic links


jbt:
If a symlink's target doesn't exist, tree-view doesn't show the link at all. Ideally it'd show some kind of indication that yes, the link exists, but the file it's pointing to doesn't.

For example: (bar links to foo and works fine, baz links to something nonexistent but doesn't show at all)

```sh
$ touch foo
$ ln -s foo bar
$ ln -s i-dont-exist baz
```

![screenshot from 2015-08-30 10 29 48](https://cloud.githubusercontent.com/assets/120506/9566555/120d8e9c-4f02-11e5-96d1-e17713ee7cd0.png)

maqdev:
+1

raptor235:
any movements on this?

50Wliu:
Brief investigation notes: this is because `fs.statSync` fails when it tries to access the location the symlink it points to, and Tree View only shows entries where `fs.(l)statSync` succeeds.  If I have time I'll experiment with labeling all unknown entries as files.

TomBugnon:
Is there any progress here?

efharkin:
Would also like to see this addressed.

briser:
bump.

PR #1173, opened 2017-09-04
Do not allow moving a folder into itself

### LINKED ISSUE #509: Copying folder content into one of its sub-folders
In the Files Tree View, when coping folder "A" into one of its sub-folders "A/B"; Atom gets into an infinity loop in which the same content  of the main folder "A" will be duplicated in the target sub-folder "A/B/A" (the first time is as desired) then again copies itself into the newly created sub-sub-folder (A/B/A/"B").. and goes on and on...

E.g: Coping folder FONTS/ into FONTS/FA will produce: "FONTS/FA/FONTS/FA/FONTS/FA/FONTS/FA/FONTS/FA/..." and will not stop until you close Atom.

50Wliu:
### Description of the Change

Do not allow any recursive copying, such as when attempting to copy `test/a` into `test/a/b`.  Before this PR, such actions would be allowed and would result in a `a/b/a/b/a/b/a/b/...` infinite directory structure.  Drag-and-dropping previously emitted a non-obvious warning (`cannot rename a to a/b`).  Both now emit a "Cannot paste/move a folder into itself" warning and return without performing any additional actions.

As a side effect of this change, Tree View no longer allows copying a directory directly inside itself, such as from `a/` to `a/` (resulting in `a/a/`).  This is consistent with behavior on Windows.

### Alternate Designs

Carefully copy the folder structure _once_ instead of recursively.  This is much harder to implement and is also not what major OSes do.

### Benefits

This fixes a usability issue where you would be unintentionally left with a massive folder structure (unable to be deleted on Windows through conventional means) that freezes Atom.

### Possible Drawbacks

The only negative impact I can think of is pasting folders directly inside themselves no longer works.

### Applicable Issues

Fixes #509
Refs #656, but I do not believe that this PR fixes that issue
Supersedes and closes #952

/cc @Ben3eeE @ungb: I'm interested in seeing whether or not pasting folders directly inside themselves works on macOS and Linux.  Also, I haven't tested this with symlinks.


ungb:
Tested it on mac without symlinks and things look good.  

With symlink does you get a weird behavior where it attempts to do it but fails after the first folder.  none of the contents in that folder are copied either.

Attempt to copy a into syma\c
image


you get the following:

Creates the folder A but nothing else under it.
image


Moving:
image

50Wliu:
Found a regression: You can't copy `a` into `aa`, even if they're on the same level as each other.  Probably due to my `startsWith` matching.

ungb:
good find @50Wliu!  on mac, if you try pasting into a symlink you get the same thing as if you try to paste a parent into a child or inside itself

image

50Wliu:
Ok, good to know.  That's different than the behavior that Windows exhibits (which is not surprising, considering its symlink support).  I think I will defer to the macOS implementation for symlinks.

jasonrudolph:
@50Wliu: I can take a look at this pull request this week and try to get it merged. Would you be willing to summarize the [verification process](https://github.com/atom/teletype/blame/b53e1445d394310367f4725ce185e92c78d6ce1a/PULL_REQUEST_TEMPLATE.md#L32-L38) that you've used to verify that these changes have the desired effects?

50Wliu:
@jasonrudolph Sorry for the delay!  Here's the test plan I came up with.
Create a `test` folder with the following subfolders: `a/`, `a/b/`, `aa/`
* Attempt to copy `test/a/` into `test/a/b/` using the `tree-view:copy` and `tree-view:paste` commands.
  * [ ] Notification should appear and the folder structure should remain unchanged.
* Attempt to move `test/a/` into `test/a/b/` using drag-and-drop.
  * [ ] Notification should appear and the folder structure should remain unchanged.
* Attempt to copy `test/a/` into `test/aa/`.
  * [ ] Operation should complete successfully and both `test/a/` as well as `test/aa/a/` should exist.
* Attempt to move `test/a/` into `test/aa/` using drag-and-drop.
  * [ ] Operation should complete successfully and only `test/aa/a/` should exist.

Now, create a symlink `syma` that points to the `a` directory.
* Attempt to copy `test/a/` into `/test/syma/b/`.
  * [ ] Notification should appear and there should not be a `a` folder under `b`.
* Attempt to move `test/a/` into `/test/syma/b/`.
  * [ ] Notification should appear and there should not be a `a` folder under `b`.

To the best of my recollection, that's how I tested this PR when I first created it.  If you'd like to see some other tests, please let me know.

jasonrudolph:
@50Wliu: When you have a moment, would you mind merging master into this branch and resolving the merge conflicts?

jasonrudolph:
> Here's the test plan I came up with. ... To the best of my recollection, that's how I tested this PR when I first created it. If you'd like to see some other tests, please let me know.

Thanks, @50Wliu. [Those tests](https://github.com/atom/tree-view/pull/1173#issuecomment-360821709) all seems useful, and I'd like to recommend a few additional ones:

- Attempt to move `test/a/` into `test/a/b/` using the `tree-view:cut` and `tree-view:paste` commands.
    - [ ] Notification should appear and the folder structure should remain unchanged.

- Attempt to copy `test/syma/` into `/test/a`.
    - [ ] Operation should complete successfully and only `test/a/syma` should exist.

Do those tests seem reasonable? If so, I recommend the following next steps:

- [ ] Merge master into this branch and resolve the merge conflicts (https://github.com/atom/tree-view/pull/1173#issuecomment-364243172)
- [ ] Run through the test plan on Windows
- [ ] Run through the test plan on either Linux or macOS
- [ ] If the test plan reveals that everything is functioning correctly, merge this pull request :+1:

50Wliu:
Assuming for the last bullet point you meant "attempt to _move_", then the test plan passes on Windows.

rsese:
Went through the verification process on macOS 10.12.6 and Ubuntu 16.04 - only 1 of the tests didn't have the expected result:

> - Attempt to move `test/syma/` into `/test/a`.
>     - [ ] Operation should complete successfully and only `test/a/syma` should exist.

On both macOS and Ubuntu, the operation completes without an error notification or anything in devtools console, but `test/a/syma` doesn't exist and `test/syma` no longer exists either.

GIF-age on macOS (behavior is the same on Ubuntu):

test-tree-view-1173

All the other tests had the expected results but let me know if I misunderstood anything with this last test @50Wliu.

50Wliu:
@rsese instead of using a relative path for the symlink, can you try specifying the absolute path to `a`?

rsese:
Just gave that a try on macOS and Ubuntu and on both, `syma` shows up in the tree-view after using an absolute path to `a/`:

test-tree-view-1173-2

50Wliu:
I've talked to rsese and the reason the symlink disappeared when it was a relative path is because of #562.  If we ignore #562, then the test plan passes (and indeed, it does when an absolute path is used).

PR #1179, opened 2017-09-10
Drag multiple items in tree-view

### LINKED ISSUE #232: Move multiple files.

UziTech:
### Description of the Change

* Allow multiple items to be dragged to a folder. This is a rebase of #793 

multi-drag screenshot

* Allow multi-select to work like it is supposed to

shift + ctrl + click screenshot

### Benefits

Multi-select will work like other file explorers

### Possible Drawbacks

none

### Applicable Issues

Fixes #232
Supersedes and closes https://github.com/atom/tree-view/pull/793

### DIFF
--- a/lib/tree-view.coffee
+++ b/lib/tree-view.coffee
@@ -898,47 +898,56 @@ class TreeView
     @element.style.display = ''
 
   onMouseDown: (e) ->
-    if entryToSelect = e.target.closest('.entry')
-      e.stopPropagation()
+    return unless entryToSelect = e.target.closest('.entry')
 
-      # return early if we're opening a contextual menu (right click) during multi-select mode
-      if @multiSelectEnabled() and entryToSelect.classList.contains('selected')
+    e.stopPropagation()
 
-        # mouse right click or ctrl click as right click on darwin platforms
-        if e.button is 2 or (e.ctrlKey and process.platform is 'darwin')
-          return
-        else
-          # allow select if not dragging
-          {shiftKey, metaKey, ctrlKey} = e
-          @selectOnMouseUp = {shiftKey, metaKey, ctrlKey}
-          return
+    # TODO: meta+click and ctrl+click should not do the same thing on Windows.
+    #       Right now removing metaKey if platform is not darwin breaks tests
+    #       that set the metaKey to true when simulating a cmd+click on macos
+    #       and ctrl+click on windows and linux.
+    cmdKey = e.metaKey or (e.ctrlKey and process.platform isnt 'darwin')
 
-      if e.shiftKey and (e.metaKey or (e.ctrlKey and process.platform isnt 'darwin'))
-        # select continuous from @lastFocusedElement but leave others
-        @selectContinuousEntries(entryToSelect, false)
-        @toggleMultiSelectMenu()
-      else if e.shiftKey
-        # select continuous from @lastFocusedElement and deselect rest
-        @selectContinuousEntries(entryToSelect)
-        @toggleMultiSelectMenu()
-      # only allow ctrl click for multi selection on non darwin systems
-      else if e.metaKey or (e.ctrlKey and process.platform isnt 'darwin')
-        @selectMultipleEntries(entryToSelect)
-        @lastFocusedElement = entryToSelect
-        @toggleMultiSelectMenu()
+    # return early if clicking on a selected entry
+    if entryToSelect.classList.contains('selected')
+
+      # mouse right click or ctrl click as right click on darwin platforms
+      if e.button is 2 or (e.ctrlKey and process.platform is 'darwin')
+        return
       else
-        @selectEntry(entryToSelect)
-        @showFullMenu()
+        # allow click on mouseup if not dragging
+        {shiftKey} = e
+        @selectOnMouseUp = {shiftKey, cmdKey}
+        return
+
+    if e.shiftKey and cmdKey
+      # select continuous from @lastFocusedElement but leave others
+      @selectContinuousEntries(entryToSelect, false)
+      @toggleMultiSelectMenu()
+    else if e.shiftKey
+      # select continuous from @lastFocusedElement and deselect rest
+      @selectContinuousEntries(entryToSelect)
+      @toggleMultiSelectMenu()
+    # only allow ctrl click for multi selection on non darwin systems
+    else if cmdKey
+      @selectMultipleEntries(entryToSelect)
+      @lastFocusedElement = entryToSelect
+      @toggleMultiSelectMenu()
+    else
+      @selectEntry(entryToSelect)
+      @showFullMenu()
 
   onMouseUp: (e) ->
     return unless @selectOnMouseUp?
 
-    {shiftKey, metaKey, ctrlKey} = @selectOnMouseUp
+    {shiftKey, cmdKey} = @selectOnMouseUp
     @selectOnMouseUp = null
 
     return unless entryToSelect = e.target.closest('.entry')
 
-    if shiftKey and (metaKey or (ctrlKey and process.platform isnt 'darwin'))
+    e.stopPropagation()
+
+    if shiftKey and cmdKey
       # select continuous from @lastFocusedElement but leave others
       @selectContinuousEntries(entryToSelect, false)
       @toggleMultiSelectMenu()
@@ -947,7 +956,7 @@ class TreeView
       @selectContinuousEntries(entryToSelect)
       @toggleMultiSelectMenu()
     # only allow ctrl click for multi selection on non darwin systems
-    else if metaKey or (ctrlKey and process.platform isnt 'darwin')
+    else if cmdKey
       @deselect([entryToSelect])
       @lastFocusedElement = entryToSelect
       @toggleMultiSelectMenu()
--- a/lib/tree-view.coffee
+++ b/lib/tree-view.coffee
@@ -1025,14 +1025,17 @@ class TreeView
       dragImage.style.willChange = "transform"
 
       for target in @getSelectedEntries()
-        nameElement = target.querySelector(".name")
-        entryPath = nameElement.dataset.path
+        entryPath = target.querySelector(".name").dataset.path
         unless path.dirname(entryPath) in initialPaths
           initialPaths.push(entryPath)
-          newNameElement = nameElement.cloneNode(true)
-          for key, value of getStyleObject(nameElement)
-            newNameElement.style[key] = value
-          dragImage.append(newNameElement)
+          newElement = target.cloneNode(true)
+          if newElement.classList.contains("directory")
+            newElement.querySelector(".entries").remove()
+          for key, value of getStyleObject(target)
+            newElement.style[key] = value
+          newElement.style.paddingLeft = "1em"
+          newElement.style.paddingRight = "1em"
+          dragImage.append(newElement)
 
 
       document.body.appendChild(dragImage)
--- a/lib/tree-view.coffee
+++ b/lib/tree-view.coffee
@@ -1058,7 +1058,8 @@ class TreeView
 
       if initialPaths
         # Drop event from Atom
-        for initialPath in initialPaths.split(',')
+        # iterate backwards so files in a dir are moved before the dir itself
+        for initialPath in initialPaths.split(',') by -1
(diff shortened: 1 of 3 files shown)


50Wliu:
@UziTech what is the purpose of 0f7a4b7?

UziTech:
@50Wliu

When trying to drag multiple files you click and hold on one of the selected files and drag to a folder.

When selecting only a single selected file when multiple are selected you just click on the selected file.

We need to make sure the user is not dragging before deselecting all files except for the one that is clicked on.  In order to do that we need to do the deselect on mouseup. 

We still want most mouse operations to occur on mousedown. Only when we need to make sure there is no dragging do we do the operation on mouseup.

Ben3eeE:
I'm signing up for testing here since I tested the original PR.

andyfleming:
Great work! Let's get this done!

I can't believe we can't drag multiple files in atom yet 😆 .

UziTech:
@Ben3eeE anything holding this up?

Ben3eeE:
@UziTech Sorry been swamped with other things. I will take a look as soon as I get a chance to, hopefully I will have some time this weekend ⚡️

UziTech:
I rebased again and fixed the merge conflicts in the tests

UziTech:
@Ben3eeE Any way we can get this out the door so I don't have to keep rebasing?

arthurlacoste:
Please, merge this feature !

arthurlacoste:
Please, merge this feature !

arthurlacoste:
Please, merge this feature !

UziTech:
Windows tests are failing because of #1203 

@50Wliu @Ben3eeE any chance this can get merged

UziTech:
@Ben3eeE is this waiting on anything?

UziTech:
@50Wliu @Ben3eeE @nathansobo any reason this isn't getting merged?

nathansobo:
@ungb @rsese Can one of you test this?

50Wliu:
@Ben3eeE and I have been trying to find a time where we can test this together and do a code review.

Ben3eeE:
@UziTech Sorry for taking so long before I could get to this.

Here is a test plan with some scenarios I can think of. If anyone can think of anything else then please let me know :bowing_man: 

## Test plan

- [ ] Selecting and deselecting multiple files and folders with ctrl/shift click (cmd on macOS?`)
  - [ ] Windows
  - [ ] Linux
  - [ ] macOS
- [ ] Multiple drag and drop
  - [ ] Multiple files
  - [ ] Multiple folders
  - [ ] Multiple files and folders
- [ ] Initiating drag 
  - [ ] From a file
  - [ ] From a folder
    - [ ] From a collapsed folder
    - [ ] From an expanded folder
    - [ ] From an expanded folder when you have files in the folder selected /cc: https://github.com/atom/tree-view/pull/793#issuecomment-222503727
- [ ] Dropping in another Atom window
  - [ ] Multiple files
  - [ ] Multiple folders
  - [ ] Multiple files and folders
- [ ] Drag image should look correct
- [ ] Conflicts
  - [ ] Dragging multiple files/folders when the project root folder is included in the selection /cc: https://github.com/atom/tree-view/pull/525
    - [ ] Dropping in another Atom window when the root directory is selected
  - [ ] Dropping in other locations of the Atom window

UziTech:
I fixed a few of the highlighting issues when dragging that occurred because we use the `'selected'` class to highlight folder that are dragged over.

I feel like the right way to fix it would be to use a different class (e.g. `'drag-over-directory'`) to highlight the directory but that would require a change in all themes.

Not that it would be a common bug but if someone checked `treeView.getSelectedEntries()` when dragging files over a directory, the directory would be added to the selected entries which would not be correct.

UziTech:
That actually doesn't fix the highlighting issue when dragging from a different Atom window. The only way I can think of to fix it would be to use a different class.

UziTech:
I believe I fixed the issue where a folder would get deselected when dragging over it with files from another Atom window.

I added the class `drag-over` as well as `selected` so themes still highlight the folder but only remove the `selected` class on leave if it also has the `drag-over` class

@Ben3eeE I manually tested all of those cases on a Windows 10 and they seem to work as expected. One other case I came across that doesn't work but probably should is dropping a selection on a file. I would expect that the selection would be moved into the same folder as the file that is being dropped on, but right now the move operation is just canceled when dropping on a file.

Actual:
screenshot2

Expected:
screenshot

Ben3eeE:
> One other case I came across that doesn't work but probably should is dropping a selection on a file. I would expect that the selection would be moved into the same folder as the file that is being dropped on, but right now the move operation is just canceled when dropping on a file.

Tested this quick and it appears to also be an issue on `master` so it's not going to block this PR from moving forward. Do you want to create an issue for this?

UziTech:
I just added it to this pr

lib/tree-view.coffee
```diff
@@ -869,7 +873,7 @@ class TreeView
       return
 
     entryName = path.basename(initialPath)
-    newPath = "#{newDirectoryPath}/#{entryName}".replace(/\s+$/, '')
+    newPath = "#{newDirectoryPath}#{path.sep}#{entryName}".replace(/\s+$/, '')
```
50Wliu: Nit: could you remove this change and add it in a separate PR?  It should also be `path.join(newDirectoryPath, entryName)`.

I ask because I'm also fixing this same line in both #1180 and #1173, so it would probably be better to just submit this change in a different PR.

lib/tree-view.coffee
```diff
@@ -894,28 +898,71 @@ class TreeView
...
 
-      # return early if we're opening a contextual menu (right click) during multi-select mode
-      if @multiSelectEnabled() and
-         entryToSelect.classList.contains('selected') and
-         # mouse right click or ctrl click as right click on darwin platforms
-         (e.button is 2 or e.ctrlKey and process.platform is 'darwin')
-        return
+    e.stopPropagation()
+
+    # TODO: meta+click and ctrl+click should not do the same thing on Windows.
```
50Wliu: Is fixing this something that would require a significant rework of the specs?

lib/tree-view.coffee
```diff
@@ -894,28 +898,71 @@ class TreeView
...
+    cmdKey = e.metaKey or (e.ctrlKey and process.platform isnt 'darwin')
 
-      if e.shiftKey
-        @selectContinuousEntries(entryToSelect)
-        @showMultiSelectMenu()
-      # only allow ctrl click for multi selection on non darwin systems
-      else if e.metaKey or (e.ctrlKey and process.platform isnt 'darwin')
-        @selectMultipleEntries(entryToSelect)
+    # return early if clicking on a selected entry
+    if entryToSelect.classList.contains('selected')
```
50Wliu: :fire: blank line

lib/tree-view.coffee
```diff
@@ -38,6 +38,8 @@ class TreeView
     @emitter = new Emitter
     @roots = []
     @selectedPath = null
+    @selectOnMouseUp = null
+    @lastFocusedElement = null
```
50Wliu: Seems like this makes more sense as `@lastFocusedEntry`?

lib/tree-view.coffee
```diff
@@ -957,103 +1005,132 @@ class TreeView
...
     @list.classList.add('full-menu')
 
-  # Public: Toggle multi-select class on the main list element to display the the
+  # Public: Toggle multi-select class on the main list element to display the
   #         menu with only items that make sense for multi select functionality
   showMultiSelectMenu: ->
     @list.classList.remove('full-menu')
     @list.classList.add('multi-select')
 
+  toggleMultiSelectMenu: ->
```
50Wliu: `toggle` generally implies "if on, turn off (and vice versa)".  Therefore a better name for this method might be `showMultiSelectMenuIfNecessary`.

lib/tree-view.coffee
```diff
@@ -957,103 +1005,132 @@ class TreeView
...
+          newElement = target.cloneNode(true)
+          if newElement.classList.contains("directory")
+            newElement.querySelector(".entries").remove()
+          for key, value of getStyleObject(target)
+            newElement.style[key] = value
+          newElement.style.paddingLeft = "1em"
+          newElement.style.paddingRight = "1em"
+          dragImage.append(newElement)
+
+
```
50Wliu: :fire: extra empty line

lib/tree-view.coffee
UziTech: Yes. Any spec that simulates a right click does so by setting e.metaKey to true.

It is also not that big of a deal because in Windows meta(Windows key)+click isn't something people usually do.

50Wliu:
Going back to your comment here: https://github.com/atom/tree-view/pull/1179#issuecomment-328366447

Perhaps `click` could be used to combine the two cases?

UziTech:
I could use `click` but I would still need a variable letting the `onClick` function know if it should execute otherwise it would deselect a file after selecting it on `mousedown`. And we do want to use the `mousedown` function for selecting because someone might want to ctrl+click then just start dragging.

screenshot

adamreisnz:
Would be great to have this 👍

50Wliu:
Code changes look good to me.  @Ben3eeE if you're good with this as well I think it can :ship:?

Ben3eeE:
@UziTech All good from me when testing on Windows 👍 @ungb is going to go over the test plan on macOS and Linux and merge if all is good.

ungb:
I'm looking into this now. going off test plan in https://github.com/atom/tree-view/pull/1179#issuecomment-360445015. I'll merge this once I'm done

ungb:
I ran into the following while dragging between two atom windows.

See gif for repro.

error


Repro steps:
Run the following command to create the folders and file
```
mkdir test1 test2
cd test1 && touch a b c && cd ../test2 && touch a b c && cd ..
atom test1 && atom test2
```

1. Drag file `a` from folder `test1` to folder `test2` in the other atom window


**Atom**: 1.23.3 x64
**Electron**: 1.6.15
**OS**: Mac OS X 10.12.6
**Thrown From**: [tree-view](https://github.com/atom/tree-view) package 0.222.0


### Stack Trace

Uncaught TypeError: Cannot read property 'collapse' of null

```
At /Users/ungb/github/packages/tree-view/lib/tree-view.coffee:1126

TypeError: Cannot read property 'collapse' of null
    at TreeView.module.exports.TreeView.onDrop (/packages/tree-view/lib/tree-view.coffee:1126:9)
    at /packages/tree-view/lib/tree-view.coffee:206:47)
```

UziTech:
Nice catch

ungb:
`Ctrl+click` doesn't work on linux.. it seems to just open up the context menu.  Shift works fine.  

@Ben3eeE does this work on windows?  I can spend some time testing windows today as well.

After the latest change, everything works great on mac!

Ben3eeE:
>@Ben3eeE does this work on windows?

@ungb It did when I tested. I have not tested on the most recent commit however.

ungb:
@Ben3eeE I think my issues is because I'm running on a VM on mac and it's doing something weird.  I'll take a look when I get home on my windows machine and then try to see if I spin up a VM on there for linux.  So far everything looks good, but every time I hit ctrl+click it opens up the context menu.  I'm noticing this on my VM and happens on the desktop on windows too.

Ben3eeE:
@ungb I have Ubuntu installed so I can test this on Linux if you run into more issues because of testing in a VM.

ungb:
I've tested this and everything looks good to me.  I'm going to merge this.  I'm going to follow up with the team later on the appveyor failure so we can get this into master.

UziTech:
hallelujah-gif-15

jonathan-dejong:
Any news on this? 
From what I can tell with Atom 1.26.1 it is not merged yet.

Ben3eeE:
@jonathan-dejong This is merged but it's not released yet because the tests are failing on `atom/atom` when upgrading the tree-view. You may want to subscribe to https://github.com/atom/atom/pull/17088 for updates on when it will be available in a released version of Atom.

PR #1180, opened 2017-09-12
Add UI for moving conflicting entries

### LINKED ISSUE #860: Failed to move entry .... already exists
### Description

Unable to drag-n-drop overwrite a file.
### Steps to Reproduce
1. Pick a file (in my case from the desktop).
2. Drag the file onto the folder structure of an open project.
3. You'll get the following error:

selection_002

**Expected behavior:** Either automatically overwrites or Atom prompts me to overwrite.

**Actual behavior:** Does nothing.
### Versions

Atom 1.7.4

I have re-opened this issue here based on my report [here](https://github.com/atom/atom/issues/11854).

This occurs for me on both Xubuntu 14.04 and 16.04; I do not believe it's reproducible on a Mac system (based on the report above).

50Wliu:
### Requirements

### Description of the Change

When an `EEXIST` error is encountered when drag-n-dropping files, Atom would previously bail out and refuse to move the file.  However, most OSes handle `EEXIST` by displaying a dialog and asking if the user would like to replace the existing file with the one being moved.  Atom now does the same.  This also works for multiple directories, where `readdirSync` will be performed on the directory and each entry will be moved.

### Alternate Designs

None.

### Benefits

The decision whether to overwrite files when drag-n-dropping can now be done inside of Atom.

### Possible Drawbacks

I'm not sure there are any.

### Applicable Issues

Fixes #860


50Wliu:
conflicting-files-ui
Cancelling will stop any further files from the same move event from being moved.  For example, if you're moving 5 conflicting files and you cancel at the 3rd file, the 3rd-5th files won't be moved.

50Wliu:
I think this is ready for general review + testing, though I may continue polishing and adding tests during the next few days.

Quite disappointed how difficult it was to test this.  It should become much easier once the files are converted to JS and async/await becomes available.

brycefranzen:
When will this issue be fixed? Still throws "failed to move entry" in latest version of atom.io :/  Would like to see this resolved.

50Wliu:
Note: the dialog box is currently sync, due to the lack of async/await in Coffeescript...

matt3224:
Any news on this?

50Wliu:
It's waiting for someone from the team to review it.

matt3224:
26 pull requests makes me think we are in for a wait. Tree view is a critical part of the Atom experience, it needs more love!

lib/tree-view.coffee
```diff
@@ -846,9 +846,49 @@ class TreeView
...
+        return @moveConflictingEntry(initialPath, newPath, newDirectoryPath)
+      else
+        @emitter.emit 'move-entry-failed', {initialPath, newPath}
+        atom.notifications.addWarning("Failed to move entry #{initialPath} to #{newDirectoryPath}", detail: error.message)
+
+    return true
+
+  moveConflictingEntry: (initialPath, newPath, newDirectoryPath) =>
+    try
+      if fs.isFileSync(initialPath)
```
rafeca: You may want to check if it's not a directory instead, to avoid handling symlinks as directories:

```suggestion
      unless fs.isDirectorySync(initialPath)
```

50Wliu:
> to avoid handling symlinks as directories

Tried to add a test for this...turns out renaming duplicate symlinks on Windows throws an EPERM error rather than an EEXIST, so that's outside of the scope of this PR 🤷‍♂.

PR #1182, opened 2017-09-15
Only destroy editors that are not modified

### LINKED ISSUE #1051: Deleting a file/folder closes associated files even with unsaved changes
If you delete a file (or a folder containing a file) that is open in the editor, tree-view will close it when the file is deleted. However, it does this even if the file had unsaved changes.

50Wliu:
### Requirements

### Description of the Change

This PR contains three bugfixes when deleting files through the Tree View.
1. Deleting a file would close its associated editor(s), even if there was unsaved content.
2. Deleting a path would cause other editors that _started with_ the deleted path to be destroyed.  For example, deleting `test` would destroy both `test` and `test1`'s editors.  The original intention was to destroy files under a deleted directory, but it was lacking appropriate checks to make sure that the deleted path was a) actually a directory, and b) that it wasn't inadvertently destroying a sibling folder's editors (e.g. deleting `lib/` would destroy editors under both `lib/` and `lib1/`).
3. The `entry-deleted` event passed its argument under the `path` variable, which I believe would override the global `path` require.  It and its companion events now pass the deleted path as `pathToDelete`.

### Alternate Designs

There probably are other alternate designs, but I don't see any.  I started at `editor.isModified()` and continued from there.

### Benefits

The primary benefit is that data loss no longer occurs when deleting files that have unsaved content.  The other benefit is some tightening of the editor-removal code.

### Possible Drawbacks

I had to add two new events: `onWillDeleteEntry` and `onDeleteEntryFailed`, as by the time `onEntryDeleted` is called `editor.isModified()` will always return true, due to its underlying file already having been deleted.

### Applicable Issues

Fixes #1051

### DIFF
--- a/lib/tree-view.coffee
+++ b/lib/tree-view.coffee
@@ -41,6 +41,7 @@ class TreeView
     @ignoredPatterns = []
     @useSyncFS = false
     @currentlyOpening = new Map
+    @editorsToDestroy = []
 
     @dragEventCounts = new WeakMap
     @rootDragAndDrop = new RootDragAndDrop(this)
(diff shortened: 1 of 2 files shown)


50Wliu:
One package uses `onEntryDeleted`: https://github.com/MonochromeSun/atom-together/blob/d9034f3725b2d3d37c39706f1da273f6b902b520/lib/init.coffee#L184

50Wliu:
Specs are up and running!  I've confirmed that all four of them fail on master but pass on this PR.

iolsen:
👍 Good fix. The new events seem reasonable to me.

50Wliu:
:+1: I'll create a PR on atom-together then.

50Wliu:
Merging.  I think it's safe to break one package to prevent data loss.

PR #1195, opened 2017-11-01
Drop `event-kit` from package dependenciess

Alhadis:
See the [`relevant `discussion](https://github.com/atom/archive-view/pull/47#issuecomment-340877566) here Basically, using eval-time constructors has a negative performance:

https://github.com/atom/archive-view/pull/47#issuecomment-341142488

/cc @nathansobo


nathansobo:
:heart:

50Wliu:
:100:

PR #1198, opened 2017-11-03
Add specs for element-icons service

Alhadis:
[As requested](https://github.com/atom/tree-view/pull/1146#issuecomment-340806545) by @nathansobo while reviewing #1146.

I'm sorry this took longer than it should have. I made the mistake of doing The Right Thing™ by collating icon-related specs into a single test-file, but forgot how flaky the current specs are. 😢 So I bailed on that and swore an oath to destroy both Jasmine and CoffeeScript in a hellish inferno. One day.

### DIFF
--- a/spec/file-icons-spec.coffee
+++ b/spec/file-icons-spec.coffee
@@ -1,12 +1,14 @@
+# vim: ts=2
 DefaultFileIcons = require '../lib/default-file-icons'
 getIconServices = require '../lib/get-icon-services'
+{Disposable} = require 'atom'
 
 describe 'IconServices', ->
-  describe 'FileIcons', ->
-    afterEach ->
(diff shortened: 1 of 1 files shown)


Alhadis:
No I don't use Vim, I added that modeline so I didn't have to change my config. There's a bug with some package where the indentation settings reset after saving, and I don't know why, [so I have this](https://github.com/Alhadis/Atom-PhoenixTheme/blob/52dea88b9dd43d2bccc0ccbee668fc81fa891ff9/src/index.js#L57-L60).

Should probably make a package for that or something. 😀

nathansobo:
✨ It didn't feel like it took a long time to me FWIW. I've been super pleased with the entire interaction with you over merging in these features. Nice work and thank you.

PR #1201, opened 2017-11-11
Ignore untitled editors when deleting folders

50Wliu:
### Requirements

### Description of the Change

When deleting folders, we loop over all open text editors and remove all that are within the folder-to-be-deleted that are not modified.  To do that, we simply compare paths.  However, Untitled editors have `undefined` paths and I forgot to account for that before comparing the path.  This PR adds an existence check to prevent an error being thrown when Untitled editors are present.

The code is now more-or-less identical to the corresponding `onWillMoveEntry` callback, which did have the existence check (and I believe was added afterwards).

### Alternate Designs

None.

### Benefits

Trying to delete a folder while having an Untitled editor open will no longer throw an error.

### Possible Drawbacks

None.

### Applicable Issues

Fixes atom/atom#16147

### DIFF
--- a/lib/tree-view.coffee
+++ b/lib/tree-view.coffee
@@ -103,12 +103,14 @@ class TreeView
       if fs.isDirectorySync(pathToDelete)
         pathToDelete += path.sep # Avoid destroying lib2's editors when lib was deleted
         for editor in editors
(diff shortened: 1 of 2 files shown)


50Wliu:
Appveyor crashes are unrelated.

Issue #1203, opened 2017-11-11
Crashing Windows test: when the project is a .git folder it does not create the tree view


50Wliu:
https://ci.appveyor.com/project/Atom/tree-view/build/444

Can be reproduced with the following three lines only, assuming that the Tree View is enabled and visible:
```coffee
dotGit = path.join(temp.mkdirSync('repo'), '.git')
fs.makeTreeSync(dotGit)
atom.project.setPaths([dotGit])
```
I can't repro in a normal Atom window though.

50Wliu:
Ok, the issue specifically appears to be the `temp.cleanup()` call that happens in the `afterEach` hook.

50Wliu:
...which is async.  `temp.cleanupSync()`, however, fails with `ENOTEMPTY` errors.

BinaryMuse:
This (the crashing) will be resolved by https://github.com/bruce/node-temp/pull/69 (and a dependency bump), but @50Wliu brings up some good points:

> Also two other questions regarding that. 1) Should we be using `temp.cleanupSync()`? and 2) why isn't cleanupSync called when the spec window exits?

50Wliu:
`temp.cleanupSync()` is automatically called on `process.exit`.  However it looks like closing the graphical spec runner doesn't fire that event.

iolsen:
We need to get the tree-view tests passing. Putting this at the top of the prioritized list.

daviwil:
On it!

daviwil:
I'm seeing a few issues related to using rimraf to delete paths that are being watched using chokidar or other watchers like pathwatcher:

- https://github.com/johnagan/clean-webpack-plugin/issues/38
- https://github.com/johnagan/clean-webpack-plugin/issues/59
- https://github.com/isaacs/rimraf/issues/119
- https://github.com/isaacs/rimraf/commit/d53235de863cfb00f691456288fcf3abbced7f6c

daviwil:
Reassinging to @50Wliu since he's already working on a fix.

PR #1209, opened 2017-12-06
Drop external folders onto the Tree View to create new project folders

### LINKED ISSUE #480: UX - Cannot drag project folder into tree-view
_From @jsphstls on April 27, 2015 16:6_

**To reproduce:**
Open Atom and clear any project folders or open files.
Open Finder and drag then drop a folder into the Atom tree-view.

**Current result:**
Project folder will not be added to tree-view.

**Expected result:**
Project folder would be added to tree-view.

**Suggested change:**
Allow project folders to be dragged into tree-view.

It is odd that you cannot add a project folder this way since the tree-view is where project folders are navigated. Instead, you can drag a folder into an editor pane, which is where individual files are edited.

_Copied from original issue: atom/atom#6535_

50Wliu:
### Requirements

### Description of the Change

Implements the popularly-requested feature #480.  Basically if no existing folder is selected the Tree View adds the entries in the data transfer event as new project folders.  There are a few rough edges with this implementation, such as but not limited to:
* Folders must be dropped to a blank section of the Tree View _after_ the last entry.  This means that if the Tree View scrolls you will be unable to create a project folder as the entries will get added to the last project.
  * I would like to add support to be able to create project folders when dropping them _between_ project folders, similar to how root drag-and-drop works currently.
* Cannot drag and drop folders onto the Tree View when no Tree View exists.

### Alternate Designs

No alternatives were considered.

### Benefits

External folders can now be dragged-and-dropped into the Tree View to be added as project folders.

### Possible Drawbacks

I don't believe there are any drawbacks with this other than the limitations addressed above.

### Applicable Issues

Fixes #480

### DIFF
--- a/lib/tree-view.coffee
+++ b/lib/tree-view.coffee
@@ -1056,6 +1056,9 @@ class TreeView
         # Drop event from OS
         for file in e.dataTransfer.files
           @moveEntry(file.path, newDirectoryPath)
+    else if e.dataTransfer.files.length
+      # Drop event from OS that isn't targeting a folder: add a new project folder
+      atom.project.addPath(entry.path) for entry in e.dataTransfer.files
 
   isVisible: ->
     @element.offsetWidth isnt 0 or @element.offsetHeight isnt 0
--- a/lib/root-drag-and-drop.coffee
+++ b/lib/root-drag-and-drop.coffee
@@ -2,7 +2,8 @@ url = require 'url'
 
 {ipcRenderer, remote} = require 'electron'
 
-_ = require 'underscore-plus'
+# TODO: Support dragging external folders and using the drag-and-drop indicators for them
+# Currently they're handled in TreeView's drag listeners
 
 module.exports =
 class RootDragAndDropHandler
@@ -120,7 +121,6 @@ class RootDragAndDropHandler
 
     {dataTransfer} = e
 
-    # TODO: support dragging folders from the filesystem -- electron needs to add support first
     return unless @isAtomTreeViewEvent(e)
 
     fromWindowId = parseInt(dataTransfer.getData('from-window-id'))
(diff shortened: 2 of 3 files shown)


lib/tree-view.coffee
```diff
@@ -1056,6 +1056,9 @@ class TreeView
         # Drop event from OS
         for file in e.dataTransfer.files
           @moveEntry(file.path, newDirectoryPath)
+    else if e.dataTransfer.files.length
```
liuderchi: Would `e.dataTransfer.files` be `undefined` ?

50Wliu: Good question.  `files` is  standard DataTransfer property that will always be available. https://developer.mozilla.org/en-US/docs/Web/API/DataTransfer/files

liuderchi: Thanks for the info. Any it says that `dataTransfer` exists on *all drag events*. Good News.

50Wliu:
I looked into implementing drag-and-drop indicators.  It will be harder than I thought because then you'll have to differentiate between moving external folders _into_ existing projects or whether to create a new project folder for it.  I'll punt that for a later PR.

here:
Note that as of 2018-03-06 , this request remains unreleased in the Atom core editor with the tree-view core package at v221.3 or v222 based on https://github.com/atom/tree-view/releases

liuderchi:
Share a shell snippet for checking releases

```sh
# atom latest releases and corresponding tree-view version
$ curl https://api.github.com/repos/atom/atom/releases/latest -s | grep 'tag_name'

  "tag_name": "v1.24.0",

$ curl https://raw.githubusercontent.com/atom/atom/v1.24.0/package.json -s | grep 'tree-view'

    "tree-view": "0.221.3",
```

50Wliu:
@here we're dealing with [tree-view spec failures](https://github.com/atom/tree-view/issues/1203) that need to be fixed before we feel comfortable updating it in Atom itself.  Sorry for the delay.

PR #1221, opened 2018-02-02
use path.join

UziTech:
### Description of the Change

use path.join instead of hard coding a slash

### Applicable Issues

https://github.com/atom/tree-view/pull/1179#discussion_r164581249

### DIFF
--- a/lib/tree-view.coffee
+++ b/lib/tree-view.coffee
@@ -869,7 +869,7 @@ class TreeView
       return
 
     entryName = path.basename(initialPath)
-    newPath = path.join(newDirectoryPath, entryName).replace(/\s+$/, '')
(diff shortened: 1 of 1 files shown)


lib/tree-view.coffee
```diff
@@ -869,7 +869,7 @@ class TreeView
       return
 
     entryName = path.basename(initialPath)
-    newPath = "#{newDirectoryPath}/#{entryName}".replace(/\s+$/, '')
+    newPath = path.join(newDirectoryPath, entryName).replace(/\s+$/, '')
```
50Wliu: I also don't think the replace is necessary.  I haven't seen any other case where we trim paths.

Issue #796, opened 2016-04-06
Tree View Unhides When Switching Between Split Panes


eviljoe:
I have my tree view hidden when I am using split panes too add more horizontal space.  I also have the "Auto Reveal" option enabled.  Whenever I move between split-panes, the tree view becomes visible again.  The auto reveal option should not make the tree view visible if it is hidden.

**Steps To Recreate**
1. Enable the "Auto Reveal" tree view option
2. Open a file
3. Split Atom's view using the `Pane: Split Right` command
4. Open a different file in the right pane
5. Hide the tree view using the `Tree View: Toggle` command
6. Click between the different open panes
   - The tree view will become visible

This behavior happens in safe mode too.

durz:
This also happens when switching between tabs, not just switching between split-panes.

Ben3eeE:
Is this a duplicate of https://github.com/atom/tree-view/issues/738? I can reproduce all cases in both issues with Auto reveal on but #738 does not mention Auto reveal.

ZuBB:
pls fix this

Arcanemagus:
@Ben3eeE This is a direct duplicate of https://github.com/atom/tree-view/issues/738#issuecomment-223133421, and fixing this/that comment would likely fix #738 as a whole since I'm guessing when the filer of #738 said "opens" they meant "becomes un-hidden".

vraravam:
My usecase is slightly different: I want the tree-view to be open all the time, but with the "auto-reveal" feature, the tree-view captures focus. When I open a file or switch the active text editor, all I was to happen is to reveal the file, but the focus should still remain on the text editor.

mupchrch:
This same issue happens when you create a new, empty editor (File > New File). It is annoying to have the tree-view open on its own for a file that isn't even in the tree.

alejandrolacasa:
What's the status of this? It is very annoying when working on small screens... I cannot have 3 panes at the same time because of the size of the tree file :(

Ben3eeE:
@alejandrolacasa Did you try disabling the `Auto Reveal` setting in the tree-view package settings?

alejandrolacasa:
Yes, sure, but then the files are not "revealed" when opening them or switching between tabs. The behaviour I expect is that the files are revealed when switching between them, even if the tree is hidden, but not to make the tree appear if I hidden it on purpose.

[...]

PR #1225, opened 2018-02-22
Use @atom/temp instead of node-temp

### LINKED ISSUE #1203: Crashing Windows test: when the project is a .git folder it does not create the tree view

daviwil:
This change switches the use of the node-temp module to our
@atom/temp fork so that we can benefit from a newer rimraf
version that will resolve some recurring test failures on
AppVeyor.

Resolves #1203.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -16,6 +16,7 @@ environment:
   matrix:
   - ATOM_CHANNEL: stable
   - ATOM_CHANNEL: beta
+  - ATOM_CHANNEL: dev
 
 install:
   - ps: Install-Product node 6
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -16,7 +16,6 @@ environment:
   matrix:
   - ATOM_CHANNEL: stable
   - ATOM_CHANNEL: beta
-  - ATOM_CHANNEL: dev
 
 install:
   - ps: Install-Product node 6
--- a/package.json
+++ b/package.json
@@ -12,7 +12,6 @@
   "dependencies": {
     "fs-plus": "^3.0.0",
     "minimatch": "~0.3.0",
-    "nan": "2.7.0",
     "pathwatcher": "^8.0.0",
     "@atom/temp": "~0.8.4",
     "underscore-plus": "^1.0.0"
--- a/package.json
+++ b/package.json
@@ -14,7 +14,7 @@
     "minimatch": "~0.3.0",
     "nan": "2.7.0",
     "pathwatcher": "^8.0.0",
-    "temp": "~0.8.1",
+    "@atom/temp": "~0.8.4",
     "underscore-plus": "^1.0.0"
   },
   "devDependencies": {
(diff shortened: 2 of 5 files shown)


daviwil:
Anything else in #1203 that I should address?  @50Wliu any suggestion on where I should call `temp.cleanupSync()` so that the graphical runner cleans up correctly?

daviwil:
Still seeing the same problem.  I think there's a conflict between rimraf and pathwatcher, will add more context to the original issue.

daviwil:
Currently trying to switch over to Atom's new `watchPath` API to avoid the issues with `pathwatcher` locking recursive directory paths, should have an update later today!

Ben3eeE:
This will also close https://github.com/atom/tree-view/issues/796 again because we can upgrade tree-view on atom/atom again.

50Wliu:
Interestingly, specs no longer crash on my machine even without these changes.  I wonder if there's a change in Atom master that helped with the crashes?

daviwil:
Hmmm, I wonder if we can run tree-view specs against Atom Dev in CI to verify?

50Wliu:
> Hmmm, I wonder if we can run tree-view specs against Atom Dev in CI to verify?

Still crashes :/.

50Wliu:
Ok, removing the cleanupSync calls seems to have fixed the issue.  @daviwil you think that's an acceptable solution?

daviwil:
Potentially acceptable to get the AppVeyor tests running again, but we definitely need to fix the underlying issue because it's had user impact for a long time according to some really old bugs I saw.  Looks like you're making progress on the `watchPath` conversion, I'd be interested to see if that actually fixes the core problem enough to get the tests passing :)

Thanks again for taking this on!

daviwil:
Thanks for the quick fix @50Wliu!  Let's go with this for now to get the build green and then finish up the `watchPath` migration later.

50Wliu:
/cc @atom/maintainers:

This PR contains a "quick-fix" to get Appveyor tests passing again.  It does so by 1) switching over to `@atom/temp` (which may or may not be required) and 2) removing the manual cleanup step after each test.

Downsides of this PR is that at least on Windows, automatic cleanup does not occur for the GUI-based spec runner, leaving many files in the temp directory.  Realistically, this should not be a concern as temp directories are cleaned regularly in the first place, and because there is no preserved state on CI.

A more complete fix which addresses the root cause (node-pathwatcher aggressively locking directories and preventing them from being removed) is in progress at #1232.

damieng:
Ship it! :)

PR #1248, opened 2018-05-02
Add command for move

elliotaplant:
### Requirements

### Description of the Change

As per Atom issue [#3212](https://github.com/atom/atom/issues/3212), there is a desire for a hotkey to rename/move the current file if the file has been saved outside the scope of the root dir and it is not visible in the tree view. A [PR](https://github.com/atom/tree-view/pull/549) has already been submitted for this issue, but it had one outstanding change required and hasn't been updated in over a year.

This change adds the hotkey for the move command within the atom editor for darwin, windows, and linux systems.

### Alternate Designs

No alternate designs were considered.

### Benefits

The benefit of this change is that users will be able to rename/move files they are editing with a hotkey even if they can't find those files in their tree-view.

### Possible Drawbacks

One possible drawback is if someone expected this hotkey to do nothing, they would be unpleasantly surprised to find that they are renaming their files with the keybinding `alt-shift-m`. This won't be a problem for people who have overridden this keybinding.

### Applicable Issues

gitub.com/atom issue:
https://github.com/atom/atom/issues/3212

currently open PR:
https://github.com/atom/tree-view/pull/549

### DIFF
--- a/keymaps/tree-view.cson
+++ b/keymaps/tree-view.cson
@@ -86,3 +86,9 @@
 '.tree-view-dialog atom-text-editor[mini]':
   'enter': 'core:confirm'
   'escape': 'core:cancel'
+
+'.platform-win32 atom-text-editor'
+  'F12': 'tree-view:move'
+
+'atom-text-editor':
+  'alt-shift-m': 'tree-view:move'
--- a/keymaps/tree-view.cson
+++ b/keymaps/tree-view.cson
@@ -87,7 +87,7 @@
   'enter': 'core:confirm'
   'escape': 'core:cancel'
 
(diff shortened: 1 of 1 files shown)


lee-dohm:
I don't think this achieves the stated goal of always renaming the current buffer since there is an option to prevent tree-view from always tracking the currently open buffer. This isn't going to be a simple change of adding a keybinding.

elliotaplant:
Good to know. Would it be worth warning the user if they tried to change a filename but tree-view wasn't tracking that file?

lee-dohm:
Not if the idea for the keybinding is to always affect the "current buffer".

PR #1251, opened 2018-05-25
Enable gitignore when opened a subdirectory

### LINKED ISSUE #100: Not reading .gitignore files properly
It seems that for me, when files are added to ignored directories, Tree View still thinks that the parent directory of those files added (but ignored) have added files in them. This makes it confusing when I actually add files that belong in VCS. Notice below: If you're at all familiar with Laravel, the directories highlighted in green are ignored... therefore, they should not be considered when new files are added inside of them. Make sense?

Tree View

denis-sokolov:
### Description of the Change

The use case that required the previous conditional is opening an ignored subdirectory.
The updated check handles this use case, but also fixes opening non-ignored subdirectories.

https://github.com/atom/tree-view/issues/100#issuecomment-41324015

### Alternate Designs

Even more exposed settings in the UI does not help the user.

### Benefits

Enables gitignore to work when Atom is opened on a subdirectory.

### Possible Drawbacks

The behavior changes for users who open a non-ignored subdirectory. The intent is to improve the behavior, but some users may have been relying on the previous functionality.

### Applicable Issues

Closes #100.

### DIFF
--- a/lib/directory.js
+++ b/lib/directory.js
@@ -185,7 +185,7 @@ class Directory {
   isPathIgnored (filePath) {
     if (atom.config.get('tree-view.hideVcsIgnoredFiles')) {
       const repo = repoForPath(this.path)
-      if (repo && repo.isProjectAtRoot() && repo.isPathIgnored(filePath)) return true
+      if (repo && !repo.isPathIgnored(this.path) && repo.isPathIgnored(filePath)) return true
     }
 
     if (atom.config.get('tree-view.hideIgnoredNames')) {
--- a/spec/tree-view-package-spec.coffee
+++ b/spec/tree-view-package-spec.coffee
@@ -2994,19 +2994,36 @@ describe "TreeView", ->
         atom.config.set('tree-view.hideIgnoredNames', true)
         expect(Array.from(treeView.element.querySelectorAll('.file')).map((f) -> f.textContent)).toEqual([])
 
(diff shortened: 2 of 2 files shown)


50Wliu:
@denis-sokolov can you please add some tests for this modified behavior?  Thank you!

denis-sokolov:
I’d be happy to, but I look at the tree-view-package-spec file and it’s a bit overwhelming. :-(
https://github.com/atom/tree-view/blob/master/spec/tree-view-package-spec.coffee

denis-sokolov:
I will try to add tests to this one myself now.

denis-sokolov:
I’ve done the closest I could in c8a50d7, @50Wliu. One of my tests is failing, and I did not finish investigating why. However, I’ve noticed the expected functionality seems to be available in the production version of tree-view (0.222.0), where the ignored files are now not hidden, but de-emphasized. This kind of defeats the purpose of the PR, so I will not continue further for now. Thanks for your help throughout, though!

Ben3eeE:
@denis-sokolov is this something that is not working in the latest master version of tree-view but is working in 0.222.0?

50Wliu:
@Ben3eeE just tried on master and I'm seeing the same behavior @denis-sokolov is describing.  Gitignored files are greyed out but still visible when you open a non-ignored subdirectory.

PR #1252, opened 2018-05-25
Enable hideIgnoredNames with hideVcsIgnoredFiles

### LINKED ISSUE #489: Ignored names ignored when there is no .git directory in the project.
Often we use ignored names on local projects not ready for git just to hide files that will be cleaned, are temporary or otherwise not needed for the project by default, but ignored names is ignored when there is no .git directory inside of the project folder.

It would be nice if this weren't the case or if it's meant to be the case allow us to disable that requirement.

denis-sokolov:
### Description of the Change

Interpreting English, when both “hide ignored names” and “hide vcs ignored files” are enabled, I would expect both to apply, and the union of the two options of files to be hidden.

Before this commit, however, the “hide vcs ignored files” would override the “hide ignored names” for no apparent benefit.

### Alternate Designs

No alternate designs were considered.

### Benefits

The behavior will more closely match the user’s expectation based on the shared interface in English.

### Possible Drawbacks

Some users may have been relying on the gitignored files overriding the custom “ignored names” setting.

### Applicable Issues

Fixes #489

### DIFF
--- a/lib/directory.js
+++ b/lib/directory.js
@@ -185,7 +185,7 @@ class Directory {
   isPathIgnored(filePath) {
     if (atom.config.get('tree-view.hideVcsIgnoredFiles')) {
       const repo = repoForPath(this.path)
-      if (repo && repo.isProjectAtRoot() && repo.isPathIgnored(filePath)) return true;
+      if (repo && repo.isProjectAtRoot() && repo.isPathIgnored(filePath)) return true
     }
 
     if (atom.config.get('tree-view.hideIgnoredNames')) {
--- a/lib/directory.js
+++ b/lib/directory.js
@@ -185,7 +185,7 @@ class Directory {
   isPathIgnored(filePath) {
     if (atom.config.get('tree-view.hideVcsIgnoredFiles')) {
       const repo = repoForPath(this.path)
-      return repo != null && repo.isProjectAtRoot() && repo.isPathIgnored(filePath)
+      if (repo && repo.isProjectAtRoot() && repo.isPathIgnored(filePath)) return true;
     }
 
     if (atom.config.get('tree-view.hideIgnoredNames')) {
--- a/spec/tree-view-package-spec.coffee
+++ b/spec/tree-view-package-spec.coffee
@@ -2939,6 +2939,18 @@ describe "TreeView", ->
         atom.config.set("tree-view.hideVcsIgnoredFiles", false)
         expect(Array.from(treeView.element.querySelectorAll('.file')).map((f) -> f.textContent)).toEqual(['.gitignore', 'ignored.txt'])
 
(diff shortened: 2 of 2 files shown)


50Wliu:
Fixes #489.

Could you please add some tests for this PR as well?

lee-dohm:
As @50Wliu mentioned, we need tests to accept this PR. Additionally, with the addition of the `if (expr) return true` construct, this now returns `undefined` instead of `false`, which is an undesired change. We would want this to maintain the behavior of only returning `true` or `false`.

Thanks for your effort on this!

50Wliu:
I believe with this change the logic will still pass down to the final `return false` call at https://github.com/atom/tree-view/pull/1252/files#diff-fa19a8d46cd6fbf5a016c0719da3b487R195, so I think the function will still only return a boolean.

denis-sokolov:
@lee-dohm, not returning false and falling through to other options is the purpose of this PR.

denis-sokolov:
I’d be happy to add tests, but I look at the tree-view-package-spec file and it’s a bit overwhelming. :-(
https://github.com/atom/tree-view/blob/master/spec/tree-view-package-spec.coffee

lee-dohm:
We would like to be in a position to mentor people on writing tests for things. Unfortunately, we're not currently there at this point. We've got some mid-term plans to hopefully change that but for now we're having to close PRs if they can't be supplied with tests. I have labeled the PR as `help-wanted` so that people searching for things that they can help with can easily find it. Keep in mind, just because we are closing it now doesn't mean that it can't be reopened by yourself or us in the future if circumstances change.

Thanks for the effort you've put into it and if there's anything I can help answer, you can find me in the [Atom Slack team](https://discuss.atom.io/t/join-us-on-slack/16638?source_topic_id=25406).

denis-sokolov:
Wait, @lee-dohm. Consider that the current behavior is less meaningful and similarly untested. Choosing between two untested behaviors, would you prefer the one that is more or less functional?

50Wliu:
@denis-sokolov added the tests for you.  Hopefully that helps give you an idea of how to write them :).

denis-sokolov:
Thanks, @50Wliu! That test commit is very helpful and straight-forward, I suppose I have the moral obligation to do #1251 now myself. :-)

If you find this feedback useful, consider that the test code itself seems reasonable when viewed as one small structural bit. If the test spec was not a single 5k-line file, but a directory structure, with submodules and 100-line files, it would be much less daunting for me as a guest contributor.

50Wliu:
Yeah...it's kinda just grown into a very large mess that no one quite wants to refactor because of how much it is.  Ctrl+F'ing is a good way to try to figure out where to put your test though.

PR #1257, opened 2018-06-20
Copy files on drop if control is pressed

### LINKED ISSUE #703: Ctrl/Cmd-Drag and drop should copy the selected file / directory

50Wliu:
### Requirements

### Description of the Change

For consistency with OS behavior, copy files when dropping if control is pressed, rather than moving them.

### Alternate Designs

None?

### Benefits

Consistency with file managers such as Windows Explorer.

### Possible Drawbacks

Unknown

### Applicable Issues

Closes #703
Supersedes and closes #800

* [x] Make it work
  * [x] Add specs
* [x] Allow files to be copied to the same directory
  * [x] Add specs
* [ ] Regression testing

### DIFF
--- a/lib/tree-view.coffee
+++ b/lib/tree-view.coffee
@@ -718,9 +718,6 @@ class TreeView
   # Public: Paste a copied or cut item.
   #         If a file is selected, the file's parent directory is used as the
   #         paste destination.
-  #
-  #
-  # Returns `destination newPath`.
   pasteEntries: ->
     selectedEntry = @selectedEntry()
     cutPaths = if window.localStorage['tree-view:cutPath'] then JSON.parse(window.localStorage['tree-view:cutPath']) else null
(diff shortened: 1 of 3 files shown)


Diesector:
I would love to get my hands on this to test, could I manually replace the spec/lib files to get this update?

50Wliu:
Verification checklist:
* [ ] Copy and pasting a file to the same directory pastes the file with a `0` behind the basename
* [ ] Cut and pasting a file to the same directory pastes the file (you should notice no change)
* [ ] Copy and pasting a directory to the same directory pastes the directory with a `0` appended (recursive copy)
* [ ] Cut and pasting a directory to the same directory pastes the directory (you should notice no change)
* [ ] Copy and pasting a file and directory to the same directory pastes both successfully, with numbers appended as appropriate
* [ ] Cut and pasting a file and directory to the same directory pastes both successfully (you should notice no change)
* [ ] Copy and pasting a file to a different directory pastes the file succesfully
* [ ] Cut and pasting a file to a different directory pastes the file successfully and removes the file from the original location
* [ ] Copy and pasting a directory to a different directory pastes the directory successfully
* [ ] Cut and pasting a directory to a different directory pastes the directory successfully
* [ ] Copy and pasting a file and directory to a different directory pastes both successfully
* [ ] Cut and pasting a file and directory to a different directory pastes both successfully

Now repeat the above with drag-and-drop (replacing "copy" with ctrl-drag, and "cut" with drag).

UziTech:
It would be great to get this in the next version of atom along with the other multi-drag improvements 👍

50Wliu:
> Does that reflect the current status of things, or have you verified that everything is working?

@nathansobo I've verified that everything's working on Windows, but not on macOS or Linux.

nathansobo:
@50Wliu It looks like the tests are hanging after merging in master. This seems to be due to a dialog prompt when pasting files that conflict with existing directory entries. Maybe something changed on master causing the prompt to be displayed in a situation where it wasn't previously. We'll kick this back to you, since you have more context. Please feel free to ping me again when you get a green build.

50Wliu:
Ugh, dialog prompts. I'm guessing something to do with #1180.

50Wliu:
@nathansobo All right, specs are 💚 again.

#### What was the problem?
In this PR, I extracted much of the logic in `pasteEntries` to the helper methods `copyEntry` and `moveEntry`. Simultaneously, in #1180, I updated `moveEntry` to show a dialog box on conflicting entries. Prior to this PR, `moveEntry` was _only_ used for drag-and-drop, and so the regular cut/paste specs were still expecting to error on conflicting entries. Now that all the move logic is centralized in one function, those specs began to hang due to the appearance of an unexpected dialog prompt.

#### The fix
I've essentially copied over all the specs I wrote in #1180 and adjusted them for the cut functions. Yes, that means we now have (almost) duplicate spec coverage for `moveEntry`, and we should probably rewrite those specs to remove the duplication. However, that would be out of the scope of this PR.

nathansobo:
I've tested everything manually on macOS per your checklist. Since this all relies on Node APIs to interface with the file system and we have thorough test coverage, I'm going to skip testing on Linux because I can't see how things could be so different on that platform that this would be broken while still building green on Travis. Perhaps famous last words.

nathansobo:
Thanks for your stewardship @50Wliu! ❤️ :zap:

Dzemy:
Maybe change the icon when you holding control will be cool :) , i first think there is no way how to copy file.

PR #1266, opened 2018-07-04
Use shell.showItemInFolder to show files

### LINKED ISSUE #583: Show in File Manager doesn't select the file in Fedora 22, Gnome, Nautilus system.
Hi I'm having an uncomfortable bug when I use the `Show in File Manager` option. The file manager opens in the correct folder path but doesn't select or highlight that file.
I'm using Atom 1.0.19 with Fedora 22, GNOME Shell 3.16.3, GNOME nautilus 3.16.2. Thanks!

### LINKED ISSUE #685: Custom file manager for "Show in File Manager" action
Hi,

I asked at [discuss.atom.io](https://discuss.atom.io/t/how-to-adjust-the-file-manager-used-by-atom-when-selecting-show-in-file-manager/24027/2) how to set the file manager used for the "Show in File Manager" action for this package and I was told that this feature does not currently exist, hence it is one I would have to request here in order to get it. I am running on the Linux distros Manjaro and Sabayon Linux and whenever I use this action in the treeview I open the folder in PCManFM. Setting my file manager using `xdg-mime default nautilus.desktop inode/directory` does not change the file manager used to open the folder. So I know that this is an issue the developers of this wonderful package will need to address in order for it to be fixed. 

Thanks for your time,
Brenton

50Wliu:
### Requirements

### Description of the Change

Electron's [`shell`](https://electronjs.org/docs/api/shell) module has a method [`shell.showItemInFolder`](https://electronjs.org/docs/api/shell#shellshowiteminfolderfullpath) to automatically open files in the OS's file manager.  I'm not sure why we decided to go ahead with our implementation instead of using that, but it makes much more sense to delegate to Electron instead of maintaining our own version.

### Alternate Designs

None.

### Benefits

`shell.showItemInFolder` will use the default file manager, rather than the one hardcoded by tree-view.  For example, on Windows, while the default file manager is commonly `explorer.exe`, that is not always the case.  This change also reduces the amount of code that tree-view needs to maintain.

### Possible Drawbacks

I have changed the menu text to always refer to a generic "File Manager", as it is no longer possible to know exactly which file manager will be used.

### Applicable Issues

Supersedes and closes #889.
_Should_ fix #685.
_Should_ fix #583.

Needs testing on macOS and Linux.

### DIFF
--- a/lib/tree-view.coffee
+++ b/lib/tree-view.coffee
@@ -537,76 +537,16 @@ class TreeView
           @emitter.emit 'move-entry-failed', {initialPath, newPath}
       dialog.attach()
 
-  # Get the outline of a system call to the current platform's file manager.
-  #
-  # pathToOpen  - Path to a file or directory.
-  # isFile      - True if the path is a file, false otherwise.
-  #
-  # Returns an object containing a command, a human-readable label, and the
-  # arguments.
-  fileManagerCommandForPath: (pathToOpen, isFile) ->
-    switch process.platform
-      when 'darwin'
-        command: 'open'
-        label: 'Finder'
-        args: ['-R', pathToOpen]
-      when 'win32'
-        args = ["/select,\"#{pathToOpen}\""]
-
-        if process.env.SystemRoot
-          command = path.join(process.env.SystemRoot, 'explorer.exe')
-        else
-          command = 'explorer.exe'
-
-        command: command
-        label: 'Explorer'
-        args: args
-      else
-        # Strip the filename from the path to make sure we pass a directory
-        # path. If we pass xdg-open a file path, it will open that file in the
-        # most suitable application instead, which is not what we want.
-        pathToOpen =  path.dirname(pathToOpen) if isFile
-
-        command: 'xdg-open'
-        label: 'File Manager'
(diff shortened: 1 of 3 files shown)


Arcanemagus:
> I'm not sure why we decided to go ahead with our implementation instead of using that

It looks like that was added in https://github.com/atom/tree-view/pull/47, which was well after the Electron stuff was added in https://github.com/electron/electron/commit/fbad5bc822b862100184f2e6e7d7243f024d4e6e so... 🤷‍♂️.

jasonrudolph:
> Needs testing on macOS and Linux.

I ran through the following tests on macOS and Ubuntu:

- [x] Inside editor, use context menu to show file in OS file manager
- [x] Inside editor, use command palette to show file in OS file manager
- [x] Inside tree-view, use context menu to show selected file in OS file manager
- [x] Inside tree-view, use command palette to show selected file in OS file manager

Everything functioned as expected. 👌

winstonford:
An alternative menu text might be "Reveal in Folder" 
( as seen in the https://github.com/fabiospampinato/notable )

But really, its hard to beat the concisivity (homemade word) of  "Show in Folder"

PR #1232, opened 2018-03-07
[WIP] Use atom.watchPath instead of PathWatcher

50Wliu:
### Requirements

* Filling out the template is required. Any pull request that does not include enough information to be reviewed in a timely manner may be closed at the maintainers' discretion.
* All new code requires tests to ensure against regressions

### Description of the Change

Replaces [pathwatcher](https://github.com/atom/node-pathwatcher) with [atom.watchPath](https://atom.io/docs/api/v1.24.0/PathWatcher).

### Alternate Designs

None.

### Benefits

A better watching library that doesn't lock files everywhere on Windows.

### Possible Drawbacks

TODO

### Applicable Issues

TODO

PR #1269, opened 2018-07-06
Remove realpathCache argument in fs.realpath

50Wliu:
### Requirements

### Description of the Change

Node 6 removed the `cache` argument from `fs.realpath` as it switched to a much faster native implementation. See https://github.com/nodejs/node/pull/3594 for details. As Atom is now on Node 8, this argument is currently sitting there doing...nothing.

### Alternate Designs

None.

### Benefits

Removal of a confusing argument from a bygone era.

### Possible Drawbacks

None.

### Applicable Issues

None, though I caught this when working on #1232.

### DIFF
--- a/lib/directory.js
+++ b/lib/directory.js
@@ -115,7 +115,7 @@ class Directory {
         this.lowerCaseRealPath = this.realPath.toLowerCase()
       }
     } else {
-      fs.realpath(this.path, {}, (error, realPath) => {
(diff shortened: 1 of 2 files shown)


Arcanemagus:
Would it be clearer to simply omit the `options` object?

50Wliu:
Yes, it would 😀.  Mind was on autopilot there.

PR #1270, opened 2018-07-09
Add option in config to always recursively collapse

brycefranzen:
### Description of the Change
Adds option in settings (package.json) for the user to select "Always Recursive Collapse."
This makes it so that whenever a parent folder is collapsed in the tree-view, all child folders are ALWAYS also collapsed.
In essence, it adds a boolean override for "isRecursive."

### Alternate Designs
No other design options considered. Simple setting addition.

### Benefits
This allows users to specify their preference as to how the folders in the tree-view should be collapsed.
This should be an option since several other editors have this functionality by default.

### Possible Drawbacks
None (defaults to false, so same default functionality will not change)

### Applicable Issues
None

### DIFF
--- a/package.json
+++ b/package.json
@@ -95,10 +95,10 @@
       "default": false,
       "description": "When opening a file, always focus an already-existing view of the file even if it's in a another pane."
     },
-    "alwaysRecursiveCollapse": {
+    "alwaysCollapseRecursively": {
       "type": "boolean",
       "default": false,
-      "description": "When collapsing a folder, always collapse the children as well"
+      "description": "When collapsing a folder, always collapse all descendants as well"
     }
   }
 }
--- a/package.json
+++ b/package.json
@@ -94,6 +94,11 @@
       "type": "boolean",
       "default": false,
       "description": "When opening a file, always focus an already-existing view of the file even if it's in a another pane."
+    },
+    "alwaysRecursiveCollapse": {
+      "type": "boolean",
+      "default": false,
+      "description": "When collapsing a folder, always collapse the children as well"
     }
   }
 }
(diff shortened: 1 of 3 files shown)


lee-dohm:
We would consider merging this if tests are added.

package.json
```diff
@@ -94,6 +94,11 @@
       "type": "boolean",
       "default": false,
       "description": "When opening a file, always focus an already-existing view of the file even if it's in a another pane."
+    },
+    "alwaysRecursiveCollapse": {
```
thomasjo: I think this reads better as `alwaysCollapseRecursively`. /cc @atom/feedback

package.json
```diff
@@ -94,6 +94,11 @@
...
+      "type": "boolean",
+      "default": false,
+      "description": "When collapsing a folder, always collapse the children as well"
```
thomasjo: Maybe we should use "descendants" instead of "children" to be more correct? Better yet, "all descendants" to really drive home the recursive nature of this option.

thomasjo (changes requested):
As stated by @lee-dohm, this needs tests before we'll consider merging. I've also left some comments regarding the configuration schema changes.

package.json
brycefranzen: agreed. Updated.

package.json
brycefranzen: Agreed. Updated.

brycefranzen:
I have added testing. :)

spec/tree-view-package-spec.coffee
```diff
@@ -769,6 +769,39 @@ describe "TreeView", ->
           expect(child).not.toHaveClass 'expanded'
         expect(treeView.roots[0]).toHaveClass 'expanded'
 
+
+  describe "the alwaysCollapseRecursively config option", ->
+    it "defaults to unset", ->
+      expect(atom.config.get("tree-view.alwaysCollapseRecursively")).toBeFalsy()
+
+    describe "when an expanded directory is clicked", ->
```
50Wliu: I would be more explicit here and add another `describe` noting that this only occurs when the setting is true.

winstliu (changes requested):

existe-deja:
up :)

PR #1277, opened 2018-08-13
Reselect entries when updating roots

50Wliu:
### Requirements

### Description of the Change

Previously, whenever the roots were updated (resulting in the entire DOM being recreated), any selected entries wouldn't stay selected.  This would happen whenever projects were added or removed or certain tree-view settings were changed.  Now, whenever the roots are updated, the selected entries are maintained.

Additionally, if you select multiple entries before closing Atom, those same entries will remain selected the next time you open Atom instead of only the first entry.

### Alternate Designs

Well, it would certainly be nice if the entire DOM wasn't erased and recreated.

### Benefits

Entries stay more consistently selected.

### Possible Drawbacks

There is one caveat: after merging this PR, the lack of `state.selectedPaths` on the first run will cause all existing entry serialization (in `state.selectedPath`) to be disregarded, and the root will be selected.  I don't see this as a problem though, considering `updateRoots` blew away the selection anyway.

### Applicable Issues

I couldn't find any.

### DIFF
--- a/lib/tree-view.coffee
+++ b/lib/tree-view.coffee
@@ -132,7 +132,7 @@ class TreeView
       @[root.directory.path] = root.directory.serializeExpansionState() for root in roots
       this)(@roots)
     deserializer: 'TreeView'
-    selectedPaths: Array.from(@getSelectedEntries())?.map((entry) -> entry.getPath())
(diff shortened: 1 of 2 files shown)


50Wliu:
Before:
entry-selection-before

After:
entry-selection-after

lee-dohm:
Thanks for this! 💖

PR #1283, opened 2018-10-09
Prevent multiple instances of tree-view being opened

shadow-light:
### Description of the Change

Currently tree-view opens itself with a call to `atom.workspace.open` which by default: "only the active pane will be searched for an existing item"

When the left dock is split, only the bottom pane of that dock is searched, and if tree-view already exists in another pane then two instances of tree-view are opened. Initially the duplicate instance is blank, but upon restarting atom it causes atom to crash with `Error: The workspace can only contain one instance of item`.

### Alternate Designs

None

### Benefits

Atom doesn't crash

### Possible Drawbacks

None

### Applicable Issues

This should fix:

- https://github.com/atom/atom/issues/6121#issuecomment-335487266
- https://github.com/atom/atom/issues/14201

### DIFF
--- a/lib/tree-view-package.js
+++ b/lib/tree-view-package.js
@@ -28,7 +28,8 @@ class TreeViewPackage {
       const showOnAttach = !atom.workspace.getActivePaneItem()
       this.treeViewOpenPromise = atom.workspace.open(treeView, {
         activatePane: showOnAttach,
-        activateItem: showOnAttach
+        activateItem: showOnAttach,
+        searchAllPanes: true
       })
     } else {
       this.treeViewOpenPromise = Promise.resolve()


50Wliu:
Reproduction steps:
1. Have the Tree View open in the left dock
2. Focus the left dock and execute `pane:split-down`
3. Move a dock item into the lower split (for example, the Git dock)
4. Focus the lower split and reload Atom
5. Two "Project"s should appear
6. Reload Atom
7. :boom:

50Wliu:
@shadow-light can you add a regression test for this?

shadow-light:
Sure, I can try. Can you point me to a similar one that I can learn from?

50Wliu:
@shadow-light I took a look at what tests we have now for package activation and it looks like there's nothing similar.  If you'd prefer, I can try adding the test for you.  Otherwise, here's what I'm thinking:

The tree-view setup is done [here](https://github.com/atom/tree-view/blob/64d36ed1437775cb934478cd4c16874659fc4818/spec/tree-view-package-spec.coffee#L56).  You can probably add a test in the "package activation" section that gets the active pane of the dock and [splits it](https://atom.io/docs/api/v1.31.2/Pane#instance-splitDown). Then require tree-view and `atom.workspace.open` it in the new split. The [active pane item](https://atom.io/docs/api/v1.31.2/Workspace#instance-getActivePaneItem) _should_ be `undefined`.

shadow-light:
If you don't mind adding it for me that would be great. Mainly because I'm not too familiar with Atom's API and already couldn't figure out how to do some things in my init script :laughing:

PR #1293, opened 2018-12-20
Restrict drag-and-drop actions to events from the Tree View

50Wliu:
### Requirements

### Description of the Change

Previously, the Tree View would respond to all items dragged over it, irrespective of whether or not any action could be performed with said item. This would result in the _illusion_ of being able to drop anything onto the Tree View, whereas in reality the drop action wouldn't do anything.

With this PR, the Tree View only responds when it knows that it can handle the event.

| Before | After |
|--------|-------|
|tree-view-drag-and-drop-before|tree-view-drag-and-drop-after|

(Note: you'll notice that an empty pane is still created. Fixed in atom/tabs#550)

### Alternate Designs

None.

### Benefits

There are two benefits.
1. Tree View folders will not be incorrectly highlighted when something that can't be dropped onto the Tree View is dragged over it.
2. Tree View does not unnecessarily prevent drag-and-drop event propagation, meaning that other packages (such as Tabs) can continue to receive those events.

### Possible Drawbacks

In the unlikely event that someone was using the `atom-tree-view-event` or `initialPaths` data items in a different package to work with the Tree View, that functionality will break. They will now need to use `atom-tree-view-root-event` and `atom-tree-view-event` (in addition to `initialPaths`), respectively.

### Applicable Issues

atom/tabs#539, which prior to this PR was not working correctly when a tab was dragged over the Tree View as it was preventing any drag-and-drop events from propagating up to the tabs package.

### DIFF
--- a/lib/root-drag-and-drop.coffee
+++ b/lib/root-drag-and-drop.coffee
@@ -27,7 +27,7 @@ class RootDragAndDropHandler
     return unless @treeView.list.contains(e.target)
 
     @prevDropTargetIndex = null
-    e.dataTransfer.setData 'atom-tree-view-event', 'true'
+    e.dataTransfer.setData 'atom-tree-view-root-event', 'true'
     projectRoot = e.target.closest('.project-root')
     directory = projectRoot.directory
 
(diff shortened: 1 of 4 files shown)


rafeca:
Thanks a lot!

Looks good to me, once you resolve the conflicts we can merge it 😃

50Wliu:
> once you resolve the conflicts we can merge it

The consequences of creating multiple PRs that _all_ add onto the end of the drag-and-drop specs 😬 😅

Issue #1307, opened 2019-03-26
Usability Concern: Context menu item "Show in File Manager" has non-standard naming on Mac/Windows


mattlubner:
### Prerequisites

* [x] Put an X between the brackets on this line if you have done all of the following:
    * Reproduced the problem in Safe Mode: http://flight-manual.atom.io/hacking-atom/sections/debugging/#using-safe-mode
    * Followed all applicable steps in the debugging guide: http://flight-manual.atom.io/hacking-atom/sections/debugging/
    * Checked the FAQs on the message board for common solutions: https://discuss.atom.io/c/faq
    * Checked that your issue isn't already filed: https://github.com/issues?utf8=✓&q=is%3Aissue+user%3Aatom
    * Checked that there is not already an Atom package that provides the described functionality: https://atom.io/packages

### Description

Opening a new issue here at the request of @50Wliu (original issue: atom/atom#18694).

The menu item "Show in File Manager" previously followed the naming conventions of the user's operating system (either "Show in Finder", "Show in Explorer", or "Show in File Manager", for Mac, Windows, or Linux, respectively). It's now hard-coded to "Show in File Manager". This is a usability regression that was accepted as part of #1266 (because it's no longer possible to know exactly which file manager will be used).

I'd like to propose reverting the name-change portion of #1266 (except on Mac it should be named "Reveal in Finder"), so that the operating system's naming conventions are followed.

I believe fewer users will be confused by pairing the old-but-accepted terminology with Atom opening the file or folder in their chosen third-party file manager, than are confused by the non-standard term "Show in File Manager" (which sounds like it invokes some new / custom Atom functionality). It stands to reason that more users expect the standard terminology than use a third-party file manager.

Would anyone have an objection to this change?

I'd be happy to submit a PR of exactly what I mean, if that would be helpful.

### Versions

```
macOS Mojave 10.14.3 (18D109)
```

```sh
# atom --version
Atom    : 1.35.1
Electron: 2.0.18
Chrome  : 61.0.3163.100
Node    : 8.9.3
```

rsese:
Thanks for opening a separate issue for this @mattlubner -

> I'd like to propose reverting the name-change portion of #1266 (except on Mac it should be named "Reveal in Finder"), so that the operating system's naming conventions are followed.

...

> I'd be happy to submit a PR of exactly what I mean, if that would be helpful.

We talked about this and are :+1: on this change, thanks for checking with us :bow:

50Wliu:
[...]

PR #1311, opened 2019-04-12
♿️ Restore platform-centric menu names for showing in file manager

mattlubner:
### Description of the Change

Addresses #1307. Restore platform-centric menu names for showing files in the user's file manager.

### Benefits

- Following each platform's idiomatic naming conventions will improve usability.

### Possible Drawbacks

- The name of the platform's default file manager is always used. This will be slightly misleading if `shell.showItemInFolder` opens a third-party file manager.

### Applicable Issues

- #1307 
- #1266
- atom/atom#18694

### DIFF
--- a/lib/tree-view.coffee
+++ b/lib/tree-view.coffee
@@ -560,13 +560,22 @@ class TreeView
     return unless filePath = @selectedEntry()?.getPath()
 
     unless shell.showItemInFolder(filePath)
-      atom.notifications.addWarning("Unable to show #{filePath} in file manager")
+      atom.notifications.addWarning("Unable to show #{filePath} in #{@getFileManagerName().toLowerCase()}")
 
   showCurrentFileInFileManager: ->
     return unless filePath = atom.workspace.getCenter().getActiveTextEditor()?.getPath()
 
     unless shell.showItemInFolder(filePath)
-      atom.notifications.addWarning("Unable to show #{filePath} in file manager")
+      atom.notifications.addWarning("Unable to show #{filePath} in #{@getFileManagerName().toLowerCase()}")
+
+  getFileManagerName: ->
+    switch process.platform
+      when 'darwin'
+        return 'Finder'
+      when 'win32'
+        return 'Explorer'
+      else
+        return 'File Manager'
 
   openSelectedEntryInNewWindow: ->
     if pathToOpen = @selectedEntry()?.getPath()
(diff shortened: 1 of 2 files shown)


mattlubner:
Hey @rsese @50Wliu, how do we proceed? Should I attempt to build Atom locally, write a test, or do something else?

This is the first time I've written _any_ Coffeescript (or contributed to Atom), so any guidance / input is appreciated! 🙏

lib/tree-view.coffee
```diff
@@ -560,13 +560,22 @@ class TreeView
     return unless filePath = @selectedEntry()?.getPath()
 
     unless shell.showItemInFolder(filePath)
-      atom.notifications.addWarning("Unable to show #{filePath} in file manager")
+      atom.notifications.addWarning("Unable to show #{filePath} in #{@getFileManagerName().toLowerCase()}")
```
lee-dohm: We should probably leave the case unmodified here.

```suggestion
      atom.notifications.addWarning("Unable to show #{filePath} in #{@getFileManagerName()}")
```

lib/tree-view.coffee
```diff
@@ -560,13 +560,22 @@ class TreeView
...
 
   showCurrentFileInFileManager: ->
     return unless filePath = atom.workspace.getCenter().getActiveTextEditor()?.getPath()
 
     unless shell.showItemInFolder(filePath)
-      atom.notifications.addWarning("Unable to show #{filePath} in file manager")
+      atom.notifications.addWarning("Unable to show #{filePath} in #{@getFileManagerName().toLowerCase()}")
```
lee-dohm: We should probably leave the case unmodified here.

```suggestion
      atom.notifications.addWarning("Unable to show #{filePath} in #{@getFileManagerName()}")
```

rsese:
Thanks @mattlubner!

> Should I attempt to build Atom locally, write a test, or do something else?

No need to build Atom locally, you can follow this guide to link your modified version of tree view:

https://flight-manual.atom.io/hacking-atom/sections/contributing-to-official-atom-packages/

I confirmed that tests aren't necessary in this case but it would be good if you can manual verify that things look ok on Windows, macOS, and Linux with your changes.

mattlubner:
Thanks for the review @rsese! Just manually verified that things look good on macOS. I don't have a Window or Linux environment that I can test on, though.

50Wliu:
Thanks!

Issue #1344, opened 2020-06-21
Build failing


ardentbriar:
### Prerequisites

* [x] Put an X between the brackets on this line if you have done all of the following:
    * Reproduced the problem in Safe Mode: <https://flight-manual.atom.io/hacking-atom/sections/debugging/#using-safe-mode>
    * Followed all applicable steps in the debugging guide: <https://flight-manual.atom.io/hacking-atom/sections/debugging/>
    * Checked the FAQs on the message board for common solutions: <https://discuss.atom.io/c/faq>
    * Checked that your issue isn't already filed: <https://github.com/issues?utf8=✓&q=is%3Aissue+user%3Aatom>
    * Checked that there is not already an Atom package that provides the described functionality: <https://atom.io/packages>

### Description

Builds are currently failing. This failed build on Travis is an unchanged copy of `master`:

https://travis-ci.org/github/ardentbriar/tree-view/jobs/700467325

### Steps to Reproduce

1. Clone a fresh copy of `tree-view`.
2. Run `apm install` to install deps.

**Expected behavior:**

Deps install successfully.

**Actual behavior:**

The node-gyp build of `pathwatcher` fails. I see the same error in the Travis build and locally:

```
$ apm install
Installing modules ✗
> pathwatcher@8.0.2 install /home/user/repo/github/tree-view/node_modules/pathwatcher
> node-gyp rebuild

make: Entering directory '/home/user/repo/github/tree-view/node_modules/pathwatcher/build'
  CXX(target) Release/obj.target/pathwatcher/src/main.o
make: Leaving directory '/home/user/repo/github/tree-view/node_modules/pathwatcher/build'

In file included from /home/user/.atom/.node-gyp/.cache/node-gyp/5.0.13/include/node/v8-internal.h:13,
                 from /home/user/.atom/.node-gyp/.cache/node-gyp/5.0.13/include/node/v8.h:25,
                 from /home/user/.atom/.node-gyp/.cache/node-gyp/5.0.13/include/node/node.h:63,
                 from ../../nan/nan.h:53,
                 from ../src/common.h:6,
                 from ../src/main.cc:1:
/home/user/.atom/.node-gyp/.cache/node-gyp/5.0.13/include/node/v8config.h:307:49: warning: ‘BufferReference’ is deprecated: Use MemorySpan<const uint8_t> [-Wdeprecated-declarations]
  307 |   declarator __attribute__((deprecated(message)))
      |                                                 ^
[...]

PR #1346, opened 2020-06-21
Add autoopen option

### LINKED ISSUE #530: Disable tree-view
How do I disable the tree-view?

I would like to be hidden at start up.

I don't want it to show up every time I open a file.

ardentbriar:
Since builds are currently failing #1344 I can't run tests, so leaving them to update after that's resolved. I'm calling this a draft PR until then.

### Description of the Change

Adds boolean config setting Show On Attach to control whether the tree view is automatically shown when a new window is opened.

Default is true to preserve current behavior.

### Alternate Designs

None. The change is small.

### Benefits

Allow keeping the tree view hidden by default. This has been requested for years in #530 and [on the forum](https://discuss.atom.io/t/hide-treeview-by-default/7718).

### Possible Drawbacks

Choice fatigue. It adds a new option which must be thought through when the configuring the package.

I don't think there are excessive options. After this change there are 8 boolean options. The default value preserves current behavior so most people won't need to think about it.

### Applicable Issues

Closes #530.

### DIFF
--- a/lib/tree-view-package.js
+++ b/lib/tree-view-package.js
@@ -21,7 +21,10 @@ class TreeViewPackage {
     }))
 
     const treeView = this.getTreeViewInstance()
-    const showOnAttach = !atom.workspace.getActivePaneItem()
+    const showOnAttach = (
+      atom.config.get('tree-view.showOnAttach') &&
+      !atom.workspace.getActivePaneItem()
+    )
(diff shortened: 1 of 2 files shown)


pokono:
Can't wait for this!!

bueler:
This would be nice to have!  Such a simple thing ...

PR #1359, opened 2020-10-03
Do not wipe original file when duplicating

gosukiwi:
### Description of the Change

When duplicating a file, if the file already exists, an error should be thrown. Node's `fs.existsSync` is already aware of filesystem case sensitivity, so just checking if a file exists or not is enough. On case-sensitive file systems, it will allow to create a new file with different case. On case-insensitive ones, it will complain that the file already exists.

So for example, in Windows, that is case-insensitive, you can't duplicate `FOO.txt` if `foo.txt` exists. On Linux, that is case-sensitive, you can.

Now this is hard to test, because it depends on the OS-specific behavior. I created a little helper function to check the current FS for case-sensitivity and ran the suite on Windows, and it worked. I also tested the feature manually on macOS. I don't have a Linux setup though, so I was not able to test on Linux. I did test on WSL and `existsSync` was working as expected.

### Alternate Designs

--

### Benefits

This will fix the bug described in https://github.com/atom/atom/issues/20657

### Possible Drawbacks

I did not test the full feature on Linux, only tested on Windows and macOS. I did test the `existsSync` functionality using WSL and it was working as expected.

### Applicable Issues

https://github.com/atom/tree-view/issues/1361

### DIFF
--- a/lib/copy-dialog.coffee
+++ b/lib/copy-dialog.coffee
@@ -23,7 +23,7 @@ class CopyDialog extends Dialog
       @close()
       return
 
-    unless @isNewPathValid(newPath)
+    if fs.existsSync(newPath)
       @showError("'#{newPath}' already exists.")
       return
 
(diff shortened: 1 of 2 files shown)


sadick254 (approved):
This looks great. 👍 . Tested on Linux and works as expected.

Issue #1368, opened 2020-11-16
One more time about this file deleting issue


pjylkka:
I have Windows 10,  project is on samba network share on Ubuntu server 20.04
Can't delete files due of this "is there recycle bin on the drive where files are stored".

I added "delete permanently package" - now shift-del asks if i want to delete permanently, - I accept, get no error messages but _file still does not disappear_.  I can delete file on explorer tho with no any problem.,
- User logged in has full permissions on samba share. 
- Running Atom as administrator doesn't seem to make difference.

As this would be so such a wonderful software - Couldn't you guys just add option in the preferences if user wants be able to 
"[X] Delete files without using recycle bin" without hacking it.  I'd very much rather code with Atom instead of that horrible visual studio code piece of crap

aminya:
I added a PR that allows deleting things permanently. I have seen this error too, and it is very annoying. So, #1373 should give a way to avoid the error.

Issue #1237, opened 2018-03-23
Unable to delete file (need gvfs-trash)


wyqydsyq:
When trying to delete a file, I get an error about `gvfs-trash`, there is no `gvfs-trash` package for my distribution (OpenSUSE Tumbleweed) so deleting via Atom Tree View is impossible for me.

There should really be a fallback for this, say, using the standard `rm` command to delete files if `gvfs-trash` is not available.

fmujakitz:
Had the same issue. Fixed by setting env var for `ELECTRON_TRASH=gio`.

lefuturiste:
Same issue on antergos archlinux 
How I can set the env var forever ? What is the command @fmujakitz

wyqydsyq:
If you want to persist an env var you can add it to your shell's initialization script e.g. [.bashrc](https://unix.stackexchange.com/questions/129143/what-is-the-purpose-of-bashrc-and-how-does-it-work)

lefuturiste:
Don't work also

rsese:
@lefuturiste does `gio trash` work ok from the command line?

```
echo hello > delete-me.txt
gio trash delete-me.txt
```

rsese:
@lefuturiste what version of Atom are you running (`atom -v`)?

@wyqydsyq do you have `gio` installed on your system?  If so, does setting `ELECTRON_TRASH` work for you?

lefuturiste:
@rsese 

Atom    : 1.28.0
Electron: 2.0.3
Chrome  : 61.0.3163.100
Node    : 8.9.3

Dzordzu:
@rsese 
Same here. The issue appeared after last arch update
gio is installed
Files are placed on the same partition, where the atom is

Atom : 1.30.0
Electron: 2.0.9
Chrome : 61.0.3163.100
Node  : 8.9.3

AlynxZhou:
It seems that gvfs updates and dropped gvfs-trash, instead we should use gio trash now?

pzmarzly:
This is an Electron issue: https://github.com/electron/electron/issues/15011

kebertxela:
> It seems that gvfs updates and dropped gvfs-trash, instead we should use gio trash now?
> Temp workaround is to create a script as a warpper:
> ```
> sudo vim /usr/bin/gvfs-trash
> #!/bin/bash
> # GVFS updated and dropped gvfs-trash to gio, but Atom didn't update.
> /usr/bin/gio trash "$@"
[...]

Issue #1046, opened 2017-03-15
Unable to delete files over network on macOS


TeapotNickS:
### Prerequisites

* [X] Put an X between the brackets on this line if you have done all of the following:
    * Reproduced the problem in Safe Mode: http://flight-manual.atom.io/hacking-atom/sections/debugging/#using-safe-mode
    * Followed all applicable steps in the debugging guide: http://flight-manual.atom.io/hacking-atom/sections/debugging/
    * Checked the FAQs on the message board for common solutions: https://discuss.atom.io/c/faq
    * Checked that your issue isn't already filed: https://github.com/issues?utf8=✓&q=is%3Aissue+user%3Aatom
    * Checked that there is not already an Atom package that provides the described functionality: https://atom.io/packages

### Description

I have a project loaded over a local network and I am not able to delete files, I get an error "Is Trash enabled on the volume where the files are stored"

It looks like the issue is simply that the networked machine doesn't have a trash bin however this means that any file that I need to delete needs to be done via Finder.

This seems very similar to [#856](https://github.com/atom/tree-view/issues/856) but I was asked to open a new bug for macOS by @50Wliu

### Steps to Reproduce

1. Load project into Atom over a network drive
2. Delete a file
3. thats all the error will be shown
![screen shot 2017-03-15 at 15 30 05](https://cloud.githubusercontent.com/assets/15928482/23956693/37388682-0995-11e7-94fa-0e7f23fcd1b8.png)

**Expected behavior:** [File should be deleted]

**Actual behavior:** [Error is shown.]

**Reproduces how often:** [100%]

### Versions

Atom 1.16.0-beta0 x64
macOS Sierra 10.12.3

50Wliu:
I believe on macOS Trash is enabled by the existence of a `/.Trash-1000` folder.  Are you able to create that folder using the following commands?
```bash
sudo mkdir /.Trash-1000
sudo chown user:user ./Trash-1000
```
(This actually looks like https://github.com/atom/tree-view/issues/345; if so, sorry for asking you to create a new issue.  I didn't find #345 in time as it's missing the mac label.)

TeapotNickS:
Not a problem, thanks for the fast replies.

When I delete files via Finder they get permanently deleted and not put in the trash (which is expected) If however you are on the server and delete a file they are put in the trash which makes sense to me.

I wouldn't actually want deleted files/folders put into a remote trash folder as someone would need to clear out the trash otherwise it could become a very large trash folder. 

[...]

Issue #345, opened 2015-01-28
delete file in treeview doesn't work when files are in specific folder


buildcomplete:
right clicking an item in the treeview and selecting 'delete' doesn't do anything (except showing the 'move to recycle dialog)

* I have read/write access to the file, and I am able to rename the file...
* I am using Atom version 0.176.0
* Atom is able to delete files that  are in my home directory!
* Settings myself as owner and group of the file outside my home directory ('/var/www/app_url/app/some-new-file) doesn't help.
* I am able to delete the file using the filemanager (Files 3.10.1)
* I am running Trisquel GNU/linux with Gnome (I can't figure out the version, but it's the one that comes with the Trisquel 7.0)

Reproduce,
* start atom.
* open a file so the gets some content in the treeview
* right click in the tree view and select 'add file' 
* type  'some-new-file'
* right click 'some-new-file' and select delete
* Click 'Move to Trash' in the dialog
* Observe that the file is not deleted as expected.
* There is not output in the developers tool hinting the problem

lee-dohm:
I was unable to reproduce this using Atom v0.177.0-e7b790c on Mac OS X 10.10.1. Here are my repro steps:

1. `sudo mkdir /usr/share/test`
1. `sudo chmod Lee /usr/share/test`
1. `atom /usr/share/test`
1. Create a new file `foo.txt` in the root of the project
1. Type `blah blah blah` into `foo.txt` and save the file
1. Right-click `foo.txt` in the Tree View and select Delete
1. Click "Move to Trash" button

**Expected and Actual:** `foo.txt` disappears from the Tree View and from the actual directory.

buildcomplete:
I tried the exact same steps as lee-dohm (substituing chmod with chown...)
the file still sits there smiling at me :)

ricardograca:
I'm using Ubuntu 14.04 and I also can't reproduce your issue by following the outlined steps.

This is most likely a permissions issue. The fact that your home directory is not the typical `/home/username` makes me think that your problematic subdirectories are probably being created with permissions that cause this issue. Maybe there's a sticky bit being set?

buildcomplete:
Hello ricardograca,
I have checked the folowing paths that was involved above
/usr
/usr/share
/var
/var/www
/var/www/app_url
/var/www/app_url/app
and none of them have the sticky bit set.
It is also possible to delete the file from the normal filemanager (nautilus).

by the way, my home folder is certainly in the standard /home/username

[...]

PR #1373, opened 2020-12-28
feat: Add a command to delete items permanently

aminya:
### Description of the Change

This adds a command to delete items permanently. It uses the Sift+delete shortcut similar to operating system shortcuts. It prompts before doing so.

### Benefits

- Allows deleting files permanently
- Provides a way to avoid the issues mentioned in the next section

### Applicable Issues

This PR does not provide an alternative trashing implementation for tree-view:remove, but it provides a new command to avoid these issues by using tree-view:remove-permanently. In later PRs, we can use the [trash](https://www.npmjs.com/package/trash) package as a fallback to fix and close these issues: 
https://github.com/atom/tree-view/issues/1368
https://github.com/atom/tree-view/issues/1237
https://github.com/atom/tree-view/issues/345
https://github.com/atom/tree-view/issues/1046

### Possible Drawbacks
none


sadick254 (commented):
Thanks for the feature. I have noticed that this feature has no specs. Would you mind adding some specs on it?

spec/tree-view-package-spec.coffee
```diff
@@ -2747,391 +2747,393 @@ describe "TreeView", ->
           atom.commands.dispatch(treeView.element, "tree-view:duplicate")
           expect(atom.workspace.getModalPanels().length).toBe(0)
 
-    describe "tree-view:remove", ->
-      beforeEach ->
-        jasmine.attachToDOM(workspaceElement)
+    for cmd in ["remove", "remove-permanently"]
```
aminya, line 2750: Here I just added a for a loop. The diff noise is because of indentation

spec/tree-view-package-spec.coffee
```diff
@@ -2747,391 +2747,393 @@ describe "TreeView", ->
...
+              spyOn(atom, 'confirm').andCallFake (options, callback) -> callback(0)
 
-          waitsFor 'directory to be deleted', ->
-            callback.mostRecentCall.args[0].pathToDelete is dirPath2
+              atom.commands.dispatch(treeView.element, "tree-view:#{cmd}")
 
-          runs ->
-            openFilePaths = atom.workspace.getTextEditors().map((editor) -> editor.getPath())
-            expect(openFilePaths).toEqual([])
+            waitsFor 'directory to be deleted', ->
```
aminya, line 2854: I am not so familiar with this old style of waiting for promises. It seems it doesn't actually wait. 

@uzitech Do you know how to make it wait for async functions?

UziTech: I think you have to use `waitsForPromise`

aminya: Maybe adding to the condition can fix the issue too. Because the promise is not accessible through the test.

```coffee
			  callback.mostRecentCall isnt undefined and
              callback.mostRecentCall.args isnt undefined and
```

aminya: It seems that the callback isn't called at all.

The first log is from the actual function and the second is what the spy sees.
image

UziTech: Can't you just copy the tests for `tree-view:remove`?

aminya: I did, but it doesn't work for some reason. The spies are not called for unknown reasons.

aminya:
I don't know why the test failed again. It was working 😥 Writing tests with CoffeeScript 1 and Jasmine 1 is a very bad experience.

aminya:
The issue was that I was comparing arrays and objects together using `==` instead of a `deep equality`. I fixed the issue now.

darangi:
Thanks for the contribution 🙇🏾‍♂️ .

 Could you fix the conflicts?

PR #1390, opened 2021-08-08
Handle text overflow in add-projects-view buttons

sid-jain:
This makes sure that text does not overflow outside of buttons (if the view width is small enough) by truncating the text and adding trailing ellipsis.

### Description of the Change

Disable wrapping and set `text-overflow` to `ellipsis` in `add-projects-view` `btn` class

**Before this change**

before

**After this change**

after

### DIFF
--- a/styles/tree-view.less
+++ b/styles/tree-view.less
@@ -34,6 +34,12 @@
     .icon::before {
       color: #c1c1c1;
     }
+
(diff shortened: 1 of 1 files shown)


darangi:
@sid-jain  Thanks for your contribution 🙇🏾

PR #1392, opened 2021-08-12
Attach elements to jasmine DOM

sadick254:
Atom migrated to using customElements from document.registerElement. Due
to this move, now there is need to attach some elements to the DOM on
some specs.


olalayray (commented):
@

PR #1389, opened 2021-06-15
Implement option to hid tree pane on startup

joaotmartins:
### Issue or RFC Endorsed by Atom's Maintainers

Refers to issues:
#1370 #1375 

### Description of the Change

Adds a new configuration option, `hiddenOnStartup`, that allows a user to ask for the project pane to be hidden when Atom starts.
If option isn't present or set to false, behavior stays the same, i.e., project pane is shown on startup.
Option is checked on package activation function, pane is hidden/shown accordingly.

### Alternate Designs

None that I could see.

### Possible Drawbacks

None that I could see. Default behavior is preserved, so users do not have to update any configuration file.

### Verification Process

Added corresponding tests in package spec to verify that:

- Absence of option preserves default behavior
- Default of option is to be false, preserving old behavior
- Setting flag hides panel as designed.

### Release Notes

Adds new configuration option 'hiddenOnStartup', that allows starting atom with the tree-view collapsed (hidden).

Issue #1370, opened 2020-12-16
[RFE] Request for an option to prevent the Project pane from opening on startup


pneedle-rh:
## Summary

This is a request for an option to prevent the **Project** pane from opening when Atom is used to open a single file.

## Motivation

I typically open single files in Atom from the command line, by using the following command:

`$ atom <file>.adoc`

I do this many times in a day and I jump in and out of different Git projects. I don't use the project features within Atom and prefer to navigate from the CLI and open files in Atom from there. Often I close Atom when I have finished editing a single file.

It seems that recently a change was introduced in Atom that results in the **Project** pane being open when Atom is started against a single file. The project pane states that "Your project is currently empty". This means that I have to close the project pane every time I open Atom, to be able to view the file in the whole Atom window.

It would be great to have the option to suppress the **Project** pane on startup, so that I don't have to close it numerous times per day.

## Describe alternatives you've considered

Using the `CTRL+\` keyboard shortcut on Fedora, or `CMD+\` on MacOS, to close the pane each time. This is not really a workaround, because I still have to do this many times per day. It is just quicker than using the mouse to close the pane each time.

pneedle-rh:
Perhaps something like a 'Restore current pane configuration when Atom is restarted' option could resolve this.

pneedle-rh:
Looking at the [Atom Release Notes](https://atom.io/releases), I am pretty sure this change in behaviour was introduced in 1.53.0 in this tree-view pull request:

* https://github.com/atom/tree-view/pull/1315 - Always show Tree View

mortar-and-pestle:
@pneedle-rh 

Yep. 

**Possible Drawbacks**
Tree View will now appear, even when no projects are open.

Hopefully, we can get the option to disable this functionality.

mortar-and-pestle:
> 
> 
> Looking at the [Atom Release Notes](https://atom.io/releases), I am pretty sure that this change in behaviour was introduced in Atom 1.53.0 through this tree-view pull request:
> 
>     * #1315 - Always show Tree View

@pneedle-rh

Yep.

Possible Drawbacks
Tree View will now appear, even when no projects are open.

Hopefully, we can get the option to disable this functionality.

DiscoveryOV:
Yeah can we get an option to disable this?

I don't need Atom telling me how I can add a project every time I open it. I know how to do it if I need to. I mostly use Atom for single-file editing, and this stupid pane just adds to the bloat.

zekeadam:
[...]

PR #1396, opened 2021-10-30
Re-implements hiding behavior based on promise to fix timing error.

joaotmartins:
### Description of the Change

Re-implemented hiding code to perform conditional hiding as a callback of treeViewOpenPromise. This fixes an error where we might have attempted to hide the pane before it existed (usually only happened when Atom was started up with a clean profile).

### Alternate Designs

None.

### Benefits

Tree-view hiding works even when there are a few timing differences in the pane initialization.

### Possible Drawbacks

None.

### Applicable Issues

See comments to atom/tree-view#1389


joaotmartins:
Fixes the original implementation for #1370 and #1375 .

Jelisacooper4 (approved):

